PR #1036, opened 2026-05-10
perf(wado-cli serve): unblock concurrency, cache TLS roots, enforce per-request timeout, drain on shutdown

gfx:
Production-readiness fixes for `wado serve`. The previous implementation
had a `Mutex<Linker>` that serialized all request instantiation, rebuilt
rustls' `ClientConfig` (and re-read CA bundles from disk) per request,
re-resolved component exports per request, and had no defense against
runaway guests or unclean shutdown.

Changes:
- Replace `Arc<Mutex<Linker>>` with `Arc<ServicePre>` so concurrent
  requests no longer block on a global mutex; each request goes
  straight to `ServicePre::instantiate_async`, which front-loads the
  per-export string lookups that `Service::instantiate_async` repeated.
- Cache `rustls::ClientConfig` in a process-wide `LazyLock` inside
  `WadoHttpHooks::new` so per-request `WasiState` construction no
  longer reads `SSL_CERT_FILE` / `SSL_CERT_DIR` from disk every call.
- Add `--timeout <seconds>` (default 30) and enforce it two ways: a
  per-store `set_epoch_deadline` that traps tight-loop guests via a
  background ticker, plus `tokio::time::timeout` for host-side hangs.
  Trapped/timed-out requests now return 504 instead of holding a tokio
  task forever.
- Wait for SIGINT/SIGTERM and drain in-flight connections (with a
  bounded deadline) instead of killing them mid-request.

https://claude.ai/code/session_01Wuv6sKcAETC3XxFae9sG7j

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2701,19 +2701,7 @@ dependencies = [
  "pin-project-lite",
  "signal-hook-registry",
  "socket2",
- "tokio-macros",
  "windows-sys 0.61.2",
-]
-
-[[package]]
-name = "tokio-macros"
-version = "2.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "385a6cb71ab9ab790c5fe8d67f1645e6c450a7ce006a33de03daa956cf70a496"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


gfx:
@copilot apply changes based on the comments in [this thread](https://github.com/wado-lang/wado/pull/1036#pullrequestreview-4258844000)

Issue #1230, opened 2026-05-29
`core:cli::println` leaks a stream resource per call (traps with "resource table has no free keys")


gfx:
## Summary

Each call to `core:cli::println` (and presumably `eprintln` / `print` / `eprint` / `print_bytes`, which share the same pattern) allocates a WASI stream/future resource that is never freed on the host side. A program that prints many lines fills the component resource table and traps.

## Minimal repro

```wado
use { println, Stdout } from "core:cli";

export fn run() with Stdout {
    for let mut i = 0; i < 200000; i += 1 {
        println(`line {i}`);
    }
}
```

```
$ wado run repro.wado
... (prints ~166k lines, then:)
Error: Runtime error: error while executing at wasm backtrace:
    0: core:cli/println
    1: .../run
Caused by:
    resource table has no free keys
```

It traps reproducibly at ~166,666 calls.

## Where

`wado-compiler/lib/core/cli.wado` — `println` (and the sibling helpers) do, per call:

```wado
let [rx, tx] = Stream::<u8>::new();
let handle = Stdout::write_via_stream(rx);
write_to_stream(tx, message, true);   // writes, then tx.drop()
drop_cli_write_future(handle);
```

`tx` is dropped and `drop_cli_write_future(handle)` is called, yet the resource table still grows without bound — something on this path (the readable end consumed by `write_via_stream`, or the future handle) is not actually being released. Net effect: one leaked resource per print call.

## Impact

Any program that prints a large number of lines in a loop crashes. Workaround today is to buffer output and make a single `print_bytes` call, but that is not viable for long-running or high-volume output (e.g. CLI tools, log-heavy programs).

## Expected

`println` and friends should release every resource they create, so an unbounded number of print calls runs without exhausting the resource table.

Issue #1229, opened 2026-05-29
Fixed-point formatting `{:.Nf}` under-rounds small-magnitude values


gfx:
## Summary

The fixed-point float formatter (template `{x:0.6f}` and friends) rounds **toward zero too aggressively for small magnitudes**: values that should round *up* into the last printed digit stay at `0`. The round-up threshold sits near `~0.95` of the last digit's unit instead of the correct `0.5`.

## Minimal repro

```wado
use { println, Stdout } from "core:cli";

export fn run() with Stdout {
    println(`{0.0000006:0.6f}`);   // prints 0.000000, should be 0.000001
}
```

`0.0000006` (= 6×10⁻⁷) is 0.6 of the last (millionths) digit, so correct round-to-nearest gives `0.000001`. Wado prints `0.000000`.

## Boundary characterization

Formatting `k·10⁻⁸` with `{:0.6f}`, comparing against Rust `{:.6}` (correct reference):

| value | Rust | Wado |
| ----- | ---- | ---- |
| 0.50×10⁻⁶ | `0.000000` | `0.000000` |
| 0.55×10⁻⁶ | `0.000001` | `0.000000` ❌ |
| 0.60×10⁻⁶ | `0.000001` | `0.000000` ❌ |
| …         | `0.000001` | `0.000000` ❌ |
| 0.90×10⁻⁶ | `0.000001` | `0.000000` ❌ |
| 0.95×10⁻⁶ | `0.000001` | `0.000001` ✅ |

- Rust starts rounding up at ~`0.5×10⁻⁶` (correct).
- Wado starts rounding up only at ~`0.95×10⁻⁶`.

So every value in roughly `[0.55×10⁻⁶, 0.95×10⁻⁶)` (and the analogous tiny ranges) is rounded **down** when it should round **up**. Normal-magnitude values (e.g. `0.1234565`) format correctly — the bug is specific to small magnitudes with leading zeros.

The observed boundary (~0.95) is consistent with a rounding bias that is ~10× too small (≈`0.05` unit instead of `0.5` unit) being added before truncation: `x + 0.05×10⁻⁶ < 1×10⁻⁶` for `x = 0.6×10⁻⁶` → truncates to `0`; only at `x ≈ 0.95×10⁻⁶` does `x + 0.05×10⁻⁶` cross `1×10⁻⁶`.

## How it was found

[...]

PR #1231, opened 2026-05-29
bench: denominate light benchmarks; refresh results (best-of-three, whole-ms) and skill

gfx:
## What

Refreshes the benchmark suite results and reporting, and scales the lightest workloads so every result reports in whole milliseconds without integer rounding swallowing real signal.

### Denomination (program changes)

The lightest workloads are scaled up; the factor lives in each program and is applied consistently across **every** language implementation:

- **sieve**: count to **100M** (was 10M) — problem-size scaling, one allocation (looping would have re-allocated the array and folded GC churn into the measurement).
- **fts**: **5M** conversions (was 500K) — problem-size scaling.
- **JSON** twitter/canada/catalog (+v2/v3): iterate **×10** (was ×1 / ×5).
- **zlib**: iterate **×100** (was ×10).

Per-iteration verification outputs (prime count, byte sums, event counts) are unchanged; benchmark smoke tests still pass.

### Results & docs

- `benchmark/README.md`: every table refreshed from **best-of-three** runs, all times now in **whole milliseconds**. Added a methodology note (best-of-three; which workloads are scaled and why). Because the denominated JSON/zlib figures are warm steady-state totals, some ratios shift from the earlier cold single-shot numbers — e.g. json-canada `14.0x → 3.8x` once cold-start tiering is excluded, and sieve `1.9x → 1.3x` in the memory-bound 100M regime.
- `.claude/skills/benchmark/SKILL.md`: now covers all 10 `benchmark-all` benchmarks plus the separate `http-routing` run (oha prerequisite, tunables, req/s, curated subset), documents the best-of-three and whole-ms conventions, and explains the denomination factors and how to retune them.
- Earlier commits on this branch also refreshed the HTTP-routing results and the `benchmark-all` tables, dropped `gh` from the dev environment, and made the SessionStart hook initialize the `vendor/wasmtime` submodule.

## Testing

- Ran each changed benchmark three times; report the best per implementation.
- `wado test` over the changed benchmark dirs: 9 smoke tests pass, all compile.
- `mise run format` clean.

## Follow-ups (filed, not fixed here)

While scaling `fts` to 5M I uncovered two pre-existing bugs, filed separately:

- #1229 — fixed-point `{:.Nf}` under-rounds small-magnitude values (causes the `fts` Wado `byte_sum` to be off-by-one vs native; benchmark itself is correct).
- #1230 — `core:cli::println` leaks a stream resource per call.

https://claude.ai/code/session_01PELiaHkTZv9VhWpBpJuDub

---
_Generated by [Claude Code](https://claude.ai/code/session_01PELiaHkTZv9VhWpBpJuDub)_

### DIFF
--- a/.claude/hooks/mise-setup.sh
+++ b/.claude/hooks/mise-setup.sh
@@ -102,7 +102,6 @@ if [ -n "$CLAUDE_ENV_FILE" ]; then
 export PATH="$HOME/.local/bin:$PATH"
 export MISE_YES=true
 export MISE_TRUSTED_CONFIG_PATHS="$HOME:$PWD"
-export MISE_GITHUB_TOKEN=$(gh auth token)
 eval "$(mise activate bash)"
 EOF
     log "Environment persisted to CLAUDE_ENV_FILE"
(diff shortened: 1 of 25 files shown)


gfx:
### Note on the red **Benchmark** check — expected, safe to merge past

The Benchmark gate fails because this PR **intentionally denominates** the benchmark workloads to keep results in whole milliseconds (sieve → 100M, fts → 5M, JSON → ×10, zlib → ×100). `run_wado_benchmarks.ts` runs the same `*.wado` programs, so the Wado-only tracked times rise ~8–20× and trip `fail-on-alert` (`fail-threshold: 200%`).

This is **not a performance regression** — it's a one-time workload step-change. The gh-pages baseline only updates on push to `main`, so this PR can't compare green against the old baseline; once merged, the main run records the new baseline and subsequent PRs compare correctly.

**Action:** merge past the red Benchmark check (admin override). No code change is warranted.


---

PR #1310, opened 2026-06-06
Make namespace imports robust across all symbol kinds

### LINKED ISSUE #1169: Move namespace-import canonicalization from resolver lookup-time to `bind` phase
## Context

PR #1165 replaced the `NsStripper` desugar pass with a lookup-time helper `Resolver::strip_ns_prefix` (`wado-compiler/src/resolver.rs:328`). When the resolver looks up an ident like `helper::foo`, the helper canonicalizes it to `foo` before hitting the various registries.

This is a strict improvement over `NsStripper`: the AST keeps `helper::foo` as the user typed it, so LSP cursors land on either `helper` or `foo` correctly. But the canonicalization step is **opt-in per call site**:

```
$ grep -n 'strip_ns_prefix' wado-compiler/src/resolver*.rs wado-compiler/src/resolver/*.rs
wado-compiler/src/resolver.rs:328:    pub(super) fn strip_ns_prefix<'s>(&self, name: &'s str) -> Option<&'s str> { … }
wado-compiler/src/resolver/expr.rs:477:        let ident = if let Some(stripped) = self.strip_ns_prefix(&ident.name) { … }
wado-compiler/src/resolver/expr.rs:3101:        let name = if let Some(stripped) = self.strip_ns_prefix(raw_name) { … }
wado-compiler/src/resolver/stmt.rs:834:                        .strip_ns_prefix(expected_name) …
wado-compiler/src/resolver/stmt.rs:1427:                    .strip_ns_prefix(variant_name) …
wado-compiler/src/resolver/stmt.rs:1700:                            .strip_ns_prefix(expected_name) …
```

Six sites. Any new resolver path that takes a name string (e.g. a future macro-like construct, a new pattern variant, a new RHS shape) must remember to call `strip_ns_prefix` before the registry lookup, or `helper::foo` references will silently fail to resolve.

## Proposal

Canonicalize at `bind` time instead. Concretely:

1. While `bind` walks the AST, record on each `Ident` / `IdentLikePattern` / etc. the **canonical name** to use for lookups, alongside its written form. The AST node still owns the source text and span (so LSP keeps working); a side table or a new optional field carries the canonical string.
2. Resolver call sites switch from `let name = self.strip_ns_prefix(&ident.name).unwrap_or(&ident.name);` to `let name = ident.canonical_name();` (or read from the side table). One uniform read, no per-callsite branching.
3. `Resolver::strip_ns_prefix` is deleted. `namespace_imports` no longer needs to be queried during expression resolution.

## Why now / cost

- **Robustness**: missing a call to `strip_ns_prefix` is a silent bug. A bind-time canonical name is impossible to forget.
- **Cost**: small AST extension (one field per ident-like node, or one side-table keyed by `AstId`). `bind` already walks every node and tracks namespace imports.
- **LSP**: unchanged — the source span and written name are preserved on the AST exactly as today. Use→def edges still resolve against the canonical name.

## Out of scope

- Multi-segment forms `ns::Type::case` continue to route through dedicated namespace paths (`resolve_ident` / `resolve_call`); the canonicalization here only covers the `ns::member` shape that `strip_ns_prefix` covers today.
- No change to the `namespace_imports` registry itself.

## Cross-refs

- PR #1165 — introduced `strip_ns_prefix`.
- `wado-compiler/CLAUDE.md` "LSP-Friendly Compiler Architecture" — use→def edges are recorded by the real resolver; this issue keeps that property while removing the per-callsite obligation.

Priority: low–medium. The lookup-time approach works; this is consolidation, not a correctness fix.

gfx:
Closes #1169.

## Summary

Namespace imports (`use ns from "..."`) only canonicalized the `ns::member` shape at a handful of annotate-time call sites, and the annotate (`resolve_ident`) and reify (`reify_ident`) phases diverged. As a result, many symbol kinds reachable through a namespace import failed to resolve — some silently, some with an ICE — and qualified access that *did* resolve collapsed to a shared bare-name registry, so two namespaces exporting the same member collided.

This PR makes namespace-import resolution uniform and module-scoped.

## What was broken (each now has an e2e fixture)

Found by sweeping every symbol kind through a namespace import:

| form | symptom before |
| --- | --- |
| `ns::GLOBAL` | resolved as a function ref → wrong type |
| `ns::func` (as a value) | ICE: unresolved `MethodCall` in WIR translation |
| `ns::Flags::Member` | ICE: invalid core Wasm (`nothing on stack`) |
| `ns::Type` (type position) | `unknown type` |
| `ns::Struct { ... }` | parse error |
| `ns::Struct { ... }` destructuring | parse error |
| `ns::Case` patterns | pattern mismatch |
| `ns::Type::CONST` | unknown identifier |
| `a::X` + `b::X` (two namespaces, same member) | resolved to whichever was imported last (silent miscompile) |

Function *calls* (`ns::func()`) already disambiguated via `resolve_call`; everything else went through the bare-name path.

## Approach

1. **Consolidate canonicalization.** A single shared helper rewrites `ns::member` to a namespace-scoped alias `ns$member` (`$` is not a valid identifier character). Both annotate and reify use it, and the type resolver applies it at the one `TypeLookup::lookup_ref` chokepoint — the two phases can no longer drift.
2. **Register namespace members as aliases.** Each namespace import registers its public symbols under `ns$member` in the same per-name maps that back `use { X as Y }` (`imported_type_sources` / `import_original_names`, `imported_globals`, the symbol-table imports, `associated_constants`). The namespace stays baked into every lookup, so `a$CONFIG` and `b$CONFIG` resolve to their own modules.
3. **Delete the special cases.** Because the alias routes through the existing import machinery, the dedicated `ns::Type::Case` branches in `resolve_ident` / `reify_ident` are redundant and removed (net **−150 lines** of resolver code).
4. **Parser** accepts `ns::Type { ... }` in expression and pattern position.

Namespace symbols are reachable only through the `ns::` prefix, matching the spec ("symbols are accessed with the `ns::` prefix"); the previously-incidental bare access was never a documented feature.

## Tests

- New e2e fixtures covering globals, function refs, flags, type position, struct literals, patterns, associated constants, return-position types, and a dedicated `namespace_import_collision` fixture (two namespaces exporting the same names).
- `mise run test` (compiler + e2e): green, 0 failures.
- `mise run test-wado` (stdlib + packages): 3183 passed, 0 failed, 33 `#[TODO]` pending.
- clippy clean; formatted.

https://claude.ai/code/session_01Fkh4H5e4YrbEZHViGiMtfY

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fkh4H5e4YrbEZHViGiMtfY)_


gfx:
Re: the `sieve (-O1)` benchmark alert — this is runner noise, not a real regression.

`benchmark/sieve/sieve.wado` uses only named imports (`use { Stdout }`, `use { MonotonicClock }`, `use { Benchmark }`) and no namespace imports. This PR only changes namespace-import resolution: the `ns::member → ns$member` canonicalization is a no-op when a module has no namespace imports (`canonical_ns_ref` returns `None` on an empty alias map), and the new registration paths, the deleted special-case branches, and the parser additions are all gated on `ns::`-qualified syntax that `sieve` doesn't contain. So `sieve`'s generated wasm is byte-identical to `main` — the throughput delta is measurement variance (a single benchmark at a single opt level flagged is the usual noise signature).

---

PR #1347, opened 2026-06-10
perf(optimizer): branch-hint-aware WIR optimization and cold-call-site inlining suppression

gfx:
Makes the WIR optimizer branch-hinting aware and uses the same coldness information to improve inlining decisions.

## Branch hints

- `BranchHint` is now a transparent annotation: `WirInstr::peel_hint` / `take_branch_hint` / `hint_condition` helpers, and condition-shape matchers across WIR passes look through the wrapper instead of being blocked by it (const-if folding, branchless increment, init-guard removal). Fixes a latent `Br`-detection gap in `sroa_variant_return`'s operand walk along the way.
- New `wir_optimize/branch_hint.rs`:
  - Trap-based hint inference — an `if` arm that always reaches `unreachable` is hinted cold; a `br_if` whose fall-through always traps is hinted likely-taken. Divergence alone (`br`/`return`) never counts. Covers `assert`/panic arms without needing `cold_path()` markers.
  - `br_if` selection — `if cond { br N }` with an empty else collapses to `br_if N-1`, hint-preserving; the emitter records `metadata.code.branch_hint` entries for `br_if` as well as `if`.
- The synthesized module-init guard and `core:json`'s escape-free string fast path are hinted via `cold_path()` markers.
- A binary-level test pins the hint-offset convention (offsets must point at the `if`/`br_if` opcode relative to the function body) so a drift can't silently disable every hint.

## `-f no-branch-hinting`

For benchmarking A/B: lowers `cold_path()` to a no-op at WIR build (keeping the NIR inliner's cold-cost exclusion identical in both configurations, so the comparison isolates the hints) and skips the inference pass — no hint section is emitted. Wired through `compile` / `run` / `serve` / `dump`.

## Cold-call-site inlining suppression

Calls in a cold context (after a `cold_path()` marker, mirroring the inline cost model's `block_cut`) are no longer inlined unless `#[inline(always)]`: the callee body would bloat the hot caller with code that rarely runs. This is the measurable win of the branch — the parser hot loops no longer carry inlined `String`/`List::grow` bodies:

- json-catalog: **+6.3%** (best-of, 8/8 interleaved rounds faster)
- sqlite-parse: **+2.2%**
- wasm size: Gale highlighter **−7.9%** (469,240 → 432,037 bytes)

Branch hints themselves measured neutral under wasmtime (it only uses them for cold-block layout), which is why the structural use of coldness (inlining) is where the benefit lands; the hint infrastructure positions Wado for engines/future wasmtime versions that consume hints more aggressively.

Also refreshes `benchmark/README.md` and `wasm-size/README.md` (best-of-3 on the current toolchain).

https://claude.ai/code/session_0179KnaHaaGX3TboAqDp2hEd

---
_Generated by [Claude Code](https://claude.ai/code/session_0179KnaHaaGX3TboAqDp2hEd)_

---

<a href="https://app.devin.ai/review/wado-lang/wado/pull/1347" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


gfx:
Investigated the `sieve (-O1)` alert with an interleaved local A/B (release builds, best-of-4 each):

| build | sieve `-O1` best |
|---|---|
| pre-suppression (hints on) | 183.2 ms/iter |
| HEAD-equivalent (hints on) | 182.0 ms/iter |
| HEAD with `-f no-branch-hinting` | 172.3 ms/iter |

So the cold-call-site inlining suppression contributes nothing here (rows 1–2 are equal), and branch-hint emission itself costs ~5% on this one benchmark at `-O1` — the emitted code is byte-identical apart from the `metadata.code.branch_hint` section, so the delta is purely wasmtime's `set_cold_block` layout. The remaining gap to the alert's 1.64x is runner variance; sieve is the most host-sensitive benchmark in the suite (memory-bandwidth-bound; we've measured 69–238 M numbers/s across hosts for the same binary). `-O2`/`-O3` show no alert, and hints measured neutral at `-O2` across the suite during development.

Leaving as-is: a ~5% layout cost on one benchmark at the development opt level doesn't justify gating the advisory hints, and `-f no-branch-hinting` exists precisely to measure/disable them.

---

PR #1367, opened 2026-06-13
refactor(optimizer): retire niri's per-local field_env for the ValueGraph

gfx:
## Summary

Retires niri's per-local `field_env` (`(local, field) → constant`) and hands all struct-field reaching-def to the engine `ValueGraph` (store-load forwarding + hash-cons CSE). The flow-sensitive const-folder now tracks only scalar local lattices; branch arms no longer fork field state (mutable locals are recorded `NonConst` up front). Net ~1000 fewer lines.

## ValueGraph precision fixes (what unblocked the deletion)

Three sound, general improvements closed the cases where `field_env` still out-folded the ValueGraph:

- **Struct-field forwarding through a `Local` producer.** A `let q = f(src)` whose inlined body returns a reassigned `mut` param (`p = S { … }; return p`) now seeds `q`'s field slots from the producing local. Closed `mut_param`, `mut_param_merged`, `ref_1`.
- **Short-circuit precise heap invalidation.** `&&` / `||` now invalidate only the heap its conditionally-run rhs may write, not a blanket `bump_all`, so a read-only aggregate survives a pure field comparison. Closed `coerce_int_3`.
- **Reflexive equality.** `x == x → true` / `x != x → false` for operands sharing a `ValueId` (excluding floats and `v128`, whose equality is not reflexively `true`). Folds identity reinterpret casts (`v as SameType == v`).

## v128 comparison fix

The reflexive fold initially mis-folded `v128` (`f32x4` etc.); excluded it. Investigating that surfaced a pre-existing bug: scalar `==` / `!=` / ordering on `v128` lowered to `i32.eq`, emitting invalid core Wasm (`type mismatch: expected i32, found v128`) — an ICE at every level. The elaborator now rejects scalar comparison on `v128` (`binop_operand_requires_trait`), directing users to the lane-wise `core:simd` comparisons; regression fixture `simd_v128_no_scalar_eq.wado` covers the diagnostic.

## Residual

`int128_cast_to_primitives` keeps one `max as i128 == neg` reinterpret assert (+36 bytes, code-size only): its source struct-literal seed and the consuming reinterpret never coexist in one per-build ValueGraph. Closing it needs the Phase-2 `mod_ref.rs` cross-pass heap model, tracked there. No fixture asserts on its output.

## Also

Comment pass over the touched files: drop change-history / `Stage N` / WEP-process narration, fix stale references to removed symbols, and state current behavior concisely.

https://claude.ai/code/session_01QRpsWY6x97PdPwzuz8da2F

---
_Generated by [Claude Code](https://claude.ai/code/session_01QRpsWY6x97PdPwzuz8da2F)_


gfx:
## Benchmark check: false positive (CI runner noise)

The Benchmark check failed on a `sieve -O3` alert (~2.05× slower). This is **not a code regression** — it's runner-to-runner timing noise.

Evidence: I dumped the final WIR for `benchmark/sieve/sieve.wado` at `-O3` on both `origin/main` and this branch's HEAD and diffed them — **byte-identical** (0 diff lines). This branch does not touch sieve's compiled output, so any throughput difference is purely the runner's wall-clock variance, which the 1.50 threshold flags on noisy runs.

No action needed on the code. The remaining `Test (Gale O3)` job is still running; all other checks are green.

---

PR #1542, opened 2026-07-10
perf(optimizer): hoist inline `&`-literal call arguments to shared globals

gfx:
## Summary

Speeds up `benchmark-json-twitter` by extending `const_object_globalization` to hoist a constant aggregate literal referenced via `&` directly at a call site (e.g. a synthesized `serde` field key, `st.field(&"id_str", …)`), not just a `let`-bound one. Previously such a literal was rebuilt on every call; it's now hoisted into a shared immutable global, matching what already happened for `let`-bound constants.

Best-of-3 A/B on `json_twitter`: serialize 268.82 MB/s vs. 254.76 MB/s baseline (+5.5%), deserialize 97.67 MB/s vs. 93.59 MB/s baseline (+4.4%).

## What changed

- `const_object_globalization`: a single exhaustive walk now collects both the whole-`let`-binding shape and the inline-`&`-literal shape, replacing the previous `Candidate` enum (now `Candidate { func_idx, ty, module_source, kind: CandidateKind }`) and fixing a reachability gap where a qualifying `let` nested in an expression-position block (e.g. a `match` arm) could be missed by both collection paths.
- `NirGlobal::prefer_fixed_string_repr`: a structural field (not a name-prefix guess) marking a global hoisted from the inline-literal case, so `wir_build` can give just that global's value a size-bounded (`INLINE_REF_EAGER_MAX_BYTES`, 64 bytes) override of the shared `string_inline_max_bytes` threshold — needed so `wir_optimize::const_global` can prove the redundant re-assignment at the call site constant and drop it.
- `wir_optimize::prune_dead_data`: drops a passive data segment `register_literal_data` speculatively registers for a literal that ends up wholly `array.new_fixed` (bounded-eager) instead of `array.new_data`.
- `wir_build::register_globals` now asserts a `NirGlobal` never has a WASI `module_source` (a WASI binding file's only candidates are CM-binding glue, already excluded), replacing what was previously an unregistered global silently resolving to Wasm global index 0.
- `tests/dedupe_const_globals.rs`: a count-based regression test proving `dedupe_const_globals` actually merges duplicate globals — the existing fixture's substring-only assertions couldn't detect the pass being disabled.
- `codegen_flags::ships_assert_diagnostic` is now representation-agnostic (checks both `array.new_data` and the flat-form `array.new_fixed` a promoted-eager global prints), since the diagnostic template can now be hoisted like any other `&`-literal call argument.


gfx:
## Benchmark alert is comparison-baseline noise, not a regression from this PR

The alert above (comparing `0bd9da957` against `d911743604`, the unrelated `#1541` merge commit) flags **34 of 34** benchmarks as regressed — including `mandelbrot`, `fts`, `zlib`, `sqlite_parse`, and `syntax_highlight`, none of which this PR's `const_object_globalization` change can affect (it only touches how constant aggregate literals are hoisted for `serde`/JSON-style call sites). Every benchmark moves by roughly the same 1.5–2.3x ratio at every `-Ox` level, which is the signature of a noisy shared CI runner, not a targeted code regression.

Checking the recorded history on `gh-pages` (`benchmarks/runtime-throughput/data.js`) confirms it — `d911743604` and its immediate predecessor `bb0ee38a11` are themselves outliers against the trend of the preceding several days:

| commit | json/twitter/de (-O2) | mandelbrot (-O1) | fts (-O2) | sqlite_parse (-O2) |
|---|---|---|---|---|
| 2026-07-07 .. 07-09 (7 runs) | 98–122 MB/s | 4.3–5.3 M px/s | 9.3–10.2 M conv/s | 4.7–6.3 MB/s |
| `bb0ee38a11` (this PR's base) | 159.57 | 6.24 | 12.41 | 7.78 |
| `d911743604` (used as "Previous" above) | **203.21** | **7.12** | **15.87** | **10.67** |

This PR's own numbers (e.g. `json/twitter/de (-O2)`: 97.86 MB/s) fall squarely within the normal multi-day range — it's `d911743604`'s recorded data that's the anomaly, likely from landing on an unusually fast/quiet runner. The intended improvement from this PR (`json_twitter` ser/de, see PR description) was validated separately via local best-of-3 A/B against the actual pre-PR baseline, not the CI benchmark comparison.

No code changes needed for this alert. It should resolve once `main` records a new, more typical benchmark run to compare against.


---

PR #1619, opened 2026-07-18
feat!: switch template interpolation from `{expr}` to `${expr}`

gfx:
## Summary

Template string interpolation moves from Rust-style `{expr}` to ES/TypeScript-style `${expr}`. Under the new syntax **braces are literal by default** — only `${` opens an interpolation — so JSON, CSS, and other brace-heavy content embed in template strings without escaping. Format specifiers are unchanged (`${expr:spec}`).

```wado
let greeting = `Hello, ${name}!`;
let json = `{"key": "${value}"}`;   // braces literal — no escaping
let hex = `${255:x}`;               // specifiers unchanged
```

This resolves the friction where `{}` collided with JSON literals (Wado's literal syntax is a JSON superset), consistent with the language's existing decision to prioritise JSON compatibility over Rust familiarity (e.g. `[a, b]` tuples).

## Language / compiler

- **Lexer**: `${` triggers interpolation; bare `{` / `}` are literal text; `\$` escapes a literal `$` before `{`. `\{` / `\}` remain accepted.
- **Escape resolution**, **formatter** (emits `${expr}`), and the **editor-grammar generator** updated; `wado.tmLanguage.json` regenerated.

## Migration (all sources)

- Every `.wado` template literal migrated (byte-exact, via the real lexer so nested templates and expression braces are handled).
- `.g4` grammar action bodies, Gale codegen emitters (`parser_gen`, `cst_gen`, `java2wado`), and `example/Wado.g4` updated so generated parsers interpolate.
- Doc comments in the standard library and packages migrated, and `stdlib-*.md` regenerated (these are generated from doc comments).
- ANTLR predicate syntax (`{cond}?`), FIRST/FOLLOW sets, struct/kv literals, imports, and WIT case-names are deliberately left as `{...}` — they are not Wado templates.
- The literal-`$` sigil used in Gale (`$name` attribute refs) is preserved as `$${name}`.

## Docs

`docs/spec.md`, `docs/cheatsheet.md`, and the template WEPs updated (each carries a one-line migration note). Cross-language comparison tables keep `{...}` for the Rust/Python/Go columns.

## Breaking change

Existing `.wado` sources using `` `{expr}` `` interpolation must migrate to `` `${expr}` ``. All in-tree sources are migrated in this branch.

https://claude.ai/code/session_01Y2HGpCjooYCeUX1rCD6qT3

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y2HGpCjooYCeUX1rCD6qT3)_


gfx:
This benchmark alert looks like CI-runner variance, not a real regression:

- This PR is a **frontend-only** change (lexer / formatter / template syntax). The desugaring of a template is unchanged, so the compiled Wasm is effectively identical — the golden diffs showed only the assert/panic *source-text* constants growing by one byte per interpolation (the `$`), which is off the hot paths.
- `mandelbrot` is pure pixel math with **no template strings at all**, yet it shows the same ~1.51× — its compiled output cannot be affected by this PR.
- The ratio is nearly uniform across all 20 benchmarks (1.51–1.66), which is the signature of a slower measurement run rather than a code regression (a real regression would hit specific workloads unevenly).

No code change is warranted. Worth re-running the benchmark job to confirm the numbers return to baseline; happy to dig in if a rerun still shows a divergence.

---

PR #1674, opened 2026-07-24
fix(gale): terminate scan on mandatory-group mismatch; reject hidden/indirect left recursion

### LINKED ISSUE #1671: gale-highlight: parser recurses unboundedly (stack overflow) on invalid Wado token input
## Summary

`wado-lang:gale-highlight-wado`'s `highlight()` — driven by `package-gale`'s generated Wado parser — enters **unbounded recursion** and traps with `wasm trap: call stack exhausted` on certain invalid Wado inputs. A parser must terminate on *any* input (error / recover); it must never loop. Raising the guest wasm stack to **500 MB** does not help, so this is non-terminating recursion, not merely deep.

## Minimal repro

`highlight()` on this 11-line snippet — the operator-reference block from `docs/cheatsheet.md`, which is a token cheatsheet, **not compilable Wado** — traps:

```
// Arithmetic
+ - * / %

// Comparison (can be chained: a < b < c → a < b && b < c)
== != < <= > >=

// Logical
&& || !

// Bitwise
& | ^ ~ << >>
```

Program (run with `wado run`):

```wado
use { highlight } from "wado-lang:gale-highlight-wado"; // or "../src/lib.wado" in-repo
use { print, Stdout } from "core:cli";

global SRC: String = #include_str("./repro.txt"); // the 11 lines above

export fn run() with Stdout {
    let _ = highlight(SRC); // -> wasm trap: call stack exhausted
}
```

The original trigger is the full operator block in `docs/cheatsheet.md` (a ` ```wado ` fence that is a token reference, not code).

## Findings

- **gale-direct** — reproduced calling `highlight()` alone, with no Marl / Component-Model / provider involved. A pure gale parser bug.
- **Unbounded, not deep** — a 500 MB `max_wasm_stack` still traps, so a stack bump is not a fix.
- **Cumulative** — individual operator lines (`+ - * / %`, `== != < <= > >=`, `& | ^ ~ << >>`) and operator-lines-only concatenations parse fine; the comment-interleaved 11-line prefix is the smallest failing input found. Adding line 11 (`& | ^ ~ << >>`) to lines 1–10 is the tipping point, so the comments interleaved with the operator runs matter.
- Most likely the generated parser's error path / ATN prediction recursing **without consuming input** on a run of unexpected tokens.

## Impact

Found while wiring gale into the site (`wado-lang.github.io`) to syntax-highlight ` ```wado ` blocks: `cheatsheet.md`'s operator block crashes the whole build. Docs routinely contain Wado *fragments*, so any highlight-over-untrusted-fragment use (the blog/docs build, and potentially the playground) can be driven into a trap. A guest trap cannot be caught by the caller, so the fix must live in gale — the parser should error/recover on unexpected token runs, never loop.

## Related

#1246 — "a generated parser should never trap on valid input." This is the invalid-input analogue: it must not *loop* on invalid input either.

gfx:
## Summary

package-gale's generated parser could recurse unboundedly (`wasm trap: call stack exhausted`) on invalid Wado input — e.g. the operator-reference block in `docs/cheatsheet.md`, which crashed `gale-highlight-wado`'s `highlight()`.

**Root cause.** A mandatory scan group was emitted `lenient` — bailing out at the entry position on no match — which is sound only for a *nullable* group. Closure's non-nullable `('||' | '|' … '|')` prefix could match zero tokens in the scan and fall into its body `expression`; since `|` ∈ FIRST(expression) via `primary : closure`, `scan_expression` re-entered `scan_closure` at the same position and never terminated.

## Changes

- **lower** — `ScanGroupElement.lenient` now tracks group nullability, so a mandatory non-nullable group fails instead of matching zero. The scan cycle is broken with a fast static scan (no ATN on the hot path), and grammars without such a group stay byte-identical.
- **parser_gen** — the first-set gate is kept for strict groups too; a gate miss on a mandatory group fails. Dropping it mispredicts the longest-match tournament and diverges scan from parse (RustParser's `crate : item*` loop grew the tree unbounded).
- **ir / g4** — `check_left_recursion` rejects hidden (`a : x? a`) and indirect (`a : b …; b : a …`) left recursion at grammar-check time. Gale, like ANTLR4, lowers only *direct* left recursion (precedence climbing); the residual nullable-prefix cycles would still generate a non-terminating scan, so they are rejected loudly — making "codegen never emits a non-terminating scan" a checked property rather than routing to ATN (which would collapse to 117×-slower full-context LL).
- Min repro `tests/grammars/scan_group_recursion.g4` + `tests/driver_cst_scan_group_recursion_test.wado`; left-recursion policy tests in `src/g4/parser_test.wado`; `antlr4-compatibility.md` soundness invariant #5. Regenerated goldens.

Closes #1671.

https://claude.ai/code/session_01GTJSbo22Scdx3QThS5dXry

---
_Generated by [Claude Code](https://claude.ai/code/session_01GTJSbo22Scdx3QThS5dXry)_


gfx:
The Benchmark check itself passed (green) — this is `github-action-benchmark`'s threshold comment, and it reflects **CI-runner variance for this run, not a regression from this PR**.

The reported ~1.5–1.8× slowdown is uniform across benchmarks this change cannot affect: `fts`, `zlib`, and the `json/*` rows are pure compute/serialization and don't use gale, while this PR is entirely within `package-gale` (the parser generator). A parser-generator change can't slow float-to-string or JSON (de)serialization, so a whole-suite uniform slowdown is the signature of a slower runner. The gale-using rows (`sqlite_parse`, `syntax_highlight`, `gale_gen`) show the **same** ~1.7× factor rather than a larger one, so there's no measurable overhead beyond that baseline. Locally the SQLite parse benchmark runs clean at the same throughput as `main`.

---

PR #1675, opened 2026-07-24
feat(reflect): Inspect over the Reflect blanket; Member/EnumCase/FlagBit tokens; typed method receivers

gfx:
## Reflection tokens

- `Member` on `VariantCase` (plus `is_unit()`), and `EnumCase` / `FlagBit` token walks — enums and flags now expose the same member-token surface as structs and variants.

## `Inspect` via the `Reflect` blanket

- Non-generic structs derive `Inspect` through a library blanket in `core:prelude/traits` — an `impl` of `Inspect` for `T` bounded on `T: Reflect` whose `Fields` associated type is a tuple of `Inspect` fields — replacing bespoke per-type `Inspect` synthesis. The derivation lives in the stdlib and the compiler's special-case surface shrinks.
- Blanket dispatch is reified into a single `BlanketImpl` descriptor (module, receiver kind, param, bounds, arity). The former per-query re-derivations (`has_universal_ref_blanket`, `is_reflect_struct_blanket`, `blanket_impl_*_for_trait`) are now selections over it, and dispatch decisions read structured metadata rather than parsing mangled names.
- The universal reference blanket — `Inspect` for a reference to any `T` — is isolated from concrete / shape ref impls: a generic `&T` formats transparently as its pointee, a concrete `&x` keeps the `&` prefix, and reference struct fields route through the blanket correctly.

## Metadata-driven method receivers

- `LocalMethodName` carries a typed `Receiver` (`Type` / `Ref(RefKind)` / `Projection`) instead of a `base_struct_name` string. Mangling is one-way (metadata → string); nothing parses a mangled name to recover the receiver shape.
- `Receiver::classify` — the last string → shape boundary — is removed. Producers build the typed receiver from the AST / resolved types at the source (`ImplFacts.receiver`, the ref-blanket shortcuts, `decompose_type_for_method_name`).
- `TraitEnv`'s `impl_index` / `all_impl_index` are keyed by the typed `Receiver`; every lookup builds its key from a typed source, so no `&`-string is ever matched to classify a receiver. The `&` / `&mut` spelling now lives only in `RefKind::prefix` / `Receiver::mangle_with_ref`.

Implements WEP 2026-06-13 items: `Member` on `VariantCase`, `EnumCase` / `FlagBit` tokens, and the struct-`Inspect`-over-`Reflect`-blanket migration.

https://claude.ai/code/session_01L9DrZZ1CmPmEuw3WroZZue


gfx:
The `Benchmark` check itself passed (conclusion: success) — this is `github-action-benchmark`'s comment-on-alert, not a failing gate.

I don't believe this is a real regression. The slowdown is uniform (~1.5–1.8×) across **every** suite, including `zlib/decompress`, `fts`, and `sqlite_parse` — workloads whose hot paths share no code with this PR. The changes here are all compile-time (name mangling, trait-index keying, reflection derivation); they don't alter the generated guest wasm for decompression or text search. A real regression would move the reflection/serde suites and leave those untouched — not scale everything by the same factor. A uniform cross-suite slowdown measured against a baseline captured on a different run is the signature of runner variance, not code.

Re-running the Benchmark workflow (or the next push) should re-measure on a fresh runner and clear the alert. If a specific suite stays red on a clean re-run, I'll profile it.

---

PR #1676, opened 2026-07-24
fix(cli): surface wado test compile/load failures in the end-of-run summary

### LINKED ISSUE #1666: CI: `Test (Gale O1)` / `Test (Gale O3)` still flakily fail with "1 compile failed" (OOM recurrence of #1648)
## Summary

The per-optlevel `package-gale` CI jobs still intermittently fail with a single compile failure out of ~1360 files while every test that compiles passes. This is a recurrence of #1648 (closed 2026-07-21 as completed) — the underlying resource/OOM flakiness is not resolved.

Observed on PR #1664, run `29958097014`, head `3f135bc5`:

- **`Test (Gale O1)`** (job `89055556317`): `compile: 1359 ok, 1 failed`, `test: 2158 passed, 0 failed`, then `[test-gale-o1] ERROR task failed` / exit 1.
- **`Test (Gale O3)`** (job `89055556186`): same shape.
- **`Test (Gale O2)` passed** on the same head.

## Evidence it is the same OOM class, not a compiler bug

- **Memory ceiling reached.** The O1 job log shows RSS climbing to `peak 2106 MiB` against a `2106 MiB` limit — the runner is at its memory ceiling while `wado test` compiles the large fixtures (`cst_sqlite`, `cst_rust`, `cst_typescript`, `kiln-consumer` — the multi-second, high-RSS compiles) in parallel.
- **Does not reproduce on roomier machines / lighter levels**, exactly as in #1648. The single failure is always a *compile* failure (never a test failure), on 1 of ~1360 files, non-deterministic.
- **No crash signature** in the logs — no ICE, panic, `type mismatch`, `invalid core Wasm module`, or the NIR-engine double-parent tripwire. Just `task failed` / exit 1 after one file's compile is killed.
- **All E2E levels pass** on the branch locally (full O0/O1/O2/O3/Os matrix, 9467/0), and the O2 golden regeneration over the whole fixture corpus is clean, so the O1/O3 optimizer output itself is fine.

The only difference from #1648 is *which* opt-level jobs happened to lose the memory race this time (O1+O3 rather than O2+O3); the mechanism is identical.

## Suggested remediation (from #1648, none yet applied)

- Bump runner memory for the `test-gale-o{1,2,3}` jobs, **or**
- Cap `wado test` compile concurrency for those jobs (fewer parallel compiles → lower peak RSS), **or**
- Profile peak compiler RSS at `-O2`/`-O3` on the largest `package-gale` fixtures to find a reducible optimizer memory spike.

Capping compile concurrency is the most robust de-flake since it removes the parallelism-driven RSS spike rather than just raising the ceiling.

## References

- Prior occurrence: #1648.
- This occurrence: run `29958097014` — `Test (Gale O1)` job `89055556317`, `Test (Gale O3)` job `89055556186` (PR #1664).

gfx:
## Summary

`wado test` reported a compile/load failure only through a live per-file notice, printed the instant the file failed. On a long run (`package-gale` is ~1360 files) that notice scrolls off the top of the log, and GitHub retains only the tail — so a `compile: N ok, 1 failed` summary was left with no pointer to *which* file failed or *why*. The compiler-diagnostic failure path made it worse by discarding the message entirely (`detail: None`), so even the live notice carried no reason.

This branch captures the failure reason and re-lists every failed file — with that reason — after the summary, so the recap is the last thing printed and survives tail-only log retention.

## Changes

- `compile`: `try_compile_with_kiln_cache` returns a `TryCompileError` carrying the rendered error diagnostics; `try_compile` stays a thin wrapper mapping back to `CompileFailure` for callers that only need the pass/fail outcome.
- `test`: `CompileFailure` gains a `message`; a shared `failure_recap_lines` renders the section for verbose / `--no-run` and the end-of-run recap alike.
- `reporters`: heartbeat and TAP accumulate failures and re-list them in `on_run_done` (TAP as `#` comments, keeping the document valid).

## Context

Investigated for #1666. The `Test (Gale Ox)` job read as a flaky OOM, but the log shows RSS peaking at 2106 MiB on a 16 GB `ubuntu-latest` runner and the process surviving to print the full summary — inconsistent with a kernel OOM-kill. The real signal was a non-deterministic single-file compile `Err`, whose message was unrecoverable from the log. This PR makes that reason visible in the log tail so the next occurrence can be diagnosed on evidence.

This does **not** close #1666 — the underlying non-deterministic failure is still to be root-caused once a captured message is in hand.

https://claude.ai/code/session_01PospSTEfJ8tKVFWLAfNnuB

---
_Generated by [Claude Code](https://claude.ai/code/session_01PospSTEfJ8tKVFWLAfNnuB)_

### DIFF
--- a/package-gale/tests/generated/action_arg_pred/wado_arg_pred.g4.kiln.json
+++ b/package-gale/tests/generated/action_arg_pred/wado_arg_pred.g4.kiln.json
@@ -2,7 +2,7 @@
   "version": 2,
   "invocation": "kiln-e80f819cb07676d0",
   "generator": "local:src/generator.wado",
-  "generator_source_hash": "307f741962d612bb705c84ed6d04c9cf6b5b87d91bafa66d74dac3ccf3b83e94",
+  "generator_source_hash": "1e9a33c2ab35fc579f4062891f3052ef0b79d933d5147968637a9d60dde7a57f",
   "primary": {
     "path": "tests/grammars/wado_arg_pred.g4",
     "hash": "b6e9fd26f1e4936b34f10ed2108b0fad8b57c16b49f42707b509a9966ec692bc"
--- a/package-gale/tests/generated/action_atn_pred/wado_atn_pred.g4.kiln.json
+++ b/package-gale/tests/generated/action_atn_pred/wado_atn_pred.g4.kiln.json
@@ -2,7 +2,7 @@
   "version": 2,
   "invocation": "kiln-873f523c716cd4d1",
   "generator": "local:src/generator.wado",
-  "generator_source_hash": "307f741962d612bb705c84ed6d04c9cf6b5b87d91bafa66d74dac3ccf3b83e94",
+  "generator_source_hash": "1e9a33c2ab35fc579f4062891f3052ef0b79d933d5147968637a9d60dde7a57f",
   "primary": {
     "path": "tests/grammars/wado_atn_pred.g4",
     "hash": "418c3fffe47dbad329d0442fa2f8777413a3fd4fed06493c857d0d38f0f020a1"
--- a/package-gale/tests/generated/action_camel/wado_camel_label.g4.kiln.json
+++ b/package-gale/tests/generated/action_camel/wado_camel_label.g4.kiln.json
@@ -2,7 +2,7 @@
   "version": 2,
   "invocation": "kiln-e3b09d9b3841e994",
   "generator": "local:src/generator.wado",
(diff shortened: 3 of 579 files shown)


gfx:
This benchmark alert is spurious for this PR. The diff is confined to `wado-cli` (test-runner output and compile-error plumbing); it does not touch `wado-compiler`, and the successful-compile path returns the compiler's output unchanged — the generated guest Wasm is byte-identical. A change that cannot alter guest codegen cannot alter `json/twitter/de` guest throughput, so the 1.53 ratio (barely over the 1.50 threshold, single-sample microbench) is run-to-run variance, not a regression from this branch.

The `Benchmark` check itself concluded `success`; this is a comment-only alert and does not block merge.

---

PR #1714, opened 2026-07-29
feat(niri): fill a container at compile time

gfx:
A container built and filled at compile time is a compile-time value. `push` and
the growth it triggers run in the frame, and one whose elements are bytes leaves
the engine as the literal a source string lowers to, so a helper that returns a
`List<u8>` never reaches run time.

## The frame writes

A compile-time frame performs each statement once, over values it built itself.
It can now write to them: a field store, an element write, an allocation, a
copy. A call writing through a `&mut` parameter runs and writes back into the
caller's place — which is the only thing running `push` produces, since it
returns nothing.

The write-back is confined to statement and `let` position, where a call runs
exactly once. The lattice projection is re-entrant, so a mutating call is
refused there: a write applied twice is worse than a fold missed.

Whether a call can run is decided by whether the frame can run its body, kept
apart from whether its result may replace the call. A unit callee runs; its
value never substitutes. `stores` does not gate a run, since the engine binds a
referent's value rather than a reference, and `Reached` holds the one case that
would make that snapshot wrong.

## Trackability divides on reach, not position

A shared receiver, a by-value argument and a builtin's source cannot be written
through, so they count wherever they appear — that is what carries a container
through the `&self` reads `push` makes of its own capacity. A `&mut` one counts
only where a frame performs the write, since a walk that performs nothing and
steps over a write leaves its target holding a value the program never produced.

Both analyses scan the reachable body only: the arena keeps every node an
in-place rewrite displaced, and one nothing refers to cannot run. Parameter
kinds come from the callee map rather than a borrow, so a fold does not turn on
which function the visitor happens to be walking.

## Builtins

`array_new` denotes a zero-filled sequence, `array_copy` splices one, `select`
folds when both arms are constant — the arm not taken has to compute rather than
trap — and `cold_path` is stepped past, which matters because it sits on the
path that grows a container. The map reaches every interpreter the optimizer
builds, including the peephole session and the ones that read globals.

A container literal is written only over the container the registry names.
Shape does not identify one: a struct over an array and an `i32` looks the same
and means something else.

## The engine is seven modules

`niri` keeps the interpreter, the lattice and the maps it is configured with.
Under it, each module answers one question: what an expression denotes
(`lattice`), what running a body does (`frame`), what becomes of an expression
once its value is known (`rewrite`), which locals a walk may hold a value for
(`trackability`), whether a pattern matches (`pattern`), what a borrow names
(`place`). The `&self` / `&mut self` line between reading an expression and
running one is a module boundary.

`const_eval` is published, so the compile-time value model is reached from where
it lives — it is shared by the interpreter, the value graph and the optimizer
alike.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EoWspzheEB3W9iez8BD4ZY)_


gfx:
The benchmark alert is the runner, not this branch. Measured both revisions on one machine, back to back:

| | best of 3 | runs |
| - | - | - |
| `main` (9951dc1) | 4.23 M px/s | 4.08 / 4.23 / 4.14 |
| this branch (32ba318) | 4.14 M px/s | 3.97 / 4.14 / 3.88 |

2% apart, inside run-to-run spread. The alert's baseline of ~7 M px/s comes from a faster runner, and the current ~4.27 from a slower one.

Three things in the alert point the same way. `mandelbrot` is a float loop with no container, string or compile-time-foldable call in it, so nothing this branch touches reaches it. The regression is uniform — 1.5–1.8x across mandelbrot, fts, zlib, json, sqlite_parse and syntax_highlight alike — which is a property of the machine rather than of six unrelated workloads. And `mandelbrot` reads 4.27 / 4.27 / 4.28 at `-O1` / `-O2` / `-O3` where the baseline reads 7.19 / 6.92 / 7.14: the optimization level stops mattering, which is what a machine-bound measurement looks like.

`benchmark/`'s own procedure is to run three times and keep the fastest, since throttling only ever slows things down. CI runs once.

---

PR #1715, opened 2026-07-29
fix(optimizer): make constant folding agree with the runtime on casts and traps

gfx:
Constant folding agrees with the code the compiler generates on two points where
it decided a program's result on its own.

## Float-to-int casts

`translate_cast` emits the trapping `trunc` family over an i32 or i64
intermediate, and reaches a narrower integer by wrapping that result.
`const_eval::float_to_int` models exactly that. It answers `None` — leaving the
cast `NonConst` so the node survives to trap at run time — for a NaN, an
infinity, or a truncation outside the intermediate's range. In range, it
truncates through the intermediate and wraps down to the target.

So `300.7 as i8` folds to 44, `300.7 as u8` folds to 44, and `1.0e30 as i32`
does not fold at all.

## Deleting a condition

Collapsing an `if` or a `match` whose arms are all the same constant deletes the
condition or the scrutinee, so it is gated on `is_discardable`: side-effect-free,
and trap-free on top of that. A `Cast` or a `FieldAccess` is neither — a
float-to-int cast traps on overflow and a field read traps on a null reference —
and deleting either erases a trap the program is entitled to observe.

## niri's shape

Each fact has one home. `AliasInfo` sits with the alias analysis in
`optimize::alias` and `LocalSet` with the arena in `nir_arena`; the pure lattice
projections sit in `niri::lattice`. `niri.rs` carries the engine's own types and
entry points.

The CTFE reduction walk descends through `Body::for_each_child`, so every child
position is reduced — call arguments, index expressions, array literals, pattern
constants — and a node kind added to the IR is walked without a second list to
keep in step. Two positions are handled apart from the generic walk: a `Match`
arm, which reduces under the bindings its own pattern makes, and an `Assign`
target, which names storage rather than a value.

A compile-time frame exchanges its per-local state as a single `FrameState`, so
entering and leaving a call moves every per-local fact together and no member
can be restored out of step with its siblings.

Comments follow the project rules: rationale lives in doc comments and in the
names of the functions it explains, and each rule of the `if` and `match`
collapses is a named function rather than a numbered step.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AsSZiaRihPSJL5Sv4Qp29n)_


gfx:
CI red on `macos-latest` (wado-vscode), failing on the base branch too — not caused by this PR.

The `vscode-test` runner downloads VS Code successfully and then cannot launch it:

```
✔ Downloaded VS Code (07b4ff1) into .vscode-test/vscode-darwin-arm64-1.131.0
✔ Found existing install in .vscode-test/vscode-darwin-arm64-1.131.0
Test error: Error: spawn .../vscode-darwin-arm64-1.131.0/Visual Studio Code.app/Contents/MacOS/Electron ENOENT
Exit code: -2
```

The identical failure is on `main` at `3608e36` — this PR's base — in [run 30442968516](https://github.com/wado-lang/wado/actions/runs/30442968516), which predates the branch. This PR changes only `wado-compiler/`; nothing under `wado-vscode/` is touched. The extension's own unit tests (51) pass in both runs; only the step that spawns the downloaded VS Code fails.

Every other check is green, including E2E at O0/O1/O2/O3/Os, Gale, Core, Wado, and the deps + wasm32 check.

I'll re-run this job once the base branch recovers.

---

PR #1717, opened 2026-07-29
feat(reflect): blanket Inspect derivation and ReflectStruct::from_fields

gfx:
## Reflection

`Inspect` is derived by blanket impls over the `Reflect*` bounds, one per
reflection kind (WEP 2026-06-13):

- `impl<S: ReflectStruct<FieldTypes = [..F]>, ..F: Inspect> Inspect for S`
- `impl<E: ReflectEnum<Members = [..M]>, ..M> Inspect for E` — writes `Name::Case`
- `impl<V: ReflectVariant<CasePayloads = [..P]>, ..P: Inspect> Inspect for V` —
  writes `Name::Case`, with the payload in parens for a non-unit case

A plain declaration reaches `Inspect` through its kind's blanket, so no per-type
impl is synthesized for it.

`ReflectStruct` gains `from_fields(fields: Self::FieldTypes) -> Self`, the build
direction to match the read direction `members()` already provides. Every kind
spells its build direction `from_<channel>`: `from_discriminant` / `from_bits`
return `Option`, since an unknown raw input is a normal deserialize error, while
`from_fields` is total because a field-value tuple is already typed.

Selecting a blanket is receiver-aware. A trait carries several disjoint value
blankets — the four reflection kinds each derive over their own bound — so
`TraitEnv::value_blanket_for_receiver` picks the one whose receiver-param bounds
the receiver satisfies, and the param, pack and module projections all read that
same blanket.

## Name namespaces as types

The compiler answers several distinct questions with a name — mangled identity,
declaration name, struct-list key, template key, instance key — and a name from
one namespace reaching a consumer that keys on another is a silent lookup miss.
`name.rs` makes the namespace part of the type: `MangledName`, `DeclName` and
`DeclPath` carry no `Deref`, `AsRef` or `From<String>`, are minted only by
`FqTypeName` / `TypeTable`, and convert only through named methods. `func_map` is
keyed by `MangledName`, built through `in_module` / `builtin_alias` /
`wasi_import`; the CM interface registry and the impl indices demand `DeclName` /
`DeclPath`, and `DeclPath` is assembled as receiver + method rather than by
formatting a pair of halves.

Supporting changes, described in WEP 2026-07-28 and WEP 2026-07-29:

- `ResolvedType::Struct` carries `decl_name` plus `type_args: Vec<TypeId>`, and
  the rendered spelling is derived by `TypeTable::struct_rendered_name`. Interning
  keys on the rendering, so head and arguments are separately readable without
  changing type identity. `struct_list_name` owns the rendered namespace the
  struct registries are keyed by.
- `LocalMethodName::struct_name` is derived from `receiver` and
  `struct_type_args` rather than stored beside them.
- `MethodOwner` is the single encoding of which type owns a method, so the
  newtype-override question reads a discriminant instead of comparing a
  declaration name against an instantiated one.
- `FqTypeName` / `TypeHead` replace `from_mangled`, `as_str` and `simple_name`,
  so no fq name is parsed back out of a string.

## Fixes that fall out of it

- A reference receiver is spelled by one function for both a definition and a
  lookup, so `&List<i32>` matches a registered ref-impl template. A regression
  test asserts the two spellings agree rather than pinning either output.
- DCE keys a method on its full name including method type arguments, so
  `field<T>` and `field<i32>` are two functions and neither is collected while
  live. Retention derives its rendering from the instantiation a struct records,
  so a name fixed before newtype / flags erasure is compared against one built
  the same way.
- `TypeTable::retain` closes over each surviving struct's `type_args`
  transitively, so a struct never survives spelling itself with an id that no
  longer resolves.
- A static call whose target names no struct-like type — a trait, an undeclared
  name, a turbofish on a non-generic — is reported instead of typing the call
  `unknown` and reaching WIR build.
- A methodful-impl probe keys `(module, name)`, so an impl on one module's
  `Widget` does not suppress synthesis for another module's.
- `wado query` resolves a `Type^Trait::method` symbol against the trait-impl
  members, and `wado doc` renders a variadic pack bound as `..F: Bound`.

`docs/compiler.md` records the method-key format: a key is
`(impl module, declared receiver, trait, method)`, so the impl and declaring
module segments repeat whenever a type is implemented in the module declaring it,
and a receiver with no declaring module has no declaration segment.

WEP 2026-07-29 tracks the remaining steps: `StructListKey` as a type, the same
declaration/instantiation split for `ResolvedType::Newtype`, carrying a struct's
`TypeId` on `TirStruct` / `NirStruct`, and a method key holding the two modules
as separate fields.


gfx:
`macos-latest` failed in the `wado-vscode` integration test, and the cause is the runner, not this diff:

```
Test error: Error: spawn .../vscode-darwin-arm64-1.131.0/Visual Studio Code.app/Contents/MacOS/Electron ENOENT
Exit code: -2
```

`vscode-test` logged `Downloaded VS Code ... into <dir>` and then `Found existing install in <dir>`, and failed to spawn the binary inside that same directory — a truncated download/extract on the macOS runner. All 51 grammar and tokenization unit tests passed before it, and `ubuntu-latest` ran the same `npm test` green.

This branch changes no file under `wado-vscode/` and does not touch `wado-compiler/src/syntax.rs`, so it cannot affect the extension or its generated grammar. `main` is green across its last 10 CI runs, including one 45 minutes before this PR opened, so it is not a pre-existing base-branch failure either.

Re-running the failed job is blocked while `windows-latest` is still in flight in the same workflow run (`403 This workflow is already running`). I will re-run it once that finishes and report back; no code change is warranted for this failure.

---

gfx:
Withdrawing the re-run I said I would do above: `macos-latest` is not a required check and is being fixed in a separate PR, so nothing is pending from me on it.

---

PR #1722, opened 2026-07-31
fix(gale): soundness bugs in char sets, lexer dispatch, and label lowering

gfx:
Clears the `TODO.md` "Soundness and compatibility divergence" entries that mis-parse valid input, plus what fixing them surfaced. Each fix lands with a failing test first.

## Char sets

`\p{...}` resolves against general-category tables generated from the UCD (`scripts/regen-unicode-tables.sh`, 17.0.0), covering all 30 categories, their long aliases and the group letters. `\P{...}` is the complement of that, so both directions are exact — a complement of an approximate table admits real members of the property, which is the failure mode a hand-maintained table cannot avoid. A property with no table (a script, a block, a binary property) is rejected naming it, rather than expanding to a set that matches nothing.

A char-range endpoint is decoded as a code point and clipped to the Unicode scalars. ANTLR4 matches UTF-16 code units, so a surrogate endpoint is legal there; Gale matches decoded scalars, which no input character equals, so `[\uD800-\uDFFF]` matches nothing. A range whose endpoints are both scalars keeps its single-range shape even when it spans the surrogate block. `'\uD800'..'\uDBFF'` reads its endpoints from source, since the token text cannot carry them.

## Lexer dispatch

A range branch carries every rule whose first set meets the range. The branches are an `if / else if` chain, so a char in two rules' overlapping ranges only ever reaches the first branch covering it, and the rules can diverge after their first char. A rule reachable only from a single-char branch stays out — those come first, so listing it in a range branch would be a `try_` that always fails.

Guards are char patterns (`first_char matches { 'a' | 'x'..='z' }`). Adjacent branches with identical call lists fold into one guard, which a large first set needs: `[\p{L}]` is ~2000 ranges. `docs/compiler.md` records why the pattern form rather than a disjunction — a `||` chain that wide overflows the compiler's stack.

## Alternatives and labels

`.` and `~X` both carry the empty static FIRST set, so an alternative opening on either is open-ended and routes through the scan tournament rather than a kind-check that commits to a concrete alt. Neither a label nor a `( … )` wrapper narrows what an element matches, so `w=.`, `x=(~X)` and `x=(.)` reach the same dispatch. At the lowered-IR level such an alt no longer gets an unreachable empty-token branch, and the overlap advisory names an open-ended alternative as the cause instead of printing an empty shared set.

`unwrap_labeled_transparent_groups` strips the syntactic wrapper off a labelled element before the stamping passes, so `x=(ID)` binds what `x=ID` binds and `x+=(ID)+` repeats what `x+=ID+` repeats. The pass runs on the `gale dump` path too, so the report matches what the emitter sees.

## Tooling

`scripts/gen_unicode_tables.wado` regenerates the tables, emitting through `src/wadopoet.wado` (`CodeWriter`, `FnSpec`, `GlobalSpec`, `str_lit`) and driven by `scripts/regen-unicode-tables.sh`, matching the other Wado scripts in that directory.

`TODO.md` keeps the two prediction-path entries with what `gale dump` reports for each, and gains the two gaps this work surfaced: `\p{...}` outside the general categories, and a lexer rule that cannot match at its first char never reaching a catch-all. `perf.md` records that the dispatch chain is linear in the ranges.

---
_Generated by [Claude Code](https://claude.ai/code/session_015pm4TmrYim6yKztBF7ssf8)_


gfx:
The `gale_gen` benchmark alert needs a decision rather than another patch, so flagging it here.

The verdict flipped from pass to fail between `14f28ffa0` and `ee4558760` — and `ee4558760` is the bot's `chore: tidy`, 625 `.kiln.json` generator-hash lines and no code. Same generated parsers, same compiler. So `gale_gen` sits on the 1.50 threshold and run-to-run noise crosses it; the first alert reported the ratio as exactly `1.50` too.

**Where the cost is.** `RustLexer.g4` opens `XID_Start` / `XID_Continue` / `WHITESPACE` on `\p{L}\p{Nl}`, `\p{Mn}\p{Mc}\p{Nd}\p{Pc}` and `\p{Zs}`. Against the real general-category assignment those are ~2000 first-char ranges where the previous hand-written table gave ~30. Everything keyed on the first-char set scales with that: the FIRST-set walk, the dispatch-group build, and the branch coalescing. The emitted dispatch is *smaller* than before (56 branches / 325 `try_` calls against 73 / 454), so this is generator CPU, not output size.

`14f28ffa0` already took the quadratic factors out of the dispatch build — the per-range `has_pattern` scan is gone and a range wider than the whole single-char branch set answers by width instead of scanning. What is left is proportional to the range count itself, which is what the correctness fix is.

**The trade.** `\p{L}` now selects the letters it names, including Hebrew, Thai and the supplementary planes, and `\P{L}` excludes exactly those — a complement over an approximate table admits real members of the property, which is why the table had to become exact. Generating a parser for a grammar that asks for every Unicode letter costs more than generating one for 14 ranges.

**Options, if the current number is not acceptable:**

- Raise the `gale_gen` threshold and record the step in `perf.md`. The regression is a one-time level change, not a trend.
- Bound the dispatch chain properly: it is a linear `if / else if` over the first-char sets, so a sorted interval table with a binary search would make the range count stop mattering for both generation and the generated lexer's own dispatch. Recorded in `perf.md`; it is its own change.
- Keep a coarse table behind a grammar option, accepting that `\P{...}` is then unsound for it.

Happy to take whichever, but I did not want to pick between correctness and the benchmark number on my own.

---

PR #1729, opened 2026-08-01
feat(elaborator): overload resolution — trait-qualified calls and argument-directed selection

gfx:
Implements [WEP: Overload Resolution](docs/wep-2026-07-31-overload-resolution.md): which candidate sets are legal for a method name, how a unique candidate is chosen from types available at the call site, and the explicit syntax for naming one when the compiler cannot.

## Language surface

- Trait-qualified (UFCS) calls: `Trait::method(recv, args…)` names the trait that must answer, and the trait turbofish `Take::< A >::take(recv, …)` additionally pins one argument list. No new grammar — a trait head in the existing static-call shape resolves as a qualified call. The receiver spells its own mode (`&mut x` for `&mut self`, with `&mut` → `&` the one coercion); a mismatched mode is an error rather than a silent copy mutation.
- Argument-directed selection: one trait implemented for one receiver at several argument lists (`impl Take< A >` beside `impl Take< B >`) is Wado's only overload set, and shallow per-argument probe classes select the unique admitted candidate. Literals admit but never select — a literal-only distinction is an ambiguity error whose fix is a cast or the turbofish, so adding an impl can never silently retarget existing calls.
- Cross-trait collisions error everywhere: a method name reachable through two trait declarations is ambiguous in every shape, including the concrete-receiver case. Blanket impls neither count toward the collision nor beat any impl written for the receiver. When same-named declarations from different modules collide, the one in scope at the call site wins; several in scope is the error.
- Conversion calls fold into the same rules: `Wrapper::from(42)` preselects the impl by the literal's probe class before the argument is elaborated (one admitted impl supplies the expected type, several are an ambiguity error), and a non-literal argument distinguishes same-head generic sources (`From< List< i32 > >` vs `From< List< String > >`) by the full argument type.

## Compiler internals

- Trait identity is structural: a `(declaring module, name)` declaration key plus resolved trait-argument types, computed in each impl's own frame — never a spelling. The required-trait filter, overload-set grouping, dedup, and both ambiguity reporters compare identities, so aliases and same-named foreign traits behave uniformly.
- A qualified call records a static dispatch carrying the resolved signature's facts (defaults, `is_mut`, parameter types with the receiver at slot 0). `StaticMethodDispatch::self_in_args` marks the receiver-in-args convention for consumers that zip parameters against arguments (effect-parameter resolution), and fn-param closure specialization keys on value-argument indices for both call spellings.
- The probe is a shallow, side-effect-free scan by design (never a speculative `resolve_expr`); its admissibility table is shared between method selection and the conversion preselect, so integer newtypes admit integer literals in both.
- Diagnostics name every candidate (receivers rendered with their type arguments), state the escape (`42 as i64`, the trait turbofish, `Trait::method(&x)`), and qualify same-named traits by their declaring module.

The WEP records the full design, including the deliberate non-goals (no free-function overloading, no return-type direction, no ranking) and the open associated-function case.

https://claude.ai/code/session_01Lukuc8Vf6dmFhenXCcCYNf


gfx:
Investigated the benchmark alert; no real regression.

- At -O1 and -O2 the guest Wasm this branch's compiler emits for `mandelbrot.wado` is byte-identical to what `main` (9b81c7ce0) emits, so the reported ~1.6× slowdowns there (and the uniform ratios across unrelated suites, including pure-float mandelbrot) can only be runner noise — identical bytes cannot run slower.
- At -O3 there was one genuine difference: `main` const-folds the inlined `count_digits_i64` loop (a cold path in the benchmark's label formatting) and this branch's head did not. That is upstream drift, not a change on this branch — `optimize/const_folding.rs` gained the loop const-eval on `main` after this branch's merge base. Merging `main` (071dd67c0) makes the -O3 output byte-identical to `main`'s as well.

---

PR #1733, opened 2026-08-01
fix(optimizer): sroa_param must not snapshot a boxed reference parameter

### LINKED ISSUE #1730: Miscompilation: `nir/sroa_param` breaks aliasing between a `&mut` and a `&` argument naming the same local
A `&mut T` and a `&T` borrow of the same local, passed to one call, are
scalarized independently by `nir/sroa_param`. The `&T` parameter ends up bound to
the caller's pre-call value, so a read through it does not see a write made
through the `&mut` one. The program's result depends on the optimization level.

## Reproduction

```wado
use { println, Stdout } from "core:cli";

fn touch(a: &mut i32, b: &i32) -> i32 {
    *a = 1;
    return *b;
}

export fn run() with Stdout {
    let mut x = 0;
    let got = touch(&mut x, &x);
    println(`got=${got} x=${x}`);
}
```

| | output |
| --- | --- |
| `-O0` | `got=1 x=1` |
| `-O1` | `got=0 x=1` |
| `-O2` | `got=0 x=1` |
| `-O3` | `got=0 x=1` |
| `-Os` | `got=0 x=1` |

`-O0` is the correct answer. References are the exception to Wado's value
semantics, so `*b` reads the current value of `x`, which the write through `a`
has already set to `1`. Note that `x` itself is `1` at every level — only the
read through `b` disagrees, so the write does happen and it is the second
parameter's binding that is stale.

## The responsible pass

Skipping one pass restores the correct result:

```
$ WADO_SKIP_PASS=nir/sroa_param wado run -O2 repro.wado
got=1 x=1
```

Every other pass in the `-O2` pipeline was skipped individually and none of them
changes the outcome — `nir/sroa_param` is the only one.

`&mut i32` lowers to a by-value `Box<i32>` (see the note in
`tests/fixtures/bug_store_load_forward_mut_call_arg.wado`), and after
`prepare_types` a shared borrow of a primitive is carried in that same boxed
form. Both parameters therefore arrive as boxes over the same storage, and
`sroa_param` scalarizes each one as if it were the only handle on it.

## Notes

- Reproduces on `main` at 2a556c019.
- Pre-existing; not introduced by #1728. It was found while reviewing that PR,
  whose changes are confined to `niri` / `const_eval` / `nir_visitor` /
  `value_copy_demote` and do not touch `sroa_param`. It also reproduces with
  that PR's `niri` aliasing check restored to its pre-PR form, and
  `niri::aliased_write_targets` does not fire on this program in either shape —
  the call is not being const-folded, so the wrong answer comes from the
  optimized runtime code rather than from CTFE.
- Wado has no borrow checker, so `touch(&mut x, &x)` is ordinary source rather
  than something the front end rejects.
- A fixture for `wado-compiler/tests/fixtures/` can be the program above with
  `__DATA__` of `{"stdout": "got=1 x=1\n"}`; it fails at every level except
  `-O0`.

gfx:
`boxing::prepare_types` collapses `&T` and `&mut T` onto one by-value `Box`
wrapper struct, and `&x` of an address-taken local lowers to a read of that
single box. A boxed reference parameter is therefore the caller's storage rather
than a copy of it, so `sroa_param` refuses to scalarize one whose box any other
access path the callee holds can write.

`touch(&mut x, &x)` hands the same box to both parameters, and `*b` reads the
write made through `a` at every optimization level.

## Aliasing model

`param_snapshot_unsound` rejects a candidate on two grounds.

`ReachableWrites` covers writes the type walk cannot attribute to a name: an
aliasing global, or anything behind an unresolved indirect call. `Opaque` is the
pass's only account of closure captures — a `fn mut()` value carries them in a
functor struct its *type* does not mention, and invoking it is the only way to
reach them — so every query over it answers conservatively.

`mut_reachable_contains` covers a sibling parameter from which a write can reach
the wrapper struct. The sibling need not be the handle itself, only carry one.
Writability is a property of the location, not of the walk's depth: a `&mut`
pointee and a mutable box are writable wherever they appear, a shared pointee
never is, and a component of a writable aggregate inherits it. So a by-value
`Holder { r: &mut i32 }` is a copy whose `r` still writes the caller's box, while
a by-value `M` writes nothing.

Two read-only handles cannot stale each other, so `f(&a, &b)` on a pair of shared
boxes still scalarizes to `fn(i32, i32)` with no allocation at the call site.

## Type table

A shared and a mutable borrow of the same payload share one wrapper struct after
the rewrite, so `TypeTable` records which wrappers came from a shared `&T` and
answers `is_mut_box`. The reference-representation WEP makes that record
normative: a consumer asking whether a boxed handle can be written through
consults it rather than the resolved type.

## Fixture

`opt_sroa_param_boxed_alias.wado` hides the aliasing handle one level further in
each case — a `&mut` sibling, two `&mut` handles, a by-value struct field, a
shared `&Holder`, a global, and a `fn mut()` capture. `shared_pair` is the
precision guard, with `wir_expect` pinning its scalarized signature so an
over-conservative guard cannot pass unnoticed.

Closes #1730

https://claude.ai/code/session_01AUNd6sdfR4e8CVJv5wP1Fd


gfx:
The alert is not caused by this branch — it emits **byte-identical wasm** for this benchmark.

I rebuilt the compiler with `origin/main`'s `optimize/sroa_param.rs` dropped into this branch's tree, and compiled `benchmark/json_twitter/json_twitter.wado -O2` with each:

```
882b5905283f6d72b4c4ec45c94694f90c207d6bf3b4e10c23d336faa3cd8f67  base.wasm   (main's pass)
882b5905283f6d72b4c4ec45c94694f90c207d6bf3b4e10c23d336faa3cd8f67  head.wasm   (this branch)
```

Same digest, same 170844 bytes; the `-O2` WIR dumps are identical too (20121 lines, clean `diff`). The comparison is meaningful rather than vacuous: `WADO_SKIP_PASS=nir/sroa_param` *does* change the output for this program, so the pass is active here — this branch simply does not change what it does to it.

The branch's other two files cannot reach codegen either: `register_shared_box` writes a set read only by `is_mut_box`, whose only caller is `sroa_param`.

On the measurement itself: the three flagged rows are `-O1`, `-O2` and `-O3` of one benchmark, sampled in a single `Benchmark` job on a single runner, so one slow runner moves all three together. `json/twitter/ser` and the `canada` / `catalog` suites are unaffected, which fits runner noise rather than a codegen change — a lost serde optimization would not spare serialization.

Happy to re-run the benchmark job for a second sample if you'd like one.

---

PR #1798, opened 2026-08-13
fix(elaborate): type a closure by its `return`, wherever the `return` nests

gfx:
## `__call` typed by its `return`

A block-bodied closure passed directly as a call argument, inside `for … of` over a `List`, hits a site that elaborates twice — and the AST walk that infers its return type reads `expression_types`, which is empty for whichever pass the site discarded. That site has no expected fn type to fall back on either.

`reify_closure` now falls back to `tir_block_return_type`, reading the returns off the body TIR the reifier has just built. That exists whichever pass recorded what, so the closure is typed by its `return` regardless.

The walk is the TIR counterpart of `control_flow::find_return_type_in_block` and covers the same constructs — `LabeledBlock`, `Loop`, `VariadicForOf`, `If`, `Let`, `Match`, `WithHandler`, `Resume`. That coverage is the correctness condition, not a nicety: a construct it cannot see is a `return` the closure's type ignores, and the closure then pushes nothing where its caller pops a value, so the module fails core-Wasm validation. `for … of` and `while` both reify as a labeled block around a loop, and `loop { return x; }` carries block type `()` rather than `!`, so the fallback runs on the AST walk coming up empty rather than on the body's own type.

`closure_block_return_in_for_of.wado` pins one case per construct.

## `benchmark/http_routing`

`app.wado` keeps its response body as the `ByteSlice` that `json::to_bytes` returns. `write_raw` lowers a slice straight into linear memory, so the bytes are never copied between the two. Measured as its own server row against the unmodified app, interleaved slice by slice over 5 rounds: +1.8% to +7.7%, faster on all five request shapes.

`benchmark/README.md` carries a fresh HTTP/1.1 table (`bench.sh`, `SLICE=4 ROUNDS=5 CONNECTIONS=50`), and `http_routing/README.md` records why h2c is not a row: `oha` opens one stream per connection, which charges HTTP/2 its per-request framing and flow control while withholding the multiplexing that pays for them, so the resulting spread ranks HTTP/2 stacks rather than routing. Restoring those rows needs a load generator driving many streams per connection.

`docs/research-http-routing-performance.md` records where the remaining gap sits. An ablation with Hono on Bun as a fourth row puts the `floor` variant — no router, no JSON, identical `wasi:http` calls — 8-19% *above* Bun while the full app sits 20-28% below it, so the gap is guest handler work rather than `wado serve`'s HTTP path. It also carries two results that keep the next round from re-deriving them: `Config::gc_heap_initial_size` measured at 16 MB and 64 MB and rejected, and the dead `apply_case` per serialized field, whose call-site fix costs `wire_name` its inlining and so loses the deserialize side its compile-time name folding.

## `profiling` profile

`lto` and `codegen-units` are pinned rather than inherited from `release`, whose `codegen-units = 1` puts the final link past ten minutes of CPU. Thin LTO across 16 units keeps cross-crate inlining while the units codegen in parallel.

https://claude.ai/code/session_015yiTASxa9uuw6UcCRFAUV6

---
_Generated by [Claude Code](https://claude.ai/code/session_015yiTASxa9uuw6UcCRFAUV6)_


gfx:
The performance alert above was runner variance, not this branch. Re-running the job on the same commit came back clean.

Independent of the re-run, this branch cannot move those numbers: compiling each flagged benchmark with a compiler built from `main` and one built from `54d9635` produces byte-identical wasm.

| Benchmark | Levels checked | `cmp` |
| --- | --- | --- |
| `mandelbrot` | -O1, -O2, -O3 | identical |
| `fts` | -O1, -O2, -O3 | identical |
| `json_twitter` | -O1, -O2, -O3 | identical |
| `sqlite_parse` | -O1, -O2 | identical |
| `json_canada`, `json_catalog`, `syntax_highlight` | -O2 | identical |

That follows from what the change does. The fallback only alters a closure whose return type would otherwise come from a block type that ignores its `return` — and such a closure pushes nothing where its caller pops a value, so the module fails core-Wasm validation. Any program that compiles and runs on `main` is untouched by construction, and `mandelbrot` has no closure in its hot path at all.

The alert's own shape also fits a slow runner rather than a code change: it spans every benchmark family at 1.5-2.0x while skipping `mandelbrot (-O2)` and `json/twitter/de`, which is what a uniform slowdown looks like when some ratios land just under the 1.50 threshold.

---

PR #1804, opened 2026-08-13
refactor(gale): remove duplicated, dead, and speculative code in package-gale/src

gfx:
Sweeps `package-gale/src` for duplication, dead code, and shapes the language already has a helper for. Source is 176 insertions against 602 deletions; generated output is unchanged apart from the two deletions noted at the bottom.

## Invariants stated as assertions

`build_overlap_dispatch` takes two lists in different index spaces: `alts` is indexed by absolute alt index, `first_sets` is parallel to `indices`. All six call sites satisfy this, but nothing said so. An `assert first_sets.len() == indices.len()` pins it, so a seventh call site that builds `first_sets` over every alt fails loudly instead of emitting a `DispatchBranch` whose tokens belong to another alternative.

## One definition per pipeline

`finish_grammar` holds the post-parse pipeline — implicit-token synthesis, `check_references`, `check_left_recursion` — and both `parse_checked` and `assemble_grammar` run it. A check added there reaches every entry point.

## Duplication removed

- The four selection sorts over alt indices are `List::sort_by` calls. Each comparator is a strict total order (the index tie-break admits no ties), so the emitted order is unchanged. `sort_group_by_mandatory_count_desc` computes its `ll_match_length` key once per alt rather than once per comparison.
- `build_repeat` takes the inner-element builder as a `fn(&mut AtnBuilder, &Element)`, so the LR-suffix walk and the plain walk share one copy of the Optional/Star/Plus state and edge wiring. State allocation order and greedy/non-greedy edge order are as before.
- `find_lexer_rule` lives in `ir.wado`, where `lexer_gen` and `latn` both use it.

## Hand-rolled scans folded onto the List helpers

41 loops re-implemented `Iterator::find` / `any` / `position`. Closure parameter types infer from the expected signature, so the folded forms carry no annotations.

Four lookups returned an owned `Option<T>` while every caller only read through it, so each call deep-copied a whole `ParserRule` / `LexerRule` / `AttrDecl` under value semantics — `dump.wado` copied a `ParserRule` per rule rendered to read `alternatives`. They return `Option<&T>` with the `stores[...]` annotation `GenContext::find_rule` already carries. `find_field`, `find_fragment`, `find_lexer_rule` and `find_inlined_fragment` keep names that carry meaning past their bodies; the single-call-site helpers fold into their callers.

## Dead and speculative code

- `collect_lexer_member_method_names` had no caller: the Java lexer path passes empty member lists by design (`codegen.wado`).
- `MultiAltRule` and `GroupOp` carry one `dispatch` field. Lower assigned the same value to a `scan_dispatch` twin at all seven construction sites, which under value semantics duplicated the dispatch tree per rule and per group.
- `is_storable_group` is `alts.len() > 1 || alts[0].elements.len() >= 2` — the `GroupClass::Transparent` complement. The four-predicate disjunction it replaces was a tautology for `len >= 2`, since `is_general_group` is the negation of the other two.
- `runtime/atn.wado` carries no `#[allow(dead_code)]`. `atn_adaptive_predict` is `pub` and seeds liveness as a production root, so everything reachable from it is live; the lint still fires on genuinely unreachable items in that module.

## Documentation

- `GenContext`'s doc no longer claims the struct holds no walker-mutable state — `current_rule_index`, `scan_group_counter` and `alt_bindings` are exactly that. It states what is true: caller follow is threaded as a parameter, never stored.
- `ends_with_scan`'s doc describes `ends_with`.
- `TODO.md` points at `package-gale-highlight-wado/grammar/Wado.g4` for the optional-scan-guard live case, and `token-id-design.md` names only symbols that exist.

## Generated output

Regenerating the goldens changes two things in the 59 parsers that embed the runtime: the `#[allow(dead_code)]` lines, and the one `runtime/tree.wado` loop that is now `iter().any(...)`. No other generated line differs, verified by a before/after `gale gen` over 36 grammars covering ATN, LR, tournament, lexer and the real-world corpus (SQLite, ANTLRv4, Rust, TypeScript, css3, HTML) — the 20 grammars that embed no ATN runtime are byte-identical.


gfx:
The benchmark alert is measurement noise on that run, not a regression from this branch. Not acting on it.

Eleven of the twenty flagged entries belong to benchmarks that do not link `package-gale` at all — `fts`, `zlib/decompress`, and the six `json/*` suites. Only `gale_gen`, `sqlite_parse` and `syntax_highlight` reach this diff (`benchmark/gale_gen/gale_gen.wado`, `benchmark/sqlite_parse/sqlite_parse.wado`, `benchmark/syntax_highlight/syntax_highlight.wado` are the only benchmark sources that reference `package-gale` or `lib:gale`), and the branch touches nothing outside `package-gale/`.

Compiling two of the non-gale benchmarks at `-O1` with `package-gale` at `origin/main` and at this branch's head gives byte-identical artifacts:

| benchmark | `origin/main` | `5a3e157` |
|-|-|-|
| `json_twitter` | `df897896c4c8c6c95add1110426e276834a1dc8ac4746c2787097e3df4cc94be` | `df897896c4c8c6c95add1110426e276834a1dc8ac4746c2787097e3df4cc94be` |
| `zlib_bench` | `e6b033b388adb573d68f9824cee5a5865392d2f6c667804bc0780ae645ec4053` | `e6b033b388adb573d68f9824cee5a5865392d2f6c667804bc0780ae645ec4053` |

`json/twitter/ser` is reported 1.79x slower while running the same bytes, so the run's numbers do not separate this branch from its base.

The flagged set is also exactly the allocation-heavy suites; `count_prime`, `mandelbrot` and `sieve` stayed under the threshold. That is the shape of GC-sensitive benchmarks on a loaded runner rather than a code change.

Worth a second look if the alert reproduces on a re-run, since `gale_gen` (1.61x at -O1) does exercise the changed code — though it moved with the suites that cannot have changed.

---

Issue #1819, opened 2026-08-15
const_object_globalization: the borrow-leak gate misses a projection root


gfx:
## Context

`const_object_globalization` hoists a constant aggregate into a module global. That is only sound while nothing hands the hoisted object's storage back out to a caller that mutates it — otherwise every caller shares one object. `SequenceLiteralBuilder::build(&self) -> List<T> { return *self; }` is exactly that shape: it returns the referent, and the caller keeps no copy because the literal it borrowed was fresh. `core:zlib`'s `build_huffman_tree` accumulated its `sym_list` across calls that way and wrote a literal-tree symbol index into the 30-element distance-tree lengths.

The gate added for that (`local_leaks_through_call` / `ref_args_that_escape`, both via `callee_ref_param_leaks` → `param_storage_escapes`) asks, per call argument, whether the callee delivers its parameter's referent out.

## The gap

The caller-side predicate `borrows_local` matches only `x`, `&x` and `&mut x`. The callee-side one it is paired with, `projection_roots_at`, already reads a field, index, cast or deref chain as naming the root's storage. So a constant borrowed through a projection slips the gate:

```wado
struct Holder { inner: List<i32> }

let h = Holder { inner: [] };
h.inner.build()      // borrows h's storage; `borrows_local` sees no `h`
```

`h` is still hoisted, the callee hands the referent back, and a later `push` accumulates across calls — the `build_huffman_tree` failure mode reached through a projection instead of a bare local.

Raised in review of #1810.

## Why it is not fixed there

Walking `FieldAccess` / `Index` / `Cast` in `borrows_local` closes it and costs four fixtures:

```
const_global_scalar_read
const_object_globalization_large_array
const_object_globalization_large_ref_array
const_object_globalization_long_literal_let
```

The loss is the widening multiplied by the callee side's over-approximation: `callee_ref_param_leaks` counts an unknown callee as leaking, so once projections are in scope, `names[i % 2]` handed to `Display::fmt` as a receiver is enough to refuse the `let names` hoist. `const_object_globalization_long_literal_let` is precisely that program. While only bare locals were in scope, that over-approximation never showed.

No reproduction exists for the projection shape today, so the branch keeps the bare-local test rather than trading four real optimizations for a theoretical hole.

## What to do

Close the gap **without** giving up the hoists. The gate is what needs sharpening, not the widening that exposed it:

[...]

PR #1822, opened 2026-08-15
perf(optimize): price the inline budget in emitted Wasm instructions

gfx:
## Inline cost model

The inline budget counts what the callee's hot path emits, in Wasm instructions. An operand leaf — a local read, a constant, a global read — is free, one operation is 1, a call is 2, a branch is 2 and a loop 3: control flow spliced into a caller splits its regions and costs it register pressure where a chain of field reads does not. A `cold_path()` marker or a diverging statement ends the walk over its block, so an error path prices nothing, and the promoted operand graph is walked alongside the skeleton with each hash-consed value charged once.

Arity is priced where it is the type's rather than the node's. Two operand leaves come free per node — what a binary, the widest fixed-arity node, already takes — and an aggregate literal or a match arm past that pays 1 apiece. A whole-struct constructor and an n-way dispatch of constants therefore cost what they emit instead of what an `if` costs.

`Buf::reserve` — the shape every `String` / `List` append goes through, a chain of field reads plus a cold-guarded capacity branch — is a handful of instructions and inlines at `-O2`. A loop around calls of the same node count stays a call. `inline_cost_model.wado` pins both.

## `slot_temp_sroa`

`sroa_variant_return` gives a helper a `[tag, slots…]` result that the call ABI passes flat. Inlined, NIR has no flat form, so `let v = f()?` allocated the tuple on the heap at every exit and read it straight back — `json-canada` paid one per coordinate. The new rule gives each projected slot a local declared ahead of the block, has the exits assign them, and the tuple never exists. It applies where `labeled_block_fusion` cannot relocate the consumer into the block, and refuses a temp declared more than once or a slot type with no zero value to declare it with.

## Aliasing in `const_object_globalization`

A constant a callee borrows is left alone when that callee delivers the referent back out. `SequenceLiteralBuilder::build(&self) -> List<T> { return *self; }` returns the storage, and the caller keeps no copy of what was a fresh literal, so hoisting would hand every caller one shared object to mutate. The gate covers a borrow in direct call-argument position, a borrow bound to a local first, a promoted argument, and an indirect or CM call, whose callee has no body to ask and so counts as leaking. `tests/integration/const_global_builder_alias.rs` compiles with the inline budget at zero, which is what keeps `build` a call, and asserts the literal stays out of a global.

The predicate matches the local and borrows of it, not projections over it: `h.inner.build()` still slips through. Widening it costs hoists the callee-side gate is too coarse to keep — #1819 tracks sharpening that side first.

## `return_every_exit`

Lowering rewrites a labeled block's value-carrying exits into multi-value returns, after which the block yields nothing and its result type must go. The check that every exit was rewritten now runs after the recursion, so an exit rewritten inside an `if` reports as rewritten. Reported as unrewritten, the block kept its result type and pushed values the void `if` around it never consumed, which codegen rejects with "values remaining on stack at end of block".

## `store_load_forward`

Nothing else queries a value in a CM export wrapper, so it reaches the pass with no `ValueGraph`, and `scoped_const_reads` only grows one that already exists. The pass builds it there, with the parameters seeded — `ensure_value_graph` never rebuilds, so what is seeded is what licm and `promote_fields` inherit.

## `core:json` and the grammar

The byte dispatches in `skip_value`, `skip_string`, `deserialize_any` and the exponent-sign scan are `match` over the scrutinee, which is the idiom the rest of the file already uses. `deserialize_any` gets `'-' | '0'..='9'` as an arm of its own.

`Wado.g4` gains `patternRange`, matching `Parser::parse_pattern_atom_or_range`: `..<` / `..=`, both endpoints required, binding tighter than `|`. The highlighter's grammar-coverage tests parse the stdlib, so a range pattern anywhere in it needs the rule.

https://claude.ai/code/session_01TxKca1U6EZ8csgD1jhQbDU

---
_Generated by [Claude Code](https://claude.ai/code/session_01TxKca1U6EZ8csgD1jhQbDU)_


gfx:
The `sqlite_highlight` size alert is the accepted side of this branch's trade, not an oversight.

Pricing the budget in emitted instructions brings shapes within reach that node count kept out, and the parser-heavy programs are where that shows on both axes. Measured against `main`, release build, same-batch best-of-three:

| Benchmark | main | branch | |
|-|-:|-:|-:|
| `syntax_highlight` | 4.21 MB/s | 4.88 MB/s | +15.9% |
| `json_catalog` ser | 404.18 MB/s | 464.41 MB/s | +14.9% |
| `json_twitter` ser | 241.44 MB/s | 271.46 MB/s | +12.4% |
| `fts` | 10.71 M/s | 11.83 M/s | +10.5% |
| `sqlite_parse` | 5.75 MB/s | 6.27 MB/s | +9.0% |
| `json_canada` ser | 91.54 MB/s | 98.85 MB/s | +8.0% |
| `sieve` | 165.70 M/s | 173.23 M/s | +4.5% |
| `zlib` compress / decompress | 42.69 / 212.12 MB/s | parity | |

Later commits on the branch were each measured neutral against that build, so the gains stand for the head.

Arity pricing already claws back most of the size. A dispatch arm and an aggregate field cost what they emit, which took `sqlite_highlight` from 466,621 bytes at a flat price to the 284,956 reported here — `-Os` `sqlite_parse` moves 478,582 → 295,889 on the same change. Charging an arm a full `BRANCH` instead of `OP` trims another 0.4% and costs `sqlite_parse` ~2% of its throughput, which is the wrong side of this trade: Wado's output is already well inside the field on size, and runtime is the priority.

---

PR #1836, opened 2026-08-17
perf(gale): emit each parser decision once, and dispatch to rules that can match

gfx:
Gale's emitters repeated themselves and reached for work the input could never need. This makes each decision appear once in the generated parser, and makes the lexer's dispatch offer a character only to rules that can start with it.

## Generated code shape

A lookahead branch is emitted once per distinct sub-tree. `prediction_node_eq` compares `Dispatch` children structurally, so first tokens leading to the same decision share one arm — TypeScript's `_parse_type_parameter__inner` is 48 lines covering 20 first tokens.

An alternative's scan body is emitted once. `plan_scan_multi_alt` counts the partitions that reach each alt and hoists any the dispatch would otherwise copy into a `<scan fn>_alt_<n>` helper; an alt reached from one partition stays inline, so a grammar without the duplication keeps byte-identical output. SQLite's `scan_expr_atom` is 76 lines.

Each rule's `no_viable` names its wanted set as an interned `&SYNC_<id>` of kind ids, and `fail` resolves those to names only for the error it keeps. `token_kind_name` therefore has five call sites in a generated parser, all in the runtime.

Three emit sites drop work the shape already decides: a rule entry binds no unread `start`, an index-pick cascade omits the test its unconditional tail duplicates, and a single-token repeat whose guard proves the body advances just advances.

## Lexer dispatch

The first-char walk recurses into any referenced lexer rule, `fragment` or not, so a rule built out of plain rules carries their first chars. It falls back to wildcard for a reference already on the walk and for a callee whose own matcher folds a case the walk does not — `callee_folds_other_case` resolves the callee's `caseInsensitive` rather than reading only what it declares, so a rule that overrides the grammar's setting still reaches the dispatch for the cases its callee folds.

A `mode` grammar dispatches on the first character the same way a modeless one does: both build their candidates through `build_mode_try_calls`, per mode. TypeScript's mode 0 is a 57-branch dispatch whose worst branch makes 10 `try_` calls.

## Codegen invariants

Every alternative reaching codegen scans in full — non-scannability's only source is a `RuleRef` to a name that is not a parser rule, which `ir.wado` rejects first — so the `Ambiguous` arm asserts that and lowers to the longest-match tournament, which is what `antlr4-compatibility.md` specifies. The LR suffix dispatch is scan-side only.

Two dispatch branches fold only when their `Ambiguous` leaves emit the same code: `ambiguous_eq` compares the alt set plus what `ambiguous_node_routes_via_atn` reads, so leaves that route differently keep their own arms.

The standing rule reads "no backtracking on the accept path" and names its one exception: the repeat-exit probe re-parses a failed element under `speculating` and rolls back all but `pending`, which is what points a diagnostic at the offending token.

## Corpus

`tests/generated/` is regenerated. SQLite 794 KB, Rust 1.22 MB, TypeScript 1.42 MB.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FKkyNYgVmGQyNk8CHqK177)_


gfx:
The benchmark alert above is runner-wide, not a regression from this branch — but it does not clear the three Gale rows either, and I can't measure those here. Details so nobody has to re-derive it.

## Why the alert is the runner

This branch changes no file outside `package-gale/`. Only `sqlite_parse`, `syntax_highlight` and `gale_gen` depend on that package; `mandelbrot`, `fts`, `zlib/decompress` and `json/*` do not, and they regressed by the same factor:

| group | rows | min / mean / max ratio |
| --- | --- | --- |
| depends on `package-gale` | 5 | 1.55 / 1.60 / 1.71 |
| independent of it | 12 | 1.50 / 1.58 / 1.68 |

`mandelbrot` is a numeric loop with no parser in it and reports 1.51. A uniform ~1.6× across every workload in the suite is the machine.

This is the failure mode `perf.md` already records as a standing rule — "Benchmark on an idle host, and only against an arm measured in the same window", noting a 27% swing for identical code on the same host hours apart. This run is a ~60% swing on workloads the diff cannot reach.

## What it does not settle

The two distributions overlap completely, so this run cannot separate a Gale-specific effect from the runner: `sqlite_parse` at 1.71 is the worst row, but `json/canada/ser` at 1.68 is second and has nothing to do with Gale.

There is one mechanism in this branch that could genuinely cost parse time. Hoisting a repeated alternative's scan body into a `<scan fn>_alt_<n>` helper turns an inlined body into a call, and `perf.md` warns that the tiniest scanners are the ones that benefit most from inlining. The counterweight is that the emitted code shrinks (SQLite's `scan_expr_atom` from a 989-line function to 76), so `-O2` has less to chew on. Which way that lands is an empirical question this alert doesn't answer.

I could not run the A/B myself: `benchmark/` needs `wasi-sdk`, and fetching it fails in my sandbox (GitHub access here is scoped to this repository). If someone has a benchmark host, `mise run benchmark-sqlite-parse` on this branch and on `main` alternating in one window is the check — that is the row to watch.

Everything else is green: `Test (Core)`, `Test (Wado)`, all three OS builds, and the `Benchmark` job itself all passed; the alert is an advisory comment, not a failing check.

---

PR #1837, opened 2026-08-17
refactor(lsp): walk the AST through AstVisitor, and move filesystem access into the filesystem host

gfx:
Cleans up the `wado-lsp` smells found in a review of the crate: a hand-rolled
AST traversal, filesystem code sitting in the wrong layer, and duplication
across span conversion, JSON-RPC transport, and test setup.

## Semantic tokens walk the AST through `AstVisitor`

`semantic_tokens.rs` classifies identifiers from a `SpanCollector` that
overrides `visit_type` / `visit_generic_params` / `visit_trait_bounds` /
`visit_function` / `visit_expr` and leaves the traversal to
`wado_compiler::ast`'s `walk_*`. A node a later AST change adds is therefore
reached by construction, which is what the rest of the crate's visitors
(`hover`, `definition`, `inlay_hints`) already relied on.

That covers declarations a hand-written walk has no reason to reach: resource
and test bodies, struct-field and parameter defaults, impl constants, trait
supertraits, and pattern type qualifiers. A resource method's signature types
classify as `type`, and a closure parameter inside a `test` block as
`parameter`.

`test`, `do`, and `resume` are contextual keywords — the lexer returns them as
identifiers — so they are classified from the AST instead: `TestDecl::span`,
`ResumeExpr::span`, and a new `WithHandlerExpr::do_span`, which the parser
captures at the `do` token.

## Filesystem access lives in the filesystem host

`wado-manifest` is pure: it says where a dependency *belongs*, given a manifest
and an already-parsed `wado.lock` — coordinates, versions, cache paths, and
`[workspace].members` glob matching. It opens no file and reads no environment,
so it holds for every target including `wasm32-unknown-unknown`.

Finding out what is actually on disk is `wado-lsp/src/host/discovery.rs`, beside
the `FilesystemCompilerHost` whose `dependency_index` is its only caller: the
governing `wado.toml` walk-up, `[workspace.package]` inheritance, `wado.lock`,
`package_lib_entry`, the `~/wado` cache root, and git worktree placement.
`wado-cli` reads dependencies through the same module, so the CLI and the
language server resolve one set of paths. `host.rs` keeps the `CompilerHost`
impl and the bridge from `discovery::resolve_all` to the compiler's
`DependencyIndex`.

## Kiln consume-only discovery goes through the host

`kiln::prepare_invocations` is async and reads every file through the
`CompilerHost` it is given — the manifest walk-up, the per-invocation
`.kiln.json` metadata, and the recorded outputs. Only best-effort path
normalisation (`safe_join`'s sandbox check and the symlink resolution behind the
`kiln:` URI) touches the filesystem directly, and it degrades to the lexical
form where there is none.

`CompilerHost::source_exists` answers presence without loading; the trait
default is a `load_source`, and `FilesystemCompilerHost` overrides it with a
stat, so the per-output presence check costs no bytes. An absent kiln cache and
one that will not read are reported separately.

## Deduplication

- `text::range_from_codepoints` is the single compiler-span-to-LSP-`Range`
  conversion; `diagnostics.rs` routes `DiagnosticSpan` through it, so a
  diagnostic and a hover over the same characters cannot disagree.
  `code_units_of` serves both directions of the encoding conversion.
- `transport::send` frames every JSON-RPC message; `send_response` /
  `send_notification` / `send_error` are the typed envelopes over it.
- `dispatch`'s `target_document!` declares `TargetDocument` for the four params
  types that name a document.
- `ast_search::{FirstMatch, find_in_module}` is the short-circuiting module walk
  hover's local renderer and definition's `#include` path finder share.
- `QueryContext::declaration_span` anchors a declaration for both
  find-references and document-highlight.
- `test_support::{open, open_at, open_files}` and `query::test_ctx::with_ctx`
  build the fixtures; `Engine` and `QueryContext` tests use them instead of
  their own setup.
- `discovery::resolve_all` reads `wado.lock` once per batch and returns
  `Ok(DependencyEntry)` or a reason phrased for the `use` site.

## Also

`FunctionId` carries `#[allow(clippy::large_enum_variant)]`: boxing `MethodName`
would cost an allocation per call-graph edge in `optimize/dce.rs`, where
`clippy -D warnings` otherwise fails.

https://claude.ai/code/session_01JocSkQXUQbqPqNQPXwZ5ju


gfx:
The benchmark alert is not attributable to this branch: it generates byte-identical wasm.

Compiled two of the flagged benchmarks at this branch's head (`e8d78ae`) and at its base (`cb77f99`) and compared:

| Benchmark | `-O` | sha256 |
| - | - | - |
| `json/twitter` | `-O2` | `41f3eec056840337d1cbb7de74c03a3a2656bd25561b578e3cc055346535e4cb` (both) |
| `fts` | `-O1` | `ba509b98c187f7bb2c7306fe17fdc349417af4d6d0f532580cf7e08780e6a602` (both) |

`cmp` reports no difference for either. Guest runtime cannot move when the guest module is the same bytes.

That matches what the branch does to the compiler — four additions totalling 17 lines, none of them on a path that reaches codegen:

- `ast.rs`: a `do_span: Span` field on `WithHandlerExpr`
- `parser.rs`: capturing that span at the `do` token
- `name.rs`: an `#[allow(clippy::large_enum_variant)]` attribute and a doc comment — no executable change
- `compiler_host.rs`: a `source_exists` trait method with a default, called only from the LSP's kiln discovery

The shape of the alert points the same way. The ratios sit in a 1.51–1.67 band across JSON serde, full-text search, SQLite parsing, syntax highlighting, and ANTLR generation — workloads with little in common. A codegen or optimizer regression concentrates in whatever it pessimises; a near-uniform factor across everything is the runner, not the compiler. The `Benchmark` job itself concluded `success`, so this is advisory and does not gate the merge.

Worth noting the baseline is `7529ba7`, which is not this PR's base — if the band persists on `main`, the regression is upstream of this branch and this comparison would not surface it.

---

PR #1847, opened 2026-08-18
perf(optimizer): keep alias-class work linear in the class size

gfx:
A function whose reference locals share one pointee struct lands them all in a single alias class — the shape every Gale-generated parser and the Gale codegen itself have. Work proportional to the square of that class size showed up in three places; each is now linear.

## Alias classes

`AliasGroups` holds each equivalence class once and indexes it by member, so a class of `m` locals costs `O(m)`. `class_of` hands back the canonical root with the members, and `build_mut_escaped`'s escape closure sweeps by that root — every member seeds its own class, so closing per seed would walk the member list `m` times.

The union-find `find` is iterative. A chain of `let dst = src` copies leaves a parent chain as long as the class, which recursion does not survive; two unit tests pin a 20 000-member star and a 20 000-link chain.

## Heap versions

`HeapState` carries one `escaped_version` for the whole `mut_escaped` set. Everything that invalidates the set as a unit — an impure call, a loop body's external writes — bumps that single version, and `version_of` maxes it in for escaped roots. Versions are only ever compared per key, so one shared version reads identically to a distinct `per_local` entry per local, and `per_local` stays sparse for `join_heap` to walk. It is an additional channel rather than a replacement: `per_local` still carries the bumps aimed at one local, such as a `&mut` borrow.

`escaped_version` snapshots, restores, seeds, and joins alongside `default_version`, taking a fresh version when any arm bumped it.

## Effect

Compiler output is unchanged where it was compared: the WAT for the five benchmarks that build is byte-identical at -O2, as is `json_twitter` at -O3.

`wado test -O3 package-gale`, measured against `origin/main` back to back on one machine:

| | main | this branch |
| --- | ---: | ---: |
| wall | 1782.16s | 1601.95s (-10.1%) |
| peak RSS | 6710 MiB | 5691 MiB (-15.2%) |

2630 tests pass either way. The CI "Test (Gale O3)" job is the run's critical path, so the wall-clock saving lands there.

### DIFF
--- a/wado-compiler/src/nir_value_graph/builder.rs
+++ b/wado-compiler/src/nir_value_graph/builder.rs
@@ -26,6 +26,12 @@ struct HeapState {
     field_global: IndexMap<u32, HeapVersion>,
     /// Version covering slots in none of the maps above.
     default_version: HeapVersion,
+    /// One version standing in for a `per_local` entry on every `mut_escaped`
+    /// local. An impure call invalidates that whole set at once, and versions
+    /// are only ever compared per key, so a shared version reads identically to
+    /// a distinct one per local — at `O(1)` per call instead of the
+    /// `O(escaped)` a bump each cost.
+    escaped_version: HeapVersion,
 }
 
 impl HeapState {
@@ -36,6 +42,7 @@ impl HeapState {
             per_local: IndexMap::default(),
             field_global: IndexMap::default(),
             default_version: HeapVersion::INITIAL,
+            escaped_version: HeapVersion::INITIAL,
         }
     }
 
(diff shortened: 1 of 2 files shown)


gfx:
On the `json/twitter/de (-O3)` alert: the compiled artifact is identical, so the branch cannot be the cause.

I compiled `benchmark/json_twitter/json_twitter.wado` at **-O3** from this branch and from `origin/main` (`67aea4457`, the exact commit the alert compares against), back to back on one machine:

```
new=4243970  old=4243970
IDENTICAL
```

Byte-for-byte, 4,243,970 bytes either way. Same guest wasm, same wasmtime — the 167.19 → 108.62 MB/s gap has no artifact difference behind it and reads as runner variance.

Correcting the PR body while I am here: its "byte-identical" claim was measured at -O2 only. It now covers -O3 for this benchmark too, and I have made the body say which levels were checked rather than implying all of them.

---

PR #1863, opened 2026-08-20
fix(elaborator): expand power-assert coverage

### LINKED ISSUE #1855: power-assert: a chained comparison renders no operand values
`assert` exists to show the values behind a failed condition, and it drops them for a chained comparison.

```wado
export fn run() {
    let index = 5;
    let used = 3;
    assert index < used, "index out of bounds";
}
```
```
Assertion failed in run at p.wado:4: index out of bounds
condition: index < used
index: 5
used: 3
```

Adding the lower bound loses both lines:

```wado
    assert 0 <= index < used, "index out of bounds";
```
```
Assertion failed in run at p.wado:4: index out of bounds
condition: 0 <= index < used
```

The condition is still checked correctly; only the diagnostic degrades, silently, so the writer of the assert has no signal that the form they chose reports less.

This is not hypothetical in the stdlib: `List::insert`, `List::remove`, and `List::swap` all assert `0 <= index <= self.used`, so their bounds failures have never shown the index. WEP-2026-06-02 Phase D wanted the same lower bound on the index traits for uniformity and reverted it, because `index_value` / `index_assign` / `index_ref` / `index_ref_mut` would have traded `index: 5` for nothing — and the whole reason those traits assert rather than let Wasm trap is the diagnostic.

Chained comparison is reportedly one of several forms the instrumentation does not cover; the general fix is to enumerate what it supports and either instrument the rest or reject at compile time what it cannot render, rather than degrade quietly.

gfx:
Widens what an `assert` failure reports, and makes the instrumentation that collects it evaluation-preserving. Closes #1855.

The contract is `docs/wep-2026-08-19-power-assert-coverage.md`: instrumentation preserves evaluation, rendering reports reach rather than just value, and no condition shape degrades silently.

## Instrumentation preserves evaluation

`assert e` evaluates `e` exactly as `let _ = e` would — the same sub-expressions, in the same order, and no others.

A capture with no short-circuit between it and the condition root is bound ahead of the condition. A capture below one is taken where the operand sits, so the short-circuit still decides whether it runs. The boundaries are the right operand of `&&` and `||`, and every operand of a comparison chain past the first comparison (`a < b < c` runs as `(a < b) && (b < c)`).

So `assert i < list.len() && list[i] == 1` with `i` out of range reports the assertion rather than trapping on the subscript, and an assert whose right operand has a side effect does not run it when the left operand short-circuits.

## Rendering reports reach

A conditional slot the run never reached renders `&lt;not evaluated&gt;`:

```
condition: i < list.len() && list[i] == 1
i: 99
list.len(): 3
i < list.len(): false
list[i]: &lt;not evaluated&gt;
list[i] == 1: &lt;not evaluated&gt;
```

Each conditional slot costs one `bool` flag, cleared ahead of the condition and set at the capture site; its text is chosen in the failure branch, so a passing assert pays nothing. The value local is never bound — Wasm locals are default-initialized — so nothing has to synthesize a zero for an arbitrary `T`. `lower` asserts that a `TirExprKind::Null` is never typed as a primitive, which is what a `ref.null` placeholder at a primitive slot would produce.

A slot whose operand is a plain binding gets no binding of its own either: the failure branch re-reads it, which straight-line code makes exact.

## Every operand position is captured

Comparison chains render their operands. The scanner descends into casts, subscript indices, tuple and struct literals, `if` and `match` operands, ranges, and `matches`. Its match over `Expr` is exhaustive, so a new variant is a compile error at the decision point rather than a fall into a leaf arm.

The `condition:` line is quoted through the formatter's `Unparser`, which prints the expression the user wrote; `unparse_expr_simple`'s readability-over-fidelity trade stays right for its own callers.

`wado dump --assert-plan` prints, for every `assert`, which operands it captures and whether a short-circuit can skip each one. `tests/integration/assert_capture_plan.rs` reads that back for one `assert` per condition shape, so what each shape covers is a test rather than prose that goes stale.

What stays out of scope, and why, is in the WEP. Its one open item covers projection receivers, subscript receivers, method receivers and `matches` scrutinees together: rendering an aggregate operand is a real use of it in the failure branch, and the escape and scalarization analyses are right to count it. `builtin::cold_path()` grants no exemption — it compiles to a branch-hint entry and nothing else — so what lifts the restriction is rematerializing at the cold use rather than ignoring it. Power assert makes that reconstruction cheap, since the failure branch wants a rendering and scalarization leaves exactly the fields `Inspect` would walk.

## A closure dispatch stub is named after its type

A `fn(..)` type names no declaration, so its rendering is its identity: `mangle_fn_type` spells the whole signature, and `TypeTable::fn_receiver_name` is the single place a receiver is derived from it. `name::is_fn_type_name` is how a consumer asks whether a base name is a closure receiver.

A stub therefore exists only for a function type every part of which is determined — parameters included. An undetermined part would put an inference variable in a function name, and no monomorphized closure can dispatch through such a stub.

## Docs

`docs/wep-2026-01-25-closure-implementation-internals.md` states the closure design as it stands: dispatch chosen by escape analysis, `fn` / `fn mut` as a type-system distinction with no backing traits, implicit captures, and the Wasm GC representation down to the two canonical-closure shapes. Its deferred work is an Open list.


gfx:
The `json/twitter/de (-O3)` benchmark alert above does not reproduce. Recording the measurement so nobody repeats it.

Four paired runs of the Wado half at `-O3`, same machine, interleaved main-then-branch to control for drift (`de`, MB/s):

| | main | this branch |
|-|-|-|
| 1 | 69.49 | 71.02 |
| 2 | 70.04 | 70.98 |
| 3 | 63.60 | 70.93 |
| 4 | 67.82 | 69.06 |
| mean | 67.7 | 70.5 |

The branch is faster in every paired run, and its spread is narrower. The alert fired at a ratio of exactly `1.50` against a `1.50` threshold, and the `Benchmark` job itself concluded success, so this reads as runner noise at the boundary rather than a real change.

Worth noting what the branch does add to that path, since it is real even though it costs nothing measurable here. `substr_bytes` guards with a chained `&&`, and `wado dump --assert-plan` shows six of its operands becoming conditional slots:

```
833: assert 0 <= start && start <= end && end <= self.used
  __v0  always       start
  __v1  always       0 <= start
  __v2  conditional  start
  __v3  conditional  end
  __v4  conditional  start <= end
  __v5  always       0 <= start && start <= end
  __v6  conditional  end
  __v7  conditional  self.used
```

That is required for correctness — an operand behind a short-circuit has to be captured where it sits, or the assert changes what the condition evaluates. Two refinements are available if this path ever does show up in a profile: a conditional slot whose operand is a plain binding needs no value local (the failure branch can re-read it, as an unconditional place slot already does), and the seen-flag is derivable from the guard value the plan already captures rather than stored at runtime. Neither is worth doing on an unmeasured hunch, so I have left them out.

---

PR #1885, opened 2026-08-23
docs(wep): slice `core:icu` by locale, collect it by reachability

gfx:
`core:icu`'s data separates along two axes, and this measures both, then shapes the design around what each is worth.

## The measurement

The locale axis, from `bdp-spike/datagen`'s new `coll-loc` and `fmt-loc` sets (icu4x 2.2, CLDR 48.2.0):

| markers | root (`und`) | `und,ja` | every locale |
|---|---:|---:|---:|
| collation | 568 KB | 660 KB | 1043 KB |
| formatting (date/time, number, list, plurals) | 21 KB | 47 KB | 4334 KB |

The rows invert. Collation is a root table with tailorings hung off it, so slicing by locale saves at most 475 KB. Formatting is almost nothing but locale data, at a 206× ratio: a program that formats one date and declares one locale pays 47 KB where an unsliced build pays 4.34 MB. No dead-code elimination substitutes, because a locale is reached through the same constructor whatever the program declares.

Fallback dedup follows the same split — 5.16 MB down to 4.34 MB on formatting, 23 bytes on collation.

## The capability axis is reachability's

wasm-ld collects over the symbol graph that `linking` and `reloc.*` carry, and `--gc-sections` is already its default. Keeping that graph in the asset lets the collection run again later against whatever exports one program reaches. Measured against the ICU asset kept relocatable (4.05 MB), every slice lands on its from-source rebuild — 89 KB for locale + casemap against ~92 KB, 1.17 MB for collation against ~1.12 MB, 2.31 MB for segmentation against ~2.35 MB — and the root set reaches granularity a rebuild cannot express: a grapheme pass alone is 23 KB.

The slices are runtime-checked, not inferred. `runtime-check` covers the all-exports re-link, and its `casemap-only` binary asserts Turkish, German and en-US casing against a 93 KB component built from a re-linked slice, with `wit-casemap/` narrowing the world so the slice can be componentized.

## What the WEP now specifies

One prebuilt ICU asset, shipped relocatable and collected per program. Data that does not vary by locale is baked into it and collected away when unreached; only locale-bearing data is external, keyed by (capability, locale). Collation reads the normalizer's NFD markers from the baked side.

Assembling a program's data is an archive lookup: the reached locale-bearing capabilities name their marker sets, the declared locales expand through likely-subtags and the fallback chain, and ICU4X's fork provider hands the asset the entries at those keys. No ICU code runs at build time, and the marker set is fixed per capability, so nothing maps Wado symbols to ICU markers.

The type-level split in the facade carries its weight for a specific reason: the collection's roots are the asset's exports that surviving Wado code calls, and `liveness` seeds every method of a live type, so one `Segmenter` carrying four operations would root all four.

`wado-wasm-embed` keeps every active data segment live unconditionally, so implementing the collection there is recorded as a known gap with its mechanism settled.

Docs and spike only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MnWpa5xAokEXRNZSExZaF2)_

## Summary by CodeRabbit

* **New Features**
  * Added locale-aware Unicode case mapping for uppercase, lowercase, titlecase, and case folding.
  * Added locale parsing and string conversion support.
  * Added locale-specific collation and formatting data generation with deduplication options.
* **Documentation**
  * Expanded guidance on ICU data sizing, post-build slicing, archive lookup, and runtime loading.
* **Validation**
  * Added runtime checks for Turkish casing and Unicode sharp-s behavior.

### DIFF
--- a/docs/wep-2026-08-09-core-icu.md
+++ b/docs/wep-2026-08-09-core-icu.md
@@ -314,12 +314,22 @@ runtime data dependencies, not taxonomy.
 
   Closing it means giving the prune a symbol graph where it has a wasm index
   graph. The `linking` and `reloc.*` sections carry exactly that — which
-  function references which data symbol — so an asset kept relocatable could be
-  collected from the live exports the way a linker's `--gc-sections` does. Worth
-  ~2.35 MB where one component serves both a grapheme pass and word
-  segmentation and a program reaches only the first, and it would let a
-  component bake its locale-independent data again. Unverified: whether those
-  sections survive the ICU asset's build with the edges intact.
+  function references which data symbol — and wasm-ld already collects over it,
+  `--gc-sections` being its default. Measured on the ICU asset kept relocatable,
+  collecting against one program's exports reproduces the from-source rebuild
+  across every capability — 89 KB for locale + casemap against a ~92 KB rebuild
+  — and the slices are runtime-checked, Turkish casing included, so the graph
+  carries the data edges and not only the code.
+
+  Two things follow. Slicing needs no rebuild, so one asset can serve every
+  program; and the root set can be finer than a rebuild can express, grapheme
+  boundaries alone being 23 KB where segmentation's four operations are 2.31 MB.
+  What remains is implementing the collection in `wado-wasm-embed`, which today
+  reads the wasm index space rather than the symbol table.
+
+  This closes the capability axis only. A locale is reached through the same
+  symbol whatever the program declares, so no collection separates `ja` from the
+  rest — that stays the data image's job.
 
 ## Open questions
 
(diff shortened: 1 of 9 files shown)


gfx:
The Merge Risk note scoped `up to e7e46` says the spike build still fails with an unknown marker-set error. That was fixed in `e7e46` itself, so the note reads one commit behind.

Verified at HEAD, from a clean state:

```
$ git show HEAD:wado-bundled-icu/bdp-spike/build.sh | sed -n 15p
( cd datagen && cargo run --release -- casemap ../casemap.blob )

$ rm -f casemap.blob && (cd datagen && cargo run --release -- casemap ../casemap.blob)
wrote ../casemap.blob (23878 bytes) — set=casemap, 2 markers   # exit 0
```

23,878 bytes matches the 23.4 KB the README records for that blob. `build.sh`'s other `datagen` invocations (lines 46-48 and 77) already named their set, so line 15 was the only one affected.

---

PR #1893, opened 2026-08-24
feat(highlight): hold the Gale highlighter to the compiler's classification

gfx:
`wado-compiler` owns Wado's syntax; `package-gale-highlight-wado` restates it
by hand in `Wado.g4` + `Wado.highlights.scm`. `check-grammar` already ties the
two on *what parses*. This ties them on *what colours*.

## The checks

- **`mise run check-highlight-vocab`** — vocabulary only, over the registries
  rather than a corpus, so it runs in milliseconds. Four invariants: every
  keyword the compiler defines is a `Wado.g4` literal and carries the capture
  its `KeywordCategory` implies; nothing else is captured as a keyword; and the
  query captures `@operator` on exactly the `is_highlight_operator` set. It
  lives in `wado-dev-tools` and runs as a unit test, so `cargo test` covers it.
- **`mise run check-highlight`** — the stdlib + fixture corpus, compared span
  by span. `tools/highlight_dump.wado` emits the Gale side's capture spans, and
  `wado-dev-tools highlight-corpus` joins them against
  `wado_lsp::semantic_tokens::classify_all`. About 17s warm; both are wired
  into the `Test (Wado)` CI job beside `check-grammar`, reusing its build.

Both vocabularies project onto one `Class`. Comment / string / number /
keyword / constant / operator are decidable without name resolution, so
disagreement there fails the check. `Ident` is not — the compiler resolves a
dozen kinds a context-free grammar cannot — so it is reported as a capability
gap and never gated, broken down by the kind the compiler resolved it to.
Divergences collapse onto `(relation, compiler class, gale class)` patterns
with one example each; there is no committed exception list, because one
rotted here before.

The shift operator carries the single exemption: `Wado.g4` spells it as two
separate greater-than tokens so a nested generic can close, and the check
asserts that spelling still exists, so the exemption cannot outlive its reason.

## Compiler-side classification

`semantic_tokens` splits in two. `classify_all` is the verdict — every token,
keyed by byte span, multi-line ones included. `compute` narrows it to what the
LSP wire accepts. Dropping a block comment is a wire constraint, not a
classification, and the comparison needs the verdict.

Keywords classify by their registry category rather than by being keyword
tokens: `true` / `false` / `null` / `self` are constants (`variable` +
`readonly` + `defaultLibrary`, which is LSP's spelling for one, mapped to
`constant.language.wado` by a new `semanticTokenScopes` contribution), and
`matches` is an operator. `task`, `trap`, and `forward` join `test`, `do`, and
`resume` as contextual keywords the AST places — `RestClauseDecl` carries the
`trap` / `forward` word's span for that. `self` is the reserved one, so it is
recognised lexically. `b'0'` classifies as a string alongside `'0'`, and a
`__DATA__` tail mutes as a comment.

Template literals expand. The lexer records each interpolation's origin and
the parser lexes and rebases it to build the AST; that call becomes
`lexer::lex_interpolation`, and the classifier uses it, so the code inside an
interpolation classifies like any other code — resolved symbols included, since
the symbol map keys on byte start. Everything outside an interpolation is
string, and a `:spec` tail mutes as a comment: the `TextMate` grammar is
regex-only and cannot parse a specifier, so it colours the fill character as an
operator.

## Grammar and query

The query gains the operator set, `@type` on type references and type
parameters, and `@property` on member names — the identifiers a context-free
grammar can place. Overrides match anywhere under their rule and the first
declared wins, so the `IDENTIFIER` rules are ordered most-specific first, and
`interpolation` gets `formatSpec` and `interpolationEnd` sub-rules so the query
can name the `:` and the closing brace without also catching the `:` of a
closure parameter written inside an interpolation. `resource X extends Y`
reaches the grammar.

What stays uncoloured is `variable`, `function`, and a few `type` and
`enumMember` — the kinds that need name resolution.


gfx:
Both findings addressed in cea48e094 — one of them not the way it was filed.

**Format-specifier offsets** — the bug does not reproduce; the offsets are already raw-spec relative. `format_spec.rs`'s `Parser::new` adds the leading-trim width back:

```rust
let lead = spec.len() - spec.trim_start().len();
let mut offsets: Vec<usize> = trimmed.char_indices().map(|(byte, _)| lead + byte).collect();
```

and `error_offsets_point_at_the_offending_character` pins `err(" 5zz").offset == 2`. The suggested repro `${x: zz}` reports `3:21`, which is the first `z`, not the space. Checked across leading whitespace in the expression, around the `:`, and inside the specifier — six cases, all landing on the offending character.

What was real is the doc: `FormatSpecError::offset` said "after trimming" and `parse` said "offset in the trimmed spec", both contradicting the code and the test. That wording is what makes the callers look off-by-`lead`. Corrected.

**`RestClauseDecl::keyword_span`** — fair. `AstSpans::contextual` keys on the span's byte start, so a shift silently stops the highlighter colouring the word while every `kind` assertion still passes. (`wado-lsp`'s `contextual_keywords_task_trap_and_forward_are_keywords` does catch it, but only from another crate.) `rest_clause_keyword_span_covers_the_word` now reads the source back through the span for both clause kinds.


---

PR #1897, opened 2026-08-25
feat(wasm-embed): prune an embedded asset's data segments by the byte

gfx:
Closes the capability half of the ICU WEP's "the prune keeps baked data" gap, piloted on `wado-bundled-libm`.

An active data segment initialises memory whether or not anything still reads what it wrote, so the wasm index space alone is never enough to drop one. `wado-wasm-embed` prunes an active segment by the byte instead, from a map of which data ranges each function reads, and emits every surviving run as a segment of its own.

## Where the map comes from

`linking` and `reloc.CODE` carry the graph — it is what wasm-ld collects over for `--gc-sections` — but only against the byte offsets of the binary it produced, and a `.wat` round trip re-encodes every relocatable immediate to its narrow form, moving every offset a relocation holds. libm ships as `.wat`, so `mise run update-bundled` resolves the graph once at asset-build time into a form that names no offset into code, and the asset carries it in a `wado.dataref` custom section:

```wat
  ;; 16 functions claim 5444 of 5448 data bytes; the rest is padding.
  (@custom "wado.dataref" (after data)
    "_RNvNtNtCs5s8uptHhuCz_4libm4math14rem_pio2_large14rem_pio2_large 0:0+344\n"
    "_RNvNtNtCs5s8uptHhuCz_4libm4math4exp24exp2                       0:496+4096\n"
    "libm_cbrt                                                        0:472+24\n"
    …
  )
```

One entry per line, sorted by offset so the map reads as a partition of the segment, with the coverage figure in the header: a map that stops describing the asset shows up as a claim well short of the segment. `dataref::resolve` is the resolver; an asset shipped relocatable keeps the sections it reads and can run it at compile time instead, which is what `core:icu` will do.

## Effect

| program | wasm | libm rodata |
| --- | ---: | ---: |
| no math | 6,046 B | — |
| `sin` | 10,946 B | 344 of 5,448 B |
| `sin`/`cos`/`tan`/`exp`/`pow` | 16,005 B | 664 of 5,448 B |

Two thirds of libm's rodata is `exp2`'s table alone. All 54 of the asset's exports answer bit-identically to the unpruned module.

## Safety

The map can only make the prune keep too much or too little of the asset's own data, and too little is what produces wrong answers, so every way it can fail to describe the module is an error rather than a narrower prune:

- a name that no function in the asset's `name` section carries;
- a range reaching past the end of its segment, or with no end to reach;
- a map naming no function at all, which reads exactly like one that failed to resolve;
- `reloc.DATA`, a pointer stored in the data itself — an edge the function-to-data map cannot express.

A range naming a segment no map can narrow is dropped rather than honoured halfway: a passive segment reaches memory through the `memory.init` that copies it, which the walk already follows, and a segment at a computed address cannot hand its pieces addresses of their own. An asset carrying no map keeps every active segment whole, which is what a hand-written `.wat` asset gets.

## Also here

A generic builtin resolves through the `core:builtin` module with its type-parameter slots, so a literal argument is re-coerced to the width the substitution chose and `let y: f32 = builtin::black_box(0.5)` compiles. The call site substitutes into the parameter types the single `lookup_function_signature` already returned, so every callee spelling — not only a bare free-function name — reaches the argument-count and type checks that path performs.

`bundled-asset` drops the relocation and DWARF sections it has spent, keeping `libm.wat` free of escaped binary blobs.


gfx:
Worked the three findings.

**`dataref.rs` — zero-size range accepted.** Real, and fixed in 359152cc0 (which landed before this review posted, so the review's `ce228` base predates it). Reproduced first: `a 0:4+0` gave `[(Some(12), [])]` — an empty active segment, plus a data section where the segment would otherwise be dropped whole. `DataRange::new` now refuses `size == 0`, which is the same diff the review proposed. `resolve` keeps its separate skip, since wasm-ld emits zero-size symbols legitimately and one is not an error there.

**`reach.rs` — assert the `splittable` bound.** Fixed in a47277704. The bound is established in `func_ranges`, which refuses a map naming a segment the asset does not have, so the assertion names where it comes from rather than leaving the index panic to speak for it.

**`docs/wep-2026-08-09-core-icu.md` — bold as a sub-section title.** Not changing this one. The rule in `docs/AGENTS.md` is about using bold *where a markdown section belongs*; a bold lead-in on a list item is the established convention across the WEPs — 295 of them, in dozens of files this branch does not touch, including other "Known gaps" bullets. The entry in question already read `- **The prune keeps baked data.**` before this branch; converting only this one to plain text would leave it inconsistent with every sibling. Happy to be overruled if the intent is that bullets should lose it too, but that is a repo-wide change rather than this PR's.


---

gfx:
@coderabbitai review

The previous run aborted because the head moved under it while I was pushing fixes for its own findings. Head is now `1d71367b9`, which carries both remaining items: the `splittable` bound stated as an assertion, and the zero-length-range test comment reduced to its reason.


---

PR #1902, opened 2026-08-26
feat(cm): the Component Model `map<K, V>` type, and wasmtime 48.0.1

gfx:
Syncs the `vendor/` spec submodules, adopts the Component Model `map<K, V>` type, and targets wasmtime 48.0.1.

## `map<K, V>`

The WASI Subgroup adopted `map<K, V>` (🗺️) on 2026-08-06, and WASI 0.3.1 (2026-08-11) names it a required feature of the baseline — a component claiming 0.3.1 must lift and lower it. Design: [`docs/wep-2026-08-25-cm-map-type.md`](docs/wep-2026-08-25-cm-map-type.md).

- `TreeMap<K, V>` is the Wado spelling of `map<k, v>`, both directions: `wado wit`, `wit_consume`, and `wado-from-idl`.
- At a component boundary `K` is one of the CM `keytype`s (`bool`, `s8`–`s64`, `u8`–`u64`, `char`, `string`). Anything else is a diagnostic naming the key type and the admissible set, rather than a silent degrade to `list<tuple<K, V>>` that would make the emitted WIT disagree with the source. Away from the boundary any `K: Ord` stays legal.
- `map` despecializes to `list<tuple<K, V>>`, so it shares that type's canonical ABI: flat `[i32 ptr, i32 count]`, 8 bytes / align 4 in memory. Only the component type section differs, where `CmDefined::Map` encodes `0x63`.
- Lift and lower go through `TreeMap`'s public API — `new`, `index_assign`, `len`, `entries` — with no intermediate `List<[K, V]>`. Wado's assignment is a copy, so a pair list would be a second full copy of the map's contents on every crossing. `entries()` yields insertion order and skips tombstones, so a lowered `map` carries no duplicate key; a lifted one resolves duplicates last-wins, the rule the spec states.
- Every method the binding calls is a registered `CompilerItem`. `liveness` prunes on the call graph, and a call minted after that pass has no edge for the graph to follow.
- The CM buffer walk (`CmBuffer`) and `store_ptr_len` are shared by the list and map paths.
- `package-cm-catalog` carries six `id_map_*` rows alongside `id_assoc_array(List<[String, u32]>)` — the despecialized shape — so the pair pins the specialization: same bytes, different type constructor.

A `map` parameter on an *imported* function is not lowered. The WEP records it as a Known gap with what closing it takes; export signatures are unaffected.

## wasmtime 48.0.1

- `wasmtime-wasi-http`'s host traits are the same across p2/p3, so `serve` and the request hooks bind `WasiHttpHooks` / `RequestOptions` / `WasiBody` from the crate root.
- Preopens are granted with `FsPerms::ReadWrite`.
- Socket creation is deny-by-default and `inherit_network` only relaxes the per-address check, so `wado run` / `test` / `serve` grant `allow_tcp` and `allow_udp` for the address check to be reached at all.
- A synchronous cancel traps on a waitable still in a set (`Trap::WaitableSyncAndAsync`). `AsyncCall::cancel` leaves the set first, and the four `Stream` / `Future` `cancel_read` / `cancel_write` methods, which expose the raw call, say so in their docs.
- `tests/fixtures/http_client_send_error.wado` pins the guest's view of a failed outgoing request: the harness has no mock for that path, its `send_request` hook fails, and the 502 the fixture answers can only come from the `Err(ErrorCode)` arm.

## The collector default

`DEFAULT_COLLECTOR` is `Collector::DeferredReferenceCounting`; `--collector copying` selects the copying collector.

The copying collector is the one we want at run time — several times faster on allocation-heavy / small-live-set workloads at the same resident set size, and it collects cycles — but wasmtime 48 makes it pathological to _compile_ against. `Component::new` on `benchmark/gale_gen/gale_gen.wado` (a 1.8 MB generated parser), same bytes, same cranelift opt level, same wasmtime feature set:

| engine                    | `Component::new`     |
| ------------------------- | -------------------- |
| wasmtime 47.0.3 + copying | 5.2 s                |
| wasmtime 48.0.1 + copying | unfinished at 14 min |
| wasmtime 48.0.1 + DRC     | 8.0 s                |

Every stack sample lands in `cranelift_codegen::alias_analysis::LastStores::update` ← `AliasAnalysis::new` ← `Context::optimize`. wasmtime 48 landed dead-store elimination in that pass (bytecodealliance/wasmtime#13806, #13947, closing #4167), restructuring it around a whole-function `observed_stores` pre-pass, and it goes superlinear on the store-dense code the copying collector's write barriers produce. DRC's barriers are sparse enough to stay out of it.

### For an upstream report

- #13806 measured the pass's compile-time cost at 1.0x–1.2x on 9/24 Sightglass benchmarks and #13947 claims full recovery, so 150x+ is outside anything upstream has seen.
- Why copying and not DRC: the same release window tags GC memory with alias regions — #13718 (GC-heap object writes), #13734 (collector heap-data structs), #13615 (`VMContext` field accesses), #14115 (precise regions for statically-known entity imports). The copying collector's barriers emit far more region-tagged stores than DRC's.
- Upstream ships copying by default since 46.0.1 (#13756), so a large enough module should reproduce there.
- No matching issue is open: alias-analysis, dead-store, compile-time, hang and superlinear searches filed after the 48.0.0 release (2026-08-20) turn up nothing.
- No fix is pending. `cranelift/codegen/src/alias_analysis.rs` on wasmtime `main` (`0ac37df998`) is byte-identical to the 0.135.1 this branch ships; the copying barrier codegen differs only by an `alias_regions` API refactor.

https://claude.ai/code/session_01SjiDajoFs6ZVT61nVwF8Ab

---
_Generated by [Claude Code](https://claude.ai/code/session_01SjiDajoFs6ZVT61nVwF8Ab)_


gfx:
Draft until wasmtime fixes the `alias_analysis` compile-time blowup this branch runs into. The `map<K, V>` work is independent of the runtime version and is ready; the wasmtime 48.0.1 upgrade is what is blocked.

The two ways to hold wasmtime 48 are both unshippable, and the choice is only which CI job goes red:

| default collector | `Test (Wado)` | `Benchmark` |
| --- | --- | --- |
| copying | `Component::new` on `benchmark/gale_gen/gale_gen.wado` does not return within 14 min | green |
| DRC (what the branch carries) | green, 1716/1716 files | 30/36 suites regress, 1.74x–28.9x |

The benchmark alert on this PR is the DRC fallback, not the alias-analysis pass — that one costs compile time only. Same binary, same wasm, `--collector` the only variable, `json_catalog -O2`:

| collector | ser | de |
| --- | --- | --- |
| DRC | 68.71 MB/s | 19.64 MB/s |
| copying | 547.02 MB/s | 122.90 MB/s |

Upstream detail is in the PR description under "For an upstream report". wasmtime 47.0.3 already carries `Config::wasm_component_model_map`, so landing `map<K, V>` on 47 is available if we would rather not wait.

---

PR #1903, opened 2026-08-26
feat(temporal): the TC39 Temporal model, and the prelude's ASCII family

gfx:
`core:temporal` covers the TC39 Temporal model: eight types, arithmetic, rounding, and text forms. Along the way the prelude gains the ASCII classification family on `char` and `u8`, and the elaborator's impl lookup stops crossing module boundaries.

## `core:temporal`

Eight types, all ISO 8601 — `Instant` and `ZonedDateTime` carry an instant, `PlainDate` / `PlainTime` / `PlainDateTime` / `PlainYearMonth` / `PlainMonthDay` are zoneless readings, and `Duration` is a signed span. `Unit` and `RoundingMode` parameterize the difference and rounding operations.

- **Arithmetic.** `add` / `subtract` / `until` / `since` on every type that can resolve the units it is handed; anything outside its row traps naming the type that can. `until` anchors at the receiver in both directions, so `a.add(a.until(b))` is `b` even when `b` is earlier. `since` is `until` negated, as in Temporal, and so does not round-trip.
- **Rounding.** All nine `RoundingMode`s. An `Instant` counts multiples from the epoch; a `ZonedDateTime` and a `PlainTime` count from local midnight, so a day-aligned unit lands on the civil boundary.
- **`Duration`.** A struct literal with field defaults is the constructor, so a mixed-sign value is representable; `add` / `subtract` rebalance through exact nanoseconds rather than summing component-wise, and the ISO 8601 form asserts on a sign mix it has no spelling for.
- **Text.** Every type parses and renders its ISO 8601 spelling, which is also the serde wire form — tag 0 under CBOR, a bare string in JSON, with an epoch-seconds number accepted on input. `Instant` additionally reads and writes RFC 7231 IMF-fixdate, plus the RFC 850 and asctime forms a recipient must accept.
- **`now()`** declares `with SystemClock`, so a program that never asks the time acquires no `wasi:clocks` import.

`time_zone` holds a string, but only `"Z"`, `"UTC"`, and `±HH:MM` are interpretable; an IANA name traps on first use. The WEP records why the zone data belongs to `core:icu` rather than a tzdb of this module's own, with the `zic` measurements behind the epoch-cutoff choice and the finding that ICU4X supplies zone identifiers but not offsets.

## Prelude: the ASCII family

`char` gains `is_ascii_digit`, `is_ascii_alphabetic`, and `is_ascii_alphanumeric`; `u8` carries the whole family, delegating to `char` so each classification keeps one definition. Every byte- and char-level caller in `core:prelude`, `core:temporal`, `package-gale`, and `package-marl` goes through them, including the case-insensitive literal comparisons in `fpfmt`'s inf/nan parser and `temporal`'s zone designator.

## Elaborator: impl lookup respects the module

`ImplModuleIndex::get` falls through from the mangled namespace to the declaration namespace, which is keyed by the bare written name. For a receiver whose mangled key carries its module, a miss there is a real absence — falling through reaches another module's same-named type. `Receiver::is_module_qualified` names that set, and the fallback is gated on it, so a package type no longer collides with a WASI binding of the same name.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hah9qeJztZensyW17xgS2S)_


gfx:
Four of the five findings were real and are fixed in 7df4afb51.

- **IANA zone serialization** — confirmed by running it: serializing a `ZonedDateTime` in `"Asia/Tokyo"` traps, because the wire form carries the resolved offset. Corrected in the module doc and in the WEP's consequences.
- **Calendar-bearing receivers** — confirmed: `PlainDate::new(2024, 1, 31).add(&Duration { years: 1, months: 1, weeks: 1, days: 1 })` succeeds. The list now names `ZonedDateTime`, `PlainDateTime`, `PlainDate`, and `PlainYearMonth`, each within the units it admits, with `Instant` and `PlainTime` taking hours and below.
- **U+0085** — it is `Cc`, not `Zs`/`Zl`/`Zp`, so the category list alone did not describe the predicate. Named explicitly.
- **`Instant::now` monotonicity** — agreed, `SystemClock` is a wall clock a host correction can move backwards. The two-read ordering is gone; the range assertion, which is what the test is for, stays.

Not taking the fifth, on `cross_module_same_name_inspect_newtype.wado`: a fixture with **no** `__DATA__` section already defaults to the test world and executes its `test` blocks — `wado-compiler/tests/e2e.rs:600-610` spells that out, and neighbouring fixtures such as `alias_shadows_same_name.wado` rely on it. Adding `{"test": {}}` would be a no-op.

---

gfx:
**Benchmark** failed on 96c34a0f8. Recording what I can establish, since a re-run is already in flight on 7df4afb51.

The regression is reported as uniform: all 36 measurements, every suite at every opt level, land between 1.52× and 2.07×. That includes `zlib/compress` (1.70), `zlib/decompress` (1.74), and `sqlite_parse` (1.95).

`core:zlib` and `benchmark/zlib/zlib_bench.wado` contain zero references to anything this branch changes — no `is_ascii_*`, no `to_ascii_*`, no `eq_ignore_ascii_case`, no `is_whitespace`, and no `core:temporal`. DEFLATE is bit and byte manipulation with no ASCII classification anywhere in it. A 1.7× change there is not reachable from this diff.

A uniform factor across suites with disjoint hot paths is the shape of a slower runner rather than a code change; a real regression would concentrate where the diff lands. The one mechanism here that *could* touch the parsing benchmarks is the ASCII delegation chain (`u8` → `char`) affecting inlining in `json`, `sqlite_parse`, and `syntax_highlight` — so zlib is the discriminator. If the re-run still shows zlib down ~1.7×, that reading is wrong and I will treat it as this PR's to root-cause.

I cannot measure locally to settle it here: `mise run benchmark-zlib` needs the `wasi-sdk` toolchain, and installing it fails with a 403 on GitHub artifact attestation verification (`GitHub access to this repository is not enabled for this session`). Not retrying a policy denial.

The push to 7df4afb51 has already re-triggered Benchmark, so no manual re-run is being spent. I will act on that result.

---

gfx:
Three findings from the review of 7df4afb51 — two addressed, one I'm declining.

**`i64::MIN` rejected** — real, and already fixed in 04a501360 (pushed before this review landed). The split used a truncating division, which for a negative count leaves a negative remainder and so a decrement still owed to `Instant::new`; the range check ran before that decrement and had to guess at the bound. It floors now, so `whole_seconds` is the count the instant will carry and the `i64` range is the exact check.

**Endpoints untested** — real. The test named the whole `i64` second range while stopping at `MAX - 1`. `MAX` round-trips, so bdefeaf1f sweeps both inclusive endpoints, each at a zero, mid, and maximal sub-second.

**Stale coverage gaps in the WEP** — declining; the two clauses still hold as written.

- *CBOR tag-0 byte encoding*: the gap says "no fixture pinning the CBOR tag-0 byte encoding **beyond the tag itself**". `temporal_test.wado:1056-1063` asserts `bytes[0] == 0xc0` and then round-trips through the decoder — it pins the tag byte and nothing after it, which is exactly what the gap describes.
- *`i64` extremes*: the gap says "no test at the extremes of the `i64` second range **where the renderer produces years far outside four digits**". The new test covers `from_epoch_nanoseconds` round trips at those seconds; it does not render them, and nothing else in the suite does either. I confirmed rendering at both extremes returns without trapping, but an untested behaviour that happens to work is still untested.

---

gfx:
Following up on the Benchmark alert with a measurement rather than the earlier argument.

I built the compiler from `origin/main` (2e580ad31) in a worktree and compiled each benchmark's source with both it and this branch's compiler, then compared the emitted wasm:

| Benchmark | wasm from both compilers | CI-reported ratio |
| --- | --- | --- |
| `zlib` (-O1, -O2, -O3) | **byte-identical** (same SHA256 at each level) | 1.66–1.74× |
| `sqlite_parse` | **byte-identical** | 1.95× |
| `fts` | **byte-identical** | 1.55–1.58× |
| `count_prime` | **byte-identical** | — |
| `mandelbrot` | **byte-identical** | — |
| `json_twitter` | differs, +412 bytes | 1.81× |

Four benchmarks whose compiled output is byte-for-byte the same from both compilers were still reported 1.55–1.95× slower. Identical bytes cannot run 1.8× slower because of a source change, so the alert reflects the measurement environment, not this branch. The baseline it compares against (373.1 MB/s for `zlib/decompress`) was recorded on main at 07:28 UTC; this PR's runs at 11:33, 11:54 and 12:45 measured ~214 each time.

`json_twitter` is the one place this branch changes the output. That is `fpfmt`'s `parse_f64` reaching `eq_ignore_ascii_case`, and the delta is two functions that do not inline at `-O2`. I am tracking that separately as a codegen question — it is real, though it is not what the 1.8× is.

I have not re-run the check: each push has re-triggered it on the new head, and it has reproduced consistently, which is what the table above explains.

---

PR #1908, opened 2026-08-27
feat(gale): run actions and predicates inside a group, and answer every required group position

gfx:
Stage C reaches a group's alternatives, and every group position that must match answers for a token it cannot.

## Actions and predicates inside a group

An action belongs to the alternative it is written in, whether that alternative is a rule's or a group's, and runs when that alternative is taken. One element walker serves every body — rule alternative, group alternative, LR suffix, a transparent group's inlined elements — taking the alternative's own actions and op bindings, so an action's position and the labels it reads come from the body it sits in. Under a repeat it runs once per iteration.

A group's alternatives are a decision like a rule's, so an alt-initial predicate takes the same three paths inside a group: it joins the branch's dispatch condition, excludes the alternative from the scan tournament, and disables it when the decision escalates to the simulator. Where a body is emitted as several shape branches, only the elements are shape-conditional — an action written between them runs in every branch.

## The value channel is the context object

What ANTLR keeps on a context object, Gale keeps in the value struct: both live exactly as long as one rule invocation. Every per-invocation capture an action reads — the rule's own start token, a labeled rule call's token range (`$x.text`), the builder row its node opens at (`$x.ctx`) — is a field of the rule's `Vals` struct, written where the element matches and read wherever an action sits.

That is what makes the reads well-scoped by construction. A rule's body is not one block: an alternative is a block, a repeat body is a block, and a multi-alt or left-recursive rule spreads its alternatives across per-alternative `_alt_N` / `_atom` / `_lr_N` functions. The struct is threaded into every helper and returned from it, so a write anywhere in the invocation is visible to every later read in it — including from a prequel, a sibling branch, or another function. The channel opens for any rule that captures something, not only one declaring args / `returns` / `locals`; grammars with actions off stay byte-identical.

## One dispatch emitter, one no-match policy

Every decision over a group's alternatives — general or rule-reference shape, overlapping first sets or not, standing alone or as a repeat's body — is emitted by one function. Its callers vary the body of a branch and the no-match answer, and that answer is a policy stated at each call site: **Required** reports a no-viable-alternative over the union of the alternatives' first sets; **Guaranteed** rests on a scan the caller already ran, so the last branch takes the rest and there is nothing to report. A gated alternative forces Guaranteed back to Required — a false predicate must not land in an `else` meant for the alternative it excludes.

Guaranteed rests on that scan, and a `+`'s mandatory first iteration is the one position no scan covers. Both answers it needs are emitted before the body, where the mandatory-first flag still reads true: a gated single-alternative body reports at the gate, and an ungated body whose own emit cannot answer — a group of several, or one alternative that matches zero tokens — reports from a first-set test built over the dispatch's own partition.

## Recognizer API

- `p.rule_stack()` — `getRuleInvocationStack()`, read off the tree builder's open nodes rather than new recognizer state.
- `p.rule_tree_at(row)` — `$x.ctx`, rendering one labeled rule call's own node.
- The expected-token set is a generation-time constant, not a runtime query: `getExpectedTokens().toString(getVocabulary())` folds to the rendering codegen computes, through ANTLR's vocabulary names. Codegen answers at a rule prequel, where the set is the rule's FIRST at entry and its FOLLOW at exit; anywhere else, and for a rule whose FOLLOW fixed point is not exact, the body is reported rather than given a set that is not ANTLR's.
- A cast `$ctx` member naming something the enclosing alternative binds resolves to that binding, and `!= null` on it becomes the token-presence test. A name the alternative does not bind stays a loud error.

## Lexer

- A repeat replays iteration by iteration: each is probed match-only first, and only an iteration that matches is walked again with its actions on. An iteration the loop tries and rolls back leaves no trace.
- A lexer member's receiver follows from its body — a method that writes no field and calls no mutating member takes a shared receiver, the rest a mutable one. A predicate is evaluated inside the rule's match helper, which takes the lexer by shared reference because the tournament must not mutate through a losing candidate; a predicate may therefore reach only the shared half, and naming a mutating member from one is refused by name.
- A predicate in an ATN-class rule is refused at generation time. The lexer ATN has no predicate transition, and matching without evaluating it is the one outcome the design does not allow.
- Codegen stages every lexer action and predicate into Wado before emit, whatever the language, so a body the translator cannot emit is a diagnostic with a `ctx` to report to rather than a panic from inside an emitter.

## superClass operations with arguments

An operation that takes arguments gets its signature from its call sites: every site must agree on arity, and each position must be a literal whose Wado type is evident (`this.p("of")` gives one `String`). Sites that disagree, or an argument that is not a literal, leave the base unwired with a diagnostic naming the operation. This runs `TypeScriptParser`'s `{this.p("of")}?` against a hand-written base, and `tests/driver_cst_typescript_test.wado` installs the Wado twin.

## Descriptor corpus

`LeftRecursion/MultipleAlternativesWithCommonLabel_4`, `ParseTrees/TokenAndRuleContextString` and `ParserErrors/LL1ErrorInfo` pass their output compare. The two `[stage_c_todo]` entries that remain are held by ambiguity reporting and error recovery rather than action execution, and the two `[stage_b_todo]` predicate entries differ only in the recovery decision — their triage lines say so.


gfx:
Worked through the review rounds. Summary of what changed and what did not, since several findings recur across rounds.

## Fixed

Generated code that would not compile:

- `rule_uses_rule_span` read each alternative's own actions while the translator substitutes for any body in the rule's frame, so a group-nested `$text` emitted `_rule_start_tok` out of scope.
- `rule_action_bodies` skipped `arg_action`, so `a[$x.text]` emitted `vals._span_lo_x` in a rule with no `vals`.
- `gen_op_repeat_optional_group` destructured the repeat's surface element unpeeled and panicked on a labelled optional group.
- `literal_arg_type` typed any digit run `i32`, so a literal past the range was spliced into an `i32` parameter.

Wrong answers, silently:

- `match_paren` was not template-aware while `match_brace`, in the same file, was. Both delegate to one `match_delim` now.
- `split_top_level_commas` tracked nesting depth alone, so `a["x,y"]` split inside the string — its doc comment already claimed otherwise.
- `find_this_calls` / `rewrite_this_calls` did not descend into an argument, so `this.p(` + "`${this.q()}`" + `)` left `this.q()` with no operation and no receiver.
- `toStringTree` on any `$`-ref returned the ref and dropped the call.
- `getExpectedTokens().toString()` and `.toString(boolean)` folded to the vocabulary string; only the `getVocabulary()` overload is that constant.
- A non-literal Java `assert` message was parsed and discarded. (`:` was not even a Java-action punct, so the branch could not run; it is now, and a non-literal message is refused.)
- The parse-side repeat strategy selectors read the surface element unpeeled while their scan-side twins peeled.

Docs: `antlr4-compatibility.md` named two no-match policies where `GroupEntry` has three; `action.md` claimed rule-span specials live on the value struct (they read a body-entry local) and that an ATN-class predicate is "refused at generation time" (both lexer-predicate paths warn and generate).

## Not changed, with reasons

**Folding an optional token's presence test.** Measured all three shapes. `e INC {…}?` folds — correct, `p.expect(TK_INC)` has already run. `e INC? {…}?` and `e INC* {…}?` are *refused* with `$ctx.INC(...): only the rule's own children are supported`, never folded. `collect_op_binding` returns nothing for `Repeat` (`parser_gen.wado`, "`A?` may not bind at all, unlike `A`"), so an optional or repeated element never reaches the bindings the fold consults.

**Java assertion enablement.** Wado assertions cannot be disabled by design, so translating a Java `assert` to one is strictly stronger than Java, never more accepting. The side-effect half is closed by refusing non-literal messages.

**`render_expected` discarding non-EOF names.** `ids` and `names` are separate accumulators: `names = ["<EOF>"]` resets only `names`, while non-EOF kinds accumulate in `ids` and are appended after. Nothing is lost, and `<EOF>` first matches ANTLR, where EOF is type −1.

**Deduplicating rendered display names.** ANTLR's `IntervalSet.toString(Vocabulary)` prints one name per set element and does not deduplicate, so `A : 'x' ; B : 'x' ;` yields `{'x', 'x'}` upstream too. Deduplicating would diverge from the oracle.

**Guarding `$x.ctx` during recovery.** Rendering empty for a node that was never opened is the correct answer — ANTLR's `$x.ctx` is null there, and `driver_action_group_span_test.wado` already pins that for the span case.

**Aligning `EOF`/`ERROR` with `Eof`/`Error`.** `render_expected` is the only consumer of `vocab_names` and substitutes `<EOF>` before indexing it; an error kind never enters a FIRST/FOLLOW set. Those entries are unreachable.

**Java lexer members.** Deliberate and documented at the call site, and a standing `TODO.md` entry; the reference is reported, not dropped.

**Parse tests in `src/g4/integration_test.wado`.** That file holds the real-world grammars, some of which have no driver. None of the 56 `wado_*.g4` fixtures has an entry, mine included — for these the driver test already proves the grammar parses, since it cannot generate a parser otherwise.

## Verification

1542 files compile, 2760 tests pass, 0 failures, including the ANTLR4 descriptor corpus and the oracle-pinned real-world grammars.

---

PR #1910, opened 2026-08-27
perf(ownership): widen value-copy elision; gale-gen 13% faster

gfx:
Four rules in the value-copy machinery decide whether a materialization deep-copies or hands over storage. Each one here reaches a shape it can prove, and gale-gen — a generator that builds an IR and walks it — stops paying for the ones it does not need.

## What holds

**A borrow that ends before the move does not pin its referent.** A `match` / `if let` over a `&place` scrutinee is the same transient borrow a `&` call argument is, so `if let Repeat(r) = &op { … }; ops.push(op)` moves `op`. What the arms bind names the scrutinee's storage, so an escape recorded against a binding closes back over the referent: a reference reaching anything that outlives the match — a `stores` callee, an aggregate literal, a variant payload, a global, a write through a place, the result — pins the scrutinee exactly as `&op` there would.

**Where a stored reference lands is two questions, not one.** `stores.rs` separates the position a callee routes into its return value from one reaching a global or the caller's own storage. A caller assumes the union; the fixpoint does not, so an iterator holding a reference to what it walks no longer makes every `&List` parameter a stored one the moment the body iterates it — a `collect()` that drops the iterator stores nothing. A builtin has no body to read, so a reference-typed call result carries every argument: `array_get_ref` aliases its container, and `List::index_ref` is that call.

**A builtin is not an opaque callee.** `confine.rs` classifies by the callee's own reference where the table it builds is silent. `a[i]` is a builtin call, so every function that indexes an array — `String::cmp`, and through it `TreeMap::find_index` — confines its parameters, and a map lookup compares its key in place.

**A call-graph component settles as a unit.** Return conventions assume every member of a strongly connected component returns owned and drop the ones a walk of their returns refutes. A cycle whose members return each other's result — `lower_element` yields `lower_group`'s value and `lower_group` yields `lower_element`'s — has no base to build up from, and the returns that avoid the cycle decide it either way.

**A type's identity is the declaration it names.** One declaration reaches the type table under more than one `TypeId`; `is_string` / `is_list` / `is_result` / `as_option` compare the `def` a nominal type carries (WEP 2026-08-12) rather than the node one id is registered under. A `String` copy is therefore right-sized to `used` — a builder's slack is not part of the value.

## Measurement

gale-gen at `-O2`, best of five interleaved rounds on one machine:

| build       |    ms/iter |  throughput |
| ----------- | ---------: | ----------: |
| `main`      |    114.915 |  307.1 KB/s |
| this branch |     99.709 |  359.1 KB/s |

The emitted component is 26 KB smaller. json-twitter serialize and deserialize each gain 2–4%; sqlite-parse, syntax-highlight, cbor-twitter, count-prime, mandelbrot and sieve are unchanged within noise.

Nine e2e fixtures pin the shapes, including the ones that must still copy: an arm binding handed to a `stores` callee or kept in an aggregate literal, a `stores` callee with no result to route through, and a caller holding a `&T` into a list it would otherwise move.

## Known gap

A `stores[p]` on a value-returning body is read as naming the result, which is not derived — closing it needs `carries` to follow a reference through the adapter chain first. Written up with the measurement that fixes the order in [WEP: Ownership Analysis](docs/wep-2026-05-21-resource-ownership.md).

`benchmark/perf.md` records the `-O2` inline budget as a dead end: raising it wins on gale-gen, cbor and syntax-highlight, costs json-twitter's serialize 17%, and grows `-Os` output 41%.


gfx:
On the two findings from the last review:

**Field-level escapes** — real, and fixed in 46fc0e0. `&s.field` in a persisting position is the spelling `walk_expr` already handles, and it handles it with `borrow_top_field`; asking the escape helper first recorded a whole-local escape that nothing downstream could narrow, so every other field of `s` lost its move. The helper now leaves that spelling alone and names the field for the ones it does take. `mise run test` and `mise run test-wado` (4917 tests) green; the golden WIR corpus is unchanged, so it costs nothing it was buying.

**One visitor resolver** — agreed as a direction, and out of this PR's scope. `is_borrowed_place` is a fifth hand-written walk over the same chain, next to `alias_root`, `borrow_read`, `borrow_top_field` and `clean_root`, four of which predate this branch. Unifying them means one resolver returning root, borrow provenance and top field together, and rewriting each caller onto it — `place.rs` already owns a `Resolver` for the sharing analysis, so that is where it belongs rather than in a widened perf PR. Filing it against [WEP: Ownership Analysis](docs/wep-2026-05-21-resource-ownership.md) is the honest next step; happy to take it as a follow-up.

For the record on the earlier **Benchmark** failure: it alerted on 31 rows at once — every benchmark at every `-O` level, ~1.5–2× — including `mandelbrot`, which is pure float arithmetic this diff cannot reach, while `gale-gen`, the one it speeds up, was not in the list. That is a slow runner, not a regression; the check has since passed on two later heads.

PR #1913, opened 2026-08-27
chore: remove the unwired CM thread builtins

gfx:
Wado declares no Component Model `thread.*` built-ins, and `docs/research-wasm-stack-switching.md` records why not.

## The declarations were dead

Eight `#[canonical("wasi", "thread-*")]` functions in `lib/core/builtin.wado`, plus `internal resource Thread` and `internal resource SuspendedThread` in `lib/core/prelude/types.wado`, had no path to codegen:

- `CanonicalIntrinsic::from_import_name` (`canonical.rs`) has no arm parsing `thread-*`, so they never lower to a `canon`.
- `cm_binding_function` (`synthesis/cm_binding/resource_rewrite.rs`) has no arm for them, so the `#[cm]` methods never bind.
- Both resources are `internal`, and nothing in the stdlib, tests, or examples calls them.

Regenerating the full WIR golden corpus after removal produces a zero-line diff, confirming nothing reached codegen.

## Why not wire the current set instead

Four `thread.*` opcodes are reassigned in place between the pinned generation and upstream:

| Opcode | `wasmparser` 0.252 (pinned) | Upstream |
| ------ | --------------------------- | -------- |
| `0x28` | `thread.suspend-to-suspended` `[i32] -> [i32]` | `thread.resume-later` `[i32] -> []` |
| `0x2a` | `thread.unsuspend` `[i32] -> []` | `thread.suspend-then-resume` `[i32] -> [i32]` |
| `0x2b` | `thread.yield-to-suspended` `[i32] -> [i32]` | `thread.yield-then-resume` `[i32] -> [i32]` |
| `0x2c` | `thread.suspend-to` `[i32] -> [i32]` | `thread.suspend-then-promote` `[i32] -> [i32]` |

The rest of the family (`0x0c` `thread.yield`, `0x26` `thread.index`, `0x27` `thread.new-indirect`, `0x29` `thread.suspend`) holds, and `0x2d` `thread.yield-then-promote` exists only upstream.

`0x2b` and `0x2c` keep their signature across the rename, so a declaration written against one generation validates under the other and means something else. `0x28` and `0x2a` differ in result arity and would at least fail validation; `0x28` also changes its immediates, reading a `cancellable` byte in the pinned generation and none upstream. The set moved too: `thread.suspend-to` is gone, `thread.{suspend,yield}-then-promote` are new.

Two further gates, recorded in the doc. The Component Model threading feature — `CM_THREADING` in `wasmparser`, `wasm_component_model_threading` in wasmtime, distinct from core Wasm `THREADS` (on by default) — is off by default in both, and wasmtime's own docs call its support "very incomplete". And `thread.new-indirect` requires the start function to take exactly one `i32` in both generations, so a GC language needs a side table to hand a spawned thread anything else; the spec defers loosening that until the Canonical ABI is extended for GC.

WASI itself is not a factor: 0.3.1 ships no threads interface, and nothing in the tree references the retired Preview-1 `wasi:threads`.

## Docs

`docs/research-wasm-stack-switching.md` § "CM Thread Primitives" names the current built-in family and the gates above. Its wasmtime status appendix moves to the pinned v47.0.3, with corrected config-flag names and a CM-threading row.


gfx:
Standing down on the latest review round. Status of its three points:

**`thread.new-indirect` `i32` limit — not scoping this; the suggestion is incorrect.** The claim is that upstream permits a single `i64` closure parameter with CM GC, so the limitation should be scoped to the pinned generation. Both validators reject anything but one `i32`:

- `wasmparser` 0.252 — `if func_ty.params() != [ValType::I32] { bail!(...) }`
- upstream — same check, message `start function must take a single 'i32' argument (currently)`

Neither has an `i64` path. The spec's `(func (param i64))` variant is gated on 🐘 (memory64), not GC, and is unimplemented in the validator either way. Scoping the sentence to the pinned generation would make it wrong, so it stays as written — it holds for both.

**Formatting — already clean.** `format-md` reformats nothing and `cargo fmt --all --check` exits 0 on the current head.

**Splitting the paragraph — skipping.** The two gates are already enumerated as "The first is…" / "The second is…", so a split is a wash.

The earlier rounds' findings were real and are fixed: per-opcode signatures in `44480633f`, table scope in `c2d009216`, gate naming in `e53bd259e`.

---

PR #1915, opened 2026-08-28
fix(lsp): keep the receiver's argument slot in path-form parameter hints

gfx:
A method call written in path form passes its receiver positionally, so the
declared `&self` occupies argument slot 0 and the remaining parameters line up
one slot later:

```
Scale::scaled(&p, factor: 2, bias: 3)   // the receiver goes unlabelled
p.scaled(factor: 2, bias: 3)            // the dot form omits it from `args`
```

The two shapes get their own name lists. `dot_form_param_names` drops the
receiver, since it is not in `args`; `path_form_param_names` keeps its slot,
named `self` so `emit_arg_param_hints` skips it. `self` is reserved in Wado and
the symbol table spells a resource method's receiver the same way, so the one
skip covers both the AST route and the symbol-table route, and a `debug_assert`
states the naming invariant it rests on.

Two more defects along the same path:

- `InlayHint` serializes `paddingLeft` / `paddingRight`, the names LSP reads.
- A recovered `b'ab'` underlines the whole literal, `b` included.
  `scan_char_raw` takes the opening position from its caller, so the byte and
  char forms anchor their error spans the same way.

## `wado query inlay-hints`

A new query kind splices the labels into the source at the anchors a client
would render them at, which is how a misplaced anchor shows up as something to
look at rather than a number to check by hand:

```
  116 |     out.sort_by(‹cmp: ›|a: &DirectoryEntry, b: &DirectoryEntry| a.name.cmp(‹other: ›&b.name));
  119 |     assert render(‹template: ›"こんにちは、{{名前}}さん！", ‹context: ›&ctx) == "こんにちは、太郎さん！";
```

`--json` prints the raw hints, positioned in the LSP's default UTF-16 encoding.
`QueryKind::ALL` with `name()` / `desc()` feeds the parse arm, the help text and
the "Available:" error, so a kind is declared once.

`text::character_to_byte_offset` converts an LSP `character` against a line the
caller already holds, keeping the splice off the document-wide rescan that
`lsp_position_to_line_col` performs. `Range::WHOLE_DOCUMENT` names the range a
caller with no viewport asks a range-filtered query for.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YN2Pewa2xb9rhf9G43Y8rK)_

### DIFF
--- a/.claude/skills/wado-cli/SKILL.md
+++ b/.claude/skills/wado-cli/SKILL.md
@@ -177,7 +177,14 @@ wado query definition --symbol core:cbor#CborDeserializer.peek_byte
 wado query references --symbol core:cli#println --base example     # all uses (workspace)
 wado query hover --line 5 --column 10 file.wado                   # position-based
 wado query diagnostics file.wado                                  # errors/warnings
+wado query inlay-hints file.wado                                  # hints, spliced into the source
 ```
+
+`inlay-hints` takes a file, not a symbol, and prints every line that carries a
+hint with the labels spliced in at the anchors an editor would render them at
+(`let x‹: i32› = add(‹a: ›1, ‹b: ›2);`), in the LSP's default UTF-16 encoding.
+That is how to check anchor placement — against `example/`, say — without
+reading positions off a list.
 
 Common options:
 
--- a/.claude/skills/wado-cli/SKILL.md
+++ b/.claude/skills/wado-cli/SKILL.md
@@ -182,9 +182,9 @@ wado query inlay-hints file.wado                                  # hints, splic
 
 `inlay-hints` takes a file, not a symbol, and prints every line that carries a
 hint with the labels spliced in at the anchors an editor would render them at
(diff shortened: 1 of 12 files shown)


gfx:
On the merge-risk note about null padding fields — checked both surfaces against a real stdio session. The protocol response is fine; the `null` was in the CLI's own JSON.

`textDocument/inlayHint` on `example/hello.wado`:

```json
{"position":{"line":5,"character":12},"label":"message: ","kind":2,"paddingRight":true}
```

`skip_serializing_if = "Option::is_none"` drops the unset flag, so no `null` reaches a client.

`wado query inlay-hints --json` built its object with the `json!` macro, where `Option::None` lands as `null`. That output exists to show what a client receives, so 2c57876b makes it report padding the way the wire does — present when set, absent otherwise.

Nothing guarded either shape, so both are pinned now: `unset_padding_is_absent_from_the_wire_form` covers the serialized `InlayHint`, and `query_inlay_hints_json_output` asserts a type hint carries neither key.

---

PR #1918, opened 2026-08-29
perf(optimizer): converge the -O3 fixed-point loop

gfx:
The NIR fixed-point loop converges on every source in the tree, and says so when it does not.

## The loop

`-O3` runs at most 20 iterations. Reaching that cap means a pass reported a change it did not make, or is taking one step per round where a single walk reaches its fixpoint — so the run logs it at debug level, naming the passes still reporting changes, and `debug_assert`s. Only `-O3`'s own cap carries that claim: `-O1`, `-O2`, and an explicit `--optimize-iterations` set budgets and say nothing about convergence.

Convergence is read off one vector — the passes that reported a change in the iteration just run. Empty ends the loop; non-empty after it is what the cap report names.

## The two passes that held it open

`sroa_param` reports its change from what it rewrote: the callers in `touched`, plus the global initializers, which have no slot there. A clone it reused rather than minted, with every call site already pointing at it, is a run that changed nothing.

`const_fold` treats an `if` whose condition the env decides as what it is — exactly one arm runs, so the walk enters it with the env intact and keeps what that arm writes. A derived `Inspect` is a chain of `if wrote_field { ", " } else { " { "; wrote_field = true }`, one link per field, each writing the next one's condition; the whole chain folds in a single walk.

## Cost

`wado test -O3 package-gale` — 1542 compiles, the heaviest `-O3` workload in the tree — spends 6992s of compile CPU, 8:37 wall on 16 cores. Compilation is 93% of that job's CPU, and the fixed-point loop is 85% of a compile.

## Tooling

`WADO_TRACE=const_fold` names each function `const_fold` changed, and the optimizer-debug skill carries the workflow for a loop that will not settle: the cap report names the pass, `WADO_TRACE=opt_loop` narrows it per iteration, and a diff of `WADO_DUMP_PASS_AFTER` across two rounds separates an oscillation from a pass taking one step per round.

The profiling skill's analyzer reads its symbolicator output unbounded — `spawnSync`'s default 1 MiB `maxBuffer` silently drops every address past the cut back to a raw `[lib]+0x…` frame.


gfx:
On the third CodeRabbit finding (`sroa_variant_return.rs` 473-487, "Discarded context, propagate you must") — the reading of the code is right, but it describes a missed optimization rather than a defect, so I have not pushed a change for it. The other two are fixed in 39029e2.

`discarded_exprs` marks the operand of a non-tail `StmtKind::Expr` and stops there, so a scalarized call in the result tail of a discarded `Block` / `If` / `Match` / `Switch` is still reboxed. That rebox builds a value nothing reads — wasted, exactly as the finding says.

It is not unsound. Reboxing is always correct; the unsound direction is the other one, skipping a rebox where a consumer does read the variant, which reaches codegen as a tuple where a variant is expected. That is why the predicate is conservative and why `debug_assert_call_sites_rewritten` reads the same site list the rewrite does.

Nor does the shape occur in the corpus today. The waste it causes is a rewrite each fixed-point iteration undoes, and this PR makes exhausting the `-O2` / `-O3` cap a `debug_assert` — which does not fire over 12542 e2e fixtures at every optimization level, 1542 `package-gale` files at `-O3`, or the 1737 modules in the tree at `-O2`.

The patch I would write closes `discarded_exprs` downward through value-yielding constructs: when a discarded expression is a block, its tail is discarded too; when it is an `if` / `match` / `switch`, so is each arm's tail. That is sound (a value thrown away throws away whatever produced it) and mechanical, but it widens a predicate whose wrong answer miscompiles, for no measured gain. Happy to add it with a fixture if you would rather close the class now than leave it as a known conservatism.

gfx:
`Test (Wado)` is red on cc07988 (and `Test` with it, being the aggregator). It is the known `core:temporal` timeout flake, not this PR's:

```
not ok - a nanosecond span beyond the i64 range traps rather than wrapping (4.55s)
  message: test timed out after 5000ms (use #[timeout_ms(N)] to increase)
```

`mise run test-wado` passes locally on this head — 1749 modules compiled, 5019 tests, 0 failures — over a wider set than CI runs, which excludes `package-gale`. Every other check on this head passes, including all five e2e levels and `Test (Gale O3)`. The last push before the failure was documentation only.

The margin is what makes it flake rather than fail: that test takes ~580ms locally against a 5000ms budget, and the CI run that tripped it took 4.55s — an 8x runner slowdown, not a regression here.

I am re-running the job once. Separately, ~580ms is itself too close to the budget for a three-line test, so I am chasing that in a worktree off main rather than widening this PR: it reproduces deterministically, and it is a property of the module rather than the test — the identical body costs 2ms in a small module and 583ms inside `temporal_test.wado`, while the same call one unit coarser (`Unit::Second`, where the assert passes) costs 18ms in that same module.

gfx:
The re-run failed the same way — same test, same 4.55s — so it is not a random flake, and I am not re-running again.

It is still not this PR's. Measured on the same machine, the same test costs 576–587ms built from `main` (ac12c6d) and 595ms built from this branch: no regression here. Every other check on this head passes.

I have the root cause. The test's `#[expect_trap]` fires a power-assert in `core:temporal`:

```wado
assert step == 0 || magnitude / step <= 9_223_372_036_854_775_807;
```

`magnitude` is an `i128`, so rendering the failed assert's operands formats an `i128` in decimal — one division per digit, ~20 of them. And `u128::div_rem` (`core/prelude/int128.wado:223`) is binary long division: 128 iterations of shift / `get_bit` / compare / subtract on software 128-bit values, per division. A guest profile over the file puts 75% of self time in `u128::get_bit`, under `u128::div_rem` → `u128::fmt_decimal_magnitude` → `i128^Inspect::inspect`, under this test.

So the 580ms is `i128` division, and every `i128` / `u128` division in the language takes that path. That is worth fixing on its own account, and it is what makes this test sit close enough to the 5000ms budget for an 8x-slower runner to cross it.

I am fixing it separately rather than widening this PR. This PR needs `main` to carry that fix and a merge; alternatively `#[timeout_ms]` on the one test would unblock it now, at the cost of leaving the cause in place. That call is the author's.

PR #1926, opened 2026-08-30
feat(cm): give stream copies their Component Model result

gfx:
A Component Model copy reports how it ended. `Stream::read` and
`StreamWritable::write` hand that report back, so the end of a stream is a value
the caller receives rather than a shape it guesses — and a stream carries data
between two components without reaching the trap that an inference from an empty
list walks into.

The exact signatures are in
[WEP: Stream Copy Results](docs/wep-2026-08-30-stream-copy-result.md); this
description spells no generic arguments, because the PR-body pipeline strips
them.

## The copy result is a type

`CopyResult` is an enum of `Completed`, `Dropped` and `Cancelled` — the three the
Canonical ABI names. A `bool` would collapse the last two, which the CM keeps
apart: only `Dropped` ends the stream.

`Stream::read` returns a `StreamChunk` — the elements it copied, paired with that
`CopyResult`. `StreamWritable::write` returns a `StreamWrite` — the count the
reader took, paired with the same. Each is one canonical call, faithful to
`CanonicalABI.md`: a copy can carry elements and the peer's drop together, and a
reader may take a prefix. A zero-length write stays expressible, because it is
the CM's readiness signal rather than a write of nothing.

The loops sit on top, in Wado: `Stream::read_to_end` drains to the writer's drop,
`StreamWritable::write_all` finishes a list, and `write_raw_all` does the same
for elements already held in one array.

`write_raw` and `write_raw_all` hand the backing array to the canonical as it
stands, which lines up with the CM buffer only for bytes; a wider element is a
diagnostic naming `write_all`. `cm_binding_function`'s stream arms are the
hand-written `u8` path, and the rewriter checks the element rather than trusting
that everything else was parameterized above it.

## A generic wrapper over a CM primitive

`read_to_end` and friends are `impl` bodies over a type parameter, so their
`#[cm]` calls name a concrete payload only once monomorphize mints the instance.
The payload-driven half of CM binding synthesis therefore runs over the
`FlatPackage` as well, and the helpers it mints call generics of their own, so it
runs to a fixpoint with a resumable `Monomorphization` rather than strictly
after it.

Effect dispatch runs first in that half: a handler installed for a byte stream
claims `read` wherever it is called from, including from inside the stdlib
wrapper.

A generic resource is a declaration like any other, so `base_type_name`,
`impl_receiver_key`, `fq_base_type_name`, `generic_dispatch_components` and
`get_struct_name_from_type` head it the way they head a generic struct, and
`wado doc` lists what an `impl` block adds to a resource.

## `async func` across a component import

A dependency's `async func` is an `async fn` returning an `AsyncCall`, lowered
with `canon lower async`, so a caller writes into a stream while the callee's
subtask reads it. An interface method and a world-level function are the same
function at two levels of the world: one value-type engine on the component
type, one core import signature.

## Behaviour this pins

- `cm_stream_binding_shared_by_two_call_shapes` — one binding helper per payload,
  whichever call shapes reach it.
- `effect_handler_resource_stream_stdlib_wrapper` — a handler claims the
  primitive through `read_to_end` / `write_all` / `write_raw_all`.
- `error_stream_write_raw_non_byte` — the diagnostic for a raw write with no
  lowering.
- `stream_write_all_empty` — writing nothing issues no copy.
- `cm_async_value_import` — a stream round-tripping through an imported
  interface, an imported world function, and a generic body.


gfx:
## Review response

Each finding was reproduced against the current code before acting. What follows is what changed and what did not.

### Fixed — a defect reproduced first

- **A generic body's stream payload escaped both scans.** Chasing a comment on `cm_binding.rs` turned up a compiler bug: a payload named by a type parameter is skipped by the pre-monomorphize scan, and nothing judged the instances it deferred. Written without a generic helper the program is a clean diagnostic; routed through one it reached WIR build with no binding to call and **panicked** (`[WIR] unresolved Call: name="cm_stream_pair"`). `reject_unresolvable_payloads_monomorphized` runs the same scan over the `FlatPackage`, ahead of the post-monomorphize rewrites as the pre-monomorphize scan runs ahead of its own. Fixture `error_stream_generic_payload_unclassifiable`, verified red then green at O0/O1/O2/O3/Os. The suggested patch — passing non-concrete payloads through `stream_payload_rejection` — would not have worked; they are not judgeable before monomorphize.
- **`parameterize_stream_cm_name` decided `u8` by name.** A `type MyByte = u8` stream could drop under one canonical and read under another. Now asks the shared `is_u8_stream_element`.
- **`serve_big_body` ignored the write result.** That fixture exists to model a stalled consumer; on client disconnect `write_all` returns `Dropped` and continuing traps.
- **`read_body_from_request` lost invalid UTF-8.** Decoding the whole body in one buffer made a strict `from_utf8` turn any bad byte into an empty body. Now `from_utf8_lossy`, which the same handler already uses for header names and values — and which also fixes a character straddling a chunk boundary.
- **`fq_type_name_spelled` headed a `GenericResource` with `builtin`.** `TypeHead::Builtin` compares by name string, `DeclaredHead` by `DefId`; `AGENTS.md` requires identity by `DefId`, never by name.
- **`StreamChunk` / `StreamWrite` literals were spelled by hand** rather than read from the compiler-item registry.
- `wasi_filesystem_dir_ops` took a single `read(100)` where a prefix is legal; the catalog copy loops stopped on the read result only; five fixture headers named `write()` / `read()` while their bodies call `write_all` / `read_to_end`; plus MD040 fence labels and the byte-only constraint on `write_raw`.

### Reverted — a change that guarded nothing

`stream_read_multiple` gained an `assert items == 0`. The writer drops before any read there, so `chunk.items` is empty by construction and the assertion could not fail. The property it was meant to pin — a terminal chunk carrying elements — is exercised where it can actually happen: `cm_async_value_import` streams between two components and `read_to_end` takes the items before it reads the result. A degenerate version passes CI and reads as coverage, which is worse than leaving the test alone.

### Skipped, with reasons

- **Quadratic `write_all` / `write_raw_all`.** Real, and stronger than the finding states: each iteration re-lowers the untransferred tail, where the compiler's old drain lowered once and advanced a pointer — so it is a regression, not merely a missing optimization. Closing it needs an internal `core:rt` drain, which in turn wants the raw family byte-only at the type level rather than by diagnostic. That is a public-API decision, so it went to the author, who chose to keep it recorded in the WEP's Known gaps for now.
- **Zero-length write guards in `rt.wado`.** A zero-length CM write is the readiness signal that blocks until a reader rendezvouses, not a write of nothing. Guarding it in the primitives would remove something the language is meant to be able to express; the guard belongs in the `write_all` wrappers, where it is.
- **`_done` future discarded at the gale read sites.** Pre-existing and unrelated to copy results — the previous `chunk.len() == 0` loops did not check it either.
- **A two-party terminal-chunk case in `stream_read_multiple`.** Not constructible there: a CM write blocks until a reader rendezvouses, so writer and reader must be separate tasks.

### The two CI failures on this branch

Both were this PR's own and are fixed:

1. **Five optimizer fixtures at O2/O3.** Not the loop — its assertion. `write_all` and `write_raw_all` each opened with an assert restating what the two lines above already enforced. Power-assert captures its operands, so those two lines put a 145-byte panic string plus `Display::fmt` / `Inspect::inspect` calls into every program that prints, which pushed `String::push_str` past the inline threshold.
2. **`error_stream_generic_payload_unclassifiable` at every level.** The harness compiles fixtures into the CLI world; the fixture carried only a `test` block and died on the missing `export fn run` before reaching the check.

`mise run test` passes locally (7870 tests, 0 failures). `mise run test-wado` is still running.

---

PR #1927, opened 2026-08-31
perf(optimizer): outline cold regions, and retune the inline thresholds

gfx:
`nir/cold_outline` moves what a `cold_path()` marker opens into a function of its
own, so `inline`'s cold discount describes the callee instead of promising a
split nobody made. With the two prices that opens, the inline thresholds are
retuned: `-O2` to 16 and `-O3` to 26.

`-O3` exists to shake NIR into new shapes, and above 28 it stops doing that — on
the Gale CSS3 parser 28, 30 and 32 all emit ~959KB from the same candidate set
while 26 emits 719KB. The Gale corpus at `-O3` finishes in 9 minutes.

## Two prices

A body costs one thing to run and another to hold, and they part company where it
keeps a cold arm. `inline_cost` is what the threshold judges; `inline_size` is
what `InlineBudget` spends, against a cap `--optimize-inline-growth` sets. No
level sets one, because `cold_outline` keeps the two in agreement.

## Compiler fixes

- **A cold region may not write what the enclosing frame keeps.** Outlining a
  region that assigns a by-value parameter put the write in the helper's frame.
  `free_vars` now names each local's crossing and asks one question of every
  crossing that leaves the storage behind.
- **A boxed `&mut` parameter still writes the caller's storage.** `&mut x` does
  not survive to NIR — lowering rehomes the local into a `Box<T>` — so a call
  node taking `is_mut` from the argument's syntax said "by value" about a slot
  the callee writes through. Const folding then read the field back as the
  literal the box was built with. Every argument now takes the stronger of the
  syntax and the callee's declared modes.
- **A shared borrow reads, wherever it appears.** `expr_readonly` had no arm for
  plain `&`, so a borrow the inliner moved from argument position into a `let`
  counted as a possible write and `const_object_globalization` declined a hoist.
- **A diverging callee is recognised by its type**, not by a name list that
  reaches neither `unwrap_failed` nor an outlined panic guard.
- **A cold helper's signature is its own**: it returns `()` or `!`, so an
  inherited `MultiValue` return ABI would put result types on a signature
  nothing produces.

## Measurement tooling

`benchmark/ab.ts` compares two arms and decides each row by whether their ranges
overlap, not by the delta, leaving the C/Rust/JavaScript rows in as the host's
control. `mise run benchmark-baseline` builds `origin/main`'s compiler once,
keeps the binary under its commit and drops the worktree; `WADO_BIN` runs it
through this tree's harness, so an A/B varies only the compiler.

## Accepted regressions

Two rows are down against `origin/main` and stay down. `sieve` costs 4.5% to
`cold_outline` for no reason the IR shows — the hot loops are identical in WIR
and in the emitted Wasm — so the cause is below the Wasm. `json-catalog ser`
costs 3% to the threshold, and is the row that turns first as it climbs. Both are
recorded where the code that causes them lives.


gfx:
Six of the seven findings were real and are fixed in d958805 — thanks.

The provider one was the best catch: the `CompilerOptions` literal says in the comment right above it that it inherits the parent's optimization settings, and then took `OptOverrides::default()`. Predates this PR; one line now.

`ab.ts` was mine and wrong exactly as described — a row one run skipped left a one-sample arm, and a point range never overlaps, so it read as a real difference off a single measurement. Verified against this branch's own logs: trimming one head log turns every row from `SLOWER` to `1-run`.

**Not taken: `valueless_blocks` never records the root block.** Confirmed by repro — a `cold_path()` at the top level of a unit-returning function produces no helper. It is a missed region rather than an unsound one, and taking it changes generated code for a shape the `never`→`cold_path` rule makes common, so it wants its own e2e and Gale O3 validation rather than a late push onto a green PR. Recorded with the pass's two other open items in its module doc.

The `set -e` finding on `AGENTS.md` held in principle but not in this harness (failed runs do write their `exit=` marker). Took it anyway in a form that stays one line: `cmd > run.log 2>&1 && s=0 || s=$?`.

PR #1929, opened 2026-08-31
perf(int128): divide by Knuth Algorithm D, render 19 digits per division

### LINKED ISSUE #1925: perf(int128): 128-bit division is still binary long division
## State

`core/prelude/int128.wado` already lowers the cheap half to native Wasm wide arithmetic:

| operation | implementation |
| --- | --- |
| `+` `-` | `builtin::i64_add128` / `i64_sub128` |
| `*` | `builtin::i64_mul_wide_u` plus the cross terms |
| base-2 / 8 / 16 rendering | shift-and-mask, no division at all |
| **`/` `%`** | **binary long division: 128 iterations** |

Wasm has no 128-bit divide, so division is the one that stays in software — but it is there in its most naive form, one bit at a time with `get_bit` / `set_bit` / compare / subtract over the software 128-bit representation.

#1918 added a fast path for the case where both operands fit in 64 bits, which is what a decimal render and a unit conversion do: the `core:temporal` test that CI kept timing out went from 588ms to 8ms. Everything outside that case still runs the 128 iterations.

## What is left

1. **Knuth Algorithm D over 32-bit limbs for the `high != 0` case.** Four limbs means four to eight iterations rather than 128. `core:temporal`'s `floor_div_i128` was 9.9% of self time in a guest profile of `temporal_test.wado`, and every large-magnitude `i128` division pays the same.
2. **A 128 ÷ 64 path.** The divisor fitting in 64 bits while the dividend does not is what the leading digits of a decimal render hit. Two limb steps. A subset of (1), so it falls out of doing (1).
3. **Render decimal in 19-digit chunks.** `fmt_decimal_magnitude` divides by ten once per digit and does it twice — once to count digits, once to write them. Dividing by `10^19` to peel a `u64` chunk cuts ~40 divisions to ~2, and it is the same trick the base-2/8/16 path already uses.

## Do the optimizer first?

While measuring (1)'s ground, an early `return` of a freshly built `[u128, u128]` measured **7.5x slower** than assigning into the locals that a single `return` at the end already reads — 577ms against 4350ms on `PlainDateTime until`, for identical arithmetic. Same inputs, same operations, only the shape of the return differs.

A tuple return should be scalarized either way. `nir/sroa_variant_return` and `nir/multi_value_return` exist to flatten exactly this, so a return that builds its tuple in one arm and reads locals in another looks like a gap in what they reach rather than something the library should be written around.

That is worth settling before (1), not after: Algorithm D carries many intermediate values through several arms, so it would keep landing on whichever shape the optimizer happens to handle, and the library would end up written to a rule nobody has stated. The measurement above is a starting point but not a diagnosis — a WIR diff of the two shapes is what would name it.

## Notes for whoever takes it

- `int128_test.wado` has 118 tests, including the boundaries (`u128::MAX`, dividend above `i64::MAX`, divisor of 1). Algorithm D goes wrong at limb-carry boundaries, so widen that set **before** rewriting.
- `wado test --profile guest,p.json,1 <file>` (added in #1918) profiles the guest over a file's test blocks; `.claude/skills/wado-performance` reads the output.

gfx:
`u128::div_rem` divides by Knuth Algorithm D over 32-bit limbs. A divisor inside 64 bits takes one native division for the dividend's high limb and one wide step for the rest; a divisor past 64 bits leaves the quotient in a single word, estimated from the normalized leading divisor limb over a halved dividend and settled by one correction. `div_rem_wide` is the 128÷64 step Wasm has no instruction for.

Decimal rendering peels 19 digits per division — 10^19 is the largest power of ten inside a u64 — so a full-width value spends two 128-bit divisions on its digits, counted and written. `count_digits_u128` counts by the same chunks.

One `write_decimal_digits` serves every integer width: it takes a `u64` and a digit count, padding with the leading zeros a decimal chunk below the leading one carries. `abs_to_u64` supplies the magnitude of a signed value, `MIN` included, so `u32`, `i64` and `u64` reach the shared path by widening.

`u128` no longer exposes `get_bit` / `set_bit`.

Throughput under `wado run` at `-O2`, over pseudo-random operands of each shape:

| operand shape | throughput |
| --- | --- |
| 128 ÷ 128 | 5.70 M ops/s |
| 128 ÷ 64 | 10.02 M ops/s |
| full-width decimal render | 3.32 M ops/s |

`int128_test.wado` covers the limb boundaries Algorithm D steps over, a `q * d + r == n ∧ r < d` sweep over 300 pseudo-random operand pairs spanning every width combination, the 19-digit chunk boundaries of `to_string`, and a `to_string` / `from_str` round trip.

Closes #1925

https://claude.ai/code/session_01Kf59d3ZR7RJDZ95X8WDW9E


gfx:
On the flagged removal of `u128::get_bit` / `set_bit`: it is deliberate, and the repository owner has confirmed keeping it.

The two helpers existed for the binary long-division loop this PR replaces. Nothing in the repository calls them, and `docs/stdlib-core-prelude.md` is regenerated without them.

The suggested deprecated compatibility wrapper is not available in this language: Wado has no user-defined attributes, and no `deprecated` marker exists in `docs/spec.md`, `docs/cheatsheet.md`, or anywhere in `wado-compiler/lib/`. So the choice is between keeping unused public API and removing it; the PR description records the removal as a user-visible break.

---

PR #1935, opened 2026-09-01
fix(elaborator): select a newtype's own blanket over its base's (#1932)

### LINKED ISSUE #1932: A newtype takes its base's blanket impl over one whose bound the newtype itself satisfies
`impl<T: Bound> Trait for T` applies to a newtype that carries its own `impl Bound for Newtype` — until a second value blanket exists that the *base* type satisfies. The receiver then peels to its base and takes the base's blanket, silently.

## Repro

```wado
trait Limit {
    fn limit() -> i32;
}

trait Describe {
    fn describe(&self) -> String;
}

impl<T: Limit> Describe for T {
    fn describe(&self) -> String {
        return `limit:${T::limit()}`;
    }
}

impl<T: ReflectStruct<FieldTypes = [..F]>, ..F> Describe for T {
    fn describe(&self) -> String {
        return `struct:${ReflectStruct::<T>::type_name()}`;
    }
}

type UserName = String;

impl Limit for UserName {
    fn limit() -> i32 {
        return 3;
    }
}

test "the newtype's own blanket applies" {
    let u: UserName = "abc" as UserName;
    assert u.describe() == "limit:3";
}

__DATA__
{"test": {}}
```

Expected `limit:3`. Actual:

```
    condition: u.describe() == "limit:3"
    u: "abc" as UserName
    u.describe(): "struct:String"
```

`T` binds to `String`, not `UserName` — the value reached the `ReflectStruct` blanket through its base, which is a struct in the stdlib.

## Why this is a bug

[Trait Derivation Policy](docs/wep-2026-06-25-trait-derivation.md): "A hand-written impl always wins over a derived one — a newtype most of all, since it exists to be a type distinct from its base." Here the `Limit` blanket is hand-written and applies to `UserName`, while the `Describe` the newtype ends up with is derived for `String`.

Three checks isolate the trigger:

- The bound genuinely holds on the newtype. `fn limit_of<T: Limit>() -> i32 { return T::limit(); }` called as `limit_of::<UserName>()` returns `3`.
- The blanket reaches the newtype when unopposed. Delete the `ReflectStruct` blanket from the repro and `u.describe()` is `limit:3`.
- The peel is not unconditional. A direct `impl Describe for UserName { … }` does win over the `ReflectStruct` blanket. Only the bound-keyed blanket loses to it.

Declaration order is not the cause either: the `Limit` blanket is declared first above and still loses.

## Where

- `wado-compiler/src/elaborator/trait_env.rs:1587` — `value_blanket_for_receiver` filters by satisfied bounds, prefers a blanket homed in the receiver's module, then takes the first registered. Its doc comment states the assumption the selection rests on: the blankets for one trait are disjoint.
- `wado-compiler/src/elaborator/method_call.rs:2528` — the static-method path (`Deserialize::deserialize`'s shape) takes the first blanket whose bounds are satisfied.

The receiver looks to be peeled to its base before blanket selection runs, so the newtype's own impls are not in scope when the bounds are checked.

## Impact

This is what a newtype-carried validation design runs into. `impl<T: Constrained> Deserialize for T` in `core:serde` would never fire for `type UserName = String`, because `String` has its own `Deserialize`: the field would deserialize unchecked, with no diagnostic anywhere. Writing the impl directly on the newtype (`impl Deserialize for UserName`) does work today and does run at a struct's wire boundary, so the defect is confined to the bound-keyed blanket path.

## Related, and not the same defect

Two value blankets that both apply to one type resolve by declaration order, with no ambiguity diagnostic. With a struct that satisfies both bounds, `impl<T: Limit>` first yields `limit:7`; swapping the two impls yields `struct:Point`. Worth its own issue if the intent is that overlap be rejected at definition time — the variadic-impl coherence rules (WEP 2026-03-14 §5) forbid overlap for `[..T]` targets, but nothing covers two bound-keyed blankets over a bare `T`.

gfx:
A call on a newtype picks the value blanket whose bound the newtype itself
satisfies. Where only the base satisfies a bound, that blanket is still a
candidate — a newtype inherits its base's impls for dispatch — but it ranks
below one holding at the receiver.

Two things make that hold.

**A rank.** `blanket_target_bounds_depth` returns how far down the receiver's
newtype chain a blanket's target bounds first hold: 0 at the receiver, 1 at its
base. `select_trait_match` sorts on it between concrete-vs-blanket and locality.
The bound check at each level is `type_implements_trait_here`, which does not
peel — the peel is the walk, so it cannot also be the answer.

**A name.** A blanket's receiver parameter is named by the `impl` block that
binds it: `TypeHead::Binder` carries a `BinderOwner`, and its mangle renders
`T#module/17`. Two blankets of one trait therefore write two templates whatever
letter each spells its parameter; a name built from the spelling alone finds a
different block's template. Every pass that names one goes through
`Elaborator::impl_receiver_binder`, and one impl block's receiver has one name
across the decl pass, method resolution, and monomorphization.

The scope map carries the two facts together: `TraitContext::type_params` maps a
name to a `BinderInScope { index, type_id, decl }`, so a writer cannot record
that a name is in scope without recording the node that declares it.
`TraitContext::impl_owner` holds the block whose parameters are in scope, paired
with its receiver's spelling — only the receiver heads a template name, so only
that binder is owned. The index bucket a type-parameter-targeted impl registers
under is `FqTypeName::param_bucket`, keyed by spelling because several impls
share it; it is not a binder.

Where two of one trait's blankets apply and nothing ranks them, the call is
reported rather than settled by declaration order:

```
ambiguous blanket impls of 'Describe' for 'Point': 'T: Limit' and
'T: ReflectStruct' apply, and nothing ranks them; write 'impl Describe for Point'
```

At the use site, because that is where the question is decidable: whether two
bounds can both hold is not answerable in an open world, and the standard
library's four reflection blankets are mutually exclusive, so no receiver
reaches this rule.

`docs/wep-2026-09-01-trait-resolution.md` states the selection order — the ranks,
the two filters that are not ranks, the two eligibility gates, and the two
ambiguities — with `spec.md`'s coherence wording, specificity ranking, and
locality's never having been proposed recorded as known gaps. The WEPs that own
a rule link to their place in it, and `select_trait_match` cites the document
instead of restating it.

Closes #1932

---
_Generated by [Claude Code](https://claude.ai/code/session_011rtZYhvJY63v74fMeB5SXK)_


gfx:
Answered the review. Each finding was reproduced against the current code before anything was touched, and the reproduction — not the severity label — decided what happened to it. `origin/main` was built separately to say whether a defect was this branch's or older.

## Fixed

**A blanket's bound must be asked of the level being ranked.** `blanket_trait_impl_applies` checked bounds with the peeling query, so `impl<T: Base> Derived for T` answered `Wrapped: Derived` at depth 0 when only `Inner: Base`. Rank 2 then had nothing to separate, and the chained shape of #1932 came out as a *false ambiguity* where `main` silently takes the base's blanket. `NewtypePeel` now rides through the query. Fixture: `newtype_blanket_beats_chained_base_blanket`.

**Rank 3 splits local from foreign and no finer.** The tie collector compared the modules themselves, so two blankets in two foreign modules were never tied and declaration order picked one. Reproduced with a trait in one module and its two blankets in two others: `limit:7`, silently, on this branch *and* on `main`. Fixture: `error_ambiguous_value_blankets_two_foreign_modules`.

**A binder is identified by the node that binds it, not by its spelling.** With `fn show<T: Limit>` inside `impl<T: Limit> Show for T`, `p.show::<Other>()` returned 7 instead of 9 — the receiver's impl, not the argument's. This one passes on `main`, so it was a regression this branch introduced.

It appeared twice, at `binder_in_scope` and at `bind_type_param`, which is the class: *a binder's identity reconstructed by matching a spelling against a list the call site chose*. Patching the second site would have guaranteed a third, so the helper changed shape instead — `bind_type_param` takes the declaring node, every caller states which item bound the name, and `BinderInScope::from_target` (which chose the list itself) is gone. `param_decl` remains the one way to ask a list, at the sites whose list is not in doubt. Fixture: `blanket_receiver_shadowed_by_method_param`.

**The level walk needed the recursion guard, and its own answer to a cycle.** Two blankets whose bounds name each other overflowed the compiler's stack on a newtype: the `Here` arm answered without pushing the `(type, trait)` frame. Reproduced as a crash; `main` compiles the same program.

Restoring the guard was not enough. The peeling query answers a repeat `true`, which is right there — it recurses into members, so a repeat is the well-founded structural case — and wrong here, where the subject stays fixed and a repeat is a bound that grounds nothing. With `true`, unfounded `Alpha` held at depth 0 and tied with the `Limit` the newtype really carries. Flipping it to `false` then re-exposed the overflow, because the bound check's `||` fallback calls the *unguarded* lookup and only the short-circuit had been hiding that; under `Here` the guarded query is now the only one that may answer. Fixture: `blanket_bound_cycle_on_newtype`.

**A parameter bucket is neither a declaration nor a shape.** `param_bucket` built `TypeHead::Shape`, the same key `ImplTargetKey::Undeclared` builds, so both landed in one index bucket. This one is honest about its evidence: I could not construct a program where it changes an answer — an impl on an undeclared target never wins a lookup that starts from a real type — so it is fixed as a type-level separation, on the same ground as everything else in this PR, not because a test failed.

## Not fixed

**"Restore the impl binder after method-level shadowing"** (`module.rs` 509-512) does not reproduce. With `fn show<T: Limit>` declared before `fn plain(&self)` in one impl, `plain` still resolves the impl's `T` *and* its `Limit` bound, returning 7 — on this branch and on `main`. Each method re-enters its own frame from `impl_declared_params` rather than sharing a mutated map, so there is nothing to restore.

## Verification

`mise run test` on the final tree: exit 0, e2e **5060 passed / 0 failed** (5052 before these fixtures). `cargo clippy` clean. The E2E reds on `a65e33e0` were `blanket_bound_cycle_on_newtype` at every optimization level — the fixture whose expectation `89f6d09f` corrects.


---

gfx:
Both new findings reproduce against `origin/main` as well, so neither is this branch's. One is recorded as a known gap; the other is not constructible.

**Ungrounded bound cycles for a concrete receiver** — real, and pre-existing. With `impl<T: Beta> Alpha for T` beside `impl<T: Alpha> Beta for T` and a `Point` that carries only `Limit`:

- `origin/main` compiles the program, selects the `T: Alpha` blanket, and dispatches into the cycle at runtime.
- This branch reports it as an ambiguity, because the `T: Alpha` candidate is still admitted.

So the branch makes the existing acceptance louder rather than causing it. The suggested fix — apply rank 2's repeat-is-false at every level — would break the case the optimistic answer exists for: `type_implements_trait` descends into members, where a repeated pair is the well-founded structural case that makes a recursive type satisfy a derived `Eq`. One stack cannot tell that recursion from a cycle at a fixed subject, so closing this needs a second stack, and for a concrete receiver it is a change to *dispatch*, not to selection — outside a PR about selection order. Recorded as a known gap in the WEP with the mechanism and what closing it takes (75e7398).

**A receiver declared in a nonzero slot** — not constructible. `impl<A, T: Limit> Make for T` is rejected on both this branch and `main`:

```
error: the type parameter `A` is not constrained by the impl target or the trait reference
```

A value blanket's receiver is slot 0 by construction, enforced by that coherence check, so the constant in `blanket_param_slot` is the invariant rather than an assumption. It is also byte-identical on `origin/main` — this PR replaced the tuple its caller returned with a named struct and changed no lookup. No fixture added: a test for a program the compiler already rejects at the header would assert the existing check, not this claim.


---

PR #1938, opened 2026-09-01
docs: a rubric for surveying emerging languages, two filled surveys, and what Wado took from them

gfx:
## What this adds

A fixed set of questions to ask about a new language, so that surveying the next
one reuses the last one's shape instead of starting from prose. Two surveys
filled in against it. And the changes to Wado's own documents that those surveys
argued for.

### The rubric — `docs/research-language-survey.md`

Three groups of questions, each about a different subject: **A. The language**
(what you can write, and what happens when you do), **B. The practice** (how the
project decides, and what keeps it honest), **C. What outlives it** (what would
still be worth having if the project stopped tomorrow).

Five rules govern how a row gets filled:

- Only ask what can be measured. Every question names what to run or what file
  to open.
- Keep the claim and the reality in separate columns. The gap between them is
  usually the finding.
- "Not built yet" is not "decided against". The first changes with time; the
  second is the design.
- A survey is a snapshot of one commit on one day, rewritten whole rather than
  maintained. There is no document comparing the surveys, because a comparison
  goes stale at the combined speed of every project in it.
- Say how you looked. An absence found by searching carries no evidence unless
  it says where the search went.

The strongest question is the cross-check: take the hardest program written in
the language and ask whether the claim still holds there. A standard library is
the floor; a compiler written in the language is the ceiling. Alongside it, count
how many implementations of the semantics exist, because whatever holds two of
them to one answer is a major finding rather than a detail.

The rubric closes by stating what it cannot see. It is document-driven, only the
cross-check can contradict a document, and it carries no scores — a 1-to-5 scale
across languages would price a decision made under one arbiter using another's.

### The surveys

`research-language-survey-almide.md` and `research-language-survey-vibe.md`, each
with a header naming what was examined and what was left alone, and each closing
with what Wado learned, took, refused, and is holding.

### Changes to Wado's documents

`docs/spec.md` opens with a `## Status` section. The specification is normative:
it says what the language is meant to be, and a program's meaning is read from
there. A disagreement with the implementation is a defect belonging to whichever
side is wrong, settled when it is found, and never left standing as an accepted
difference. A table places the WEPs, the philosophy, the cheatsheet and the
generated stdlib pages relative to it.

`docs/spec.md` also states that copies are as-if: a program may rely on the value
each expression denotes, not on the number of copies performed. And colorless
async moves from a philosophy bullet into the concurrency section, where a
signature not recording whether a function suspends is a rule rather than an
aside about a keyword.

`docs/design-philosophy.md` is rebuilt around one root. Wado is in full control
of the wasm it emits; small binaries and the absence of unpredictability both
follow from that rather than being pursued separately. Overloading, coercion and
derivation all exist and each is asked to be predictable, with the boundary
stated: a distant file may at most make a call legal that was not, never change
what a legal call meant. Speed joins the principles, as a requirement rather
than a bonus, covering both runtime and compile time. The closing section says
that the language has no features for AI, because complexity is where a model
fails and that is where people fail too.

`docs/wep-2026-08-31-wado-lint.md` proposes `wado lint`, a container for
corpus-level checks. Its first check reports one function that reimplements
another, comparing canonicalized TIR bodies and unifying a hand-written function
against the generic it instantiates. The forwarder rule, the minimum body size
and the near-duplicate threshold are Known gaps, each with the measurement that
closes it.

`.claude/commands/distill.md` becomes `.claude/skills/distill/SKILL.md`, and its
Markdown rules now lead with the goal: prose a reader understands on the first
pass. Cutting redundancy is one way to get there and not the point of it.


gfx:
## Review response

Six of seven findings were real. Details below, skips included.

### Fixed

**`research-language-survey-almide.md:40` — `Codec` is a boundary mechanism.**
Correct, and it was self-contradictory inside one cell: the row said no
mechanism a user can invoke, then described `Codec`, which a user invokes by
writing `: Codec`. The conclusion built on it was worse — "no generation
mechanism and no serialization framework" — where only the first half holds.

Corrected, the argument gets sharper rather than weaker. A wire format does have
a route into an Almide program. A foreign grammar or IDL does not, because
nothing turns one into source at build time, and a Gale starts by consuming a
grammar file.

**`research-language-survey-vibe.md:143` — ADR-0076 has not landed.** Correct.
The ADR puts itself at phase 0, design only, no code changes, and the survey used
the present tense. This is the "not built yet" against "decided against"
distinction the rubric asks every surveyed project about, missed in the rubric's
own survey.

**`research-language-survey.md:96-97` — the two counts used different file sets.**
Correct, and this was the rubric's only concrete procedure. `grep -r` recurses
and a bare `ls` does not, so a stdlib with subdirectories yields a numerator
larger than its own denominator. Both counts now come off one `find` list.

**`research-language-survey.md:19` — the rows do not name evidence.** The claim
is right; the suggested fix is not. An evidence column cannot name a path,
because the path differs in every project. So the rule was the thing overstating,
not the rows. It now says a question has to point at something specific you can
go and look at, and defers what you actually opened to the existing "Say how you
looked" rule.

**`research-language-survey-vibe.md:139` — the perform ceiling.** 128 KiB over 8
bytes is 16,384, and the figure quoted is 16,382. It is theirs, not a calculation
of ours, and it is now attributed as such with the two-slot discrepancy noted
rather than smoothed over.

**`wep-2026-08-31-wado-lint.md:98` — the corpus boundary heading.** The heading
could be read as "only the stdlib is matched against", which the dependency flag
contradicts. Renamed to what the section actually says — the stdlib never
receives a finding — with dependencies stated to work the same way.

### Skipped

**Both survey headers against their claims.** Checked and not reproducible. No
claim in either survey rests on material its header lists as unexamined. The
vibe A5 row is sourced from `README.md`, and the header's note that the ADRs
behind A5 went unread is that rule working rather than being broken. The Almide
survey cites `docs/specs/edit-locality.md`, which is in the examined list; the
1,607-line `SPEC.md` it lists as unread is cited nowhere.

### Also

The correction to the Almide row turned out to be a gap in the rubric, so it
gained a clause: which kind of boundary mechanism a project has decides which
programs it can produce. Serialization admits a wire format at run time, which is
enough for an application. Generation admits a grammar at build time, which is
what a parser generator needs. Having the first does not imply being able to
build the second.

PR #1939, opened 2026-09-01
perf(optimizer): cut the derived struct deserializer's quadratic field dispatch

gfx:
A `ReflectStruct`-derived `Deserialize` routes a field to its slot with
`for let [i, _] of slots.enumerate() { if i == index { … } }`. The variadic `for`
is unrolled at compile time and its arms are plain siblings, so nothing leaves
the chain once one fires: a struct pays one comparison per declared field for
every field on the wire.

A sweep over struct width at constant total work shows it — 2,000,000 fields
decoded from ~270 KB of CBOR either way:

| fields/record |  5  | 10  | 20  | 40  | 80  |
| ------------- | --- | --- | --- | --- | --- |
| ms            | 147 | 154 | 173 | 262 | 486 |

## `nir/if_chain_to_match`

Fuses a run of sibling `if K == x { … }` statements over one local into a single
`Match`. Distinct constants over a local no arm writes are what make the guards
exclusive; the constant bindings the unroll leaves between the arms move ahead of
the run. It has no width threshold of its own — the `Match` lowers to an
early-exit `else if`, always fewer comparisons than the flat run.

`match_to_switch` holds the one threshold, at twelve arms: a `br_table` trades a
cascade the predictor gets right for one indirect branch, and only pays once the
cascade is long. At nine arms json-catalog's derived deserializers lose 3.6% on
deserialize against the `else if` chain the same `Match` lowers to otherwise;
cbor-twitter's 12-to-40-arm structs prefer the table.

## UTF-8 validation

`String::from_utf8_slice` and `is_valid_utf8` share one scan whose ASCII test is
written out rather than left to `utf8_sequence_len`, which the inliner declines
on size — a call a byte otherwise.

## Results

Deserialize throughput, best of three alternating pairs on an idle host:

| row             | before      | after       |        |
| --------------- | ----------- | ----------- | ------ |
| cbor-twitter de | 212.33 MB/s | 235.11 MB/s | +10.7% |
| json-twitter de | 163.43 MB/s | 173.18 MB/s | +6.0%  |
| cbor-catalog de | 650.81 MB/s | 664.36 MB/s | +2.1%  |

Every other row is flat, serialize included, and so are `sqlite-parse`,
`syntax-highlight` and `gale-gen`, where the switch threshold retires gale's
keyword-set tables.

## Also here

`Engine::ensure_value_graph`'s `VG_MAX_EXPRS` gate is gone. It skipped the graph
for a body over 5000 expressions to save memory it does not save: `mise run
test-wado` over the same 1781 files peaks at 10.55 GB without it against 10.68
with, for 7755 s of compile CPU against 7870.

`arena_query::local_written_by` is the one definition of the write channels,
asked by both a whole-body scan and a subtree one, which previously enumerated
different sets and each missed what the other had.

Five fixtures named for the `br_table` they compile now assert it. A threshold
is exactly what silently drops such a fixture off the path it exists to guard
while it stays green.

`scripts/exclusive.sh` refuses a second `test` / `test-wado` / `test-gale-o3`
while one holds the lock, naming the holder's pid, and says on stderr when it
runs unlocked for want of `flock`. Two core-saturating runs do not fail, they
starve each other, and the newcomer's silence reads as a hang.


gfx:
Thanks — two of the three were my own text contradicting itself, now fixed in cb1c146.

**`if_chain_to_match.rs` / `docs/optimizer.md` — "always fewer comparisons"**: correct, that is false. An unmatched value and a last-arm value both test every key either way. Reworded to "never tests more keys than the flat run and stops at the arm that fires", which is what actually justifies fusing without a width threshold.

**`SKILL.md` — profiler attribution**: also correct, the headline named function entries while the next sentence said entries *and* loop headers. Now "A sample lands at the next epoch check, not where the time went."

**`scripts/exclusive.sh` — fail closed without `flock`**: not taking this one, and here is why. The lock guards against a mistake — starting a second core-saturating run and having the two starve each other — it is not a correctness requirement; nothing is wrong with the tree if two runs overlap, they are just both slow. macOS ships no `flock`, and `ci.yml`, which is where `test` / `test-wado` / `test-gale-o3` run, is Linux-only, so failing closed would cost macOS developers the ability to run the suite at all in exchange for enforcing a convenience CI never exercises. It now says on stderr when it runs unlocked, so the gap is visible rather than silent.

PR #1949, opened 2026-09-02
feat(elaborator): trait resolution as functions of a self-contained program

### LINKED ISSUE #1711: Bare `Ok` / `Err` / `Some` / `None` constructor calls ICE in `wir_build` instead of erroring
## Summary

Calling one of the four built-in constructor names without its type prefix — `Ok(1)` rather than `Result::Ok(1)` — passes the elaborator with no diagnostic and then panics in `wir_build`. An undefined name gets a proper error; these four do not.

```wado
export fn run() {
    let x = Ok(1);
}
```

```
thread 'main' panicked at wado-compiler/src/wir_build/translate.rs:1881:13:
[WIR] unresolved Call: name="Ok" builtin=None
```

`Some(1)` and `Err(1)` panic identically. The control case is clean:

```wado
export fn run() {
    let x = nosuchfn(1);
}
```

```
error: unknown function 'nosuchfn'
```

With an expected type in scope, the diagnostic that does appear points at the wrong thing — the constructor is reported as producing unit:

```wado
export fn run() {
    let r: Result&lt;String, String&gt; = Ok("x");
}
```

```
error: type mismatch: expected 'Result&lt;String, String&gt;', found '()'
```

## Cause

`resolve_call` (`wado-compiler/src/elaborator/call.rs`, the branch commented "a built-in type constructor (Ok, Err, Some, None)") matches these four names through the `CompilerItem` registry and resolves them to a local `CalleeRef`. That is what keeps `UnknownFunction` from being emitted. Nothing downstream builds a variant construction for them, so `wir_build` looks up a function that does not exist and panics.

## Which way to fix it

`docs/spec.md` and `docs/cheatsheet.md` construct these with the type prefix (`Result::Ok(x)`, `Result::&lt;i32, String&gt;::Ok(42)`), and the stdlib only ever uses the bare names in *patterns*. So the two coherent outcomes are:

- Bare constructor calls are not part of the language → that branch should emit `UnknownFunction` like any other unresolved name, and the ICE becomes an ordinary error.
- Bare constructor calls are intended → the construction has to be built, and the expected type has to drive the payload/variant inference (the unit type in the third example is the symptom of it not being).

## Reproduction

`wado build --lib -O0` on any of the sources above; the world does not matter.

Found while writing the `task.return` reclamation test for #1708 — `task return Ok(s)` in an `export async fn` hit this before reaching the CM binding.

gfx:
## Summary

`docs/wep-2026-09-01-trait-resolution.md` states the selection order, and `wado-compiler/src/trait_solver/` is where it is checkable: `coherence_errors`, `derive`, `holds` and `rank` are functions of a self-contained `Program`, under unit tests with no Wado source. `elaborator::solver_bridge` lowers the compiler's tables into that program.

## The solver

- `Program` carries impls, traits (supertraits, `holds_for_all`, argument defaults, a reference rule), newtype bases, reflection facts gated by asking module, and associated-type bindings. Every id is a plain index.
- `holds` answers a bound by a bound in force, a fact, a trait holding for all, or an impl whose own bounds hold in turn, with the newtype chain and auto-deref as the two inherited steps. A repeated `(type, trait)` goal grounds nothing. A bound asks for the trait at its declared defaults and may pin an associated type.
- `derive` models derivation as impl generation: a declaration whose members all satisfy a structural trait contributes `impl<Pi: Tr, …> Tr for D<P1..Pn>`, decided as an assume-then-refute fixpoint over the declarations. Variants never derive `Ord`.
- `coherence_errors` rejects two impls of one `(Trait, Type)` pair and an unbounded value blanket. Both are compiler diagnostics, reported where the impl is written.
- `rank` applies ranks 0-3 and names the two ambiguities.

## The differential

In debug builds the bridge is built once every declaration is resolved, and `type_implements_trait` asserts the solver's answer against its own at every outermost query. The fixture corpus is the drift detector; `explain` renders the lowered question, environment, and the trait's impls when they disagree.

## Compiler changes the differential drove

- A generic struct's field is read at the instance (`items: List<T>` is `List<i32>` at `Gen<i32>`), however deep the parameter sits.
- The recursion guard counts member descents: a repeated question with one between the askings is a recursive type and holds, one reached through bounds alone does not.
- `()` is the unit type, distinct from the empty tuple `[]`. The prelude implements `Eq` and `Ord` for it directly.
- A case may be written without its type name only where the expected type supplies it: an annotation, a parameter, a return type, a payload. `Some(())`, `None`, `Leaf`, `Red` construct there by one rule, with no prelude special-casing, and `let red = Red;` is an error naming the qualification to write.

## Docs and fixtures

The WEP records the order, the solver's input, derivation as impl generation, and every remaining gap as a Known gap. New `trait_*` fixtures pin each rule. A fixture for a behaviour the compiler does not have yet carries `#[TODO]` with its expectation in place, so the harness reports it as pending and fails hard once the gap closes.

Closes #1711

https://claude.ai/code/session_01XWHpELinHCpgZmbw7BvbXt


gfx:
Review round 1, answered in a9f4be469 and f8e621fb4.

Fixed:

- `holds.rs`: a derived or marker impl's request names the impl's trait. `impl Sub for Point` answering `Point: Base` owes the `Sub` body. Unit test `a_request_names_the_answering_impl_s_trait`.
- `coherence.rs`: a parameter's bounds are compared as a set, so `impl<T: A + B>` and `impl<T: B + A>` are one pair. Unit test `two_blankets_at_reordered_bounds_are_a_duplicate`.
- `elaborator.rs`: a bare ident has no segments, so the `[]` arm now records the use-to-definition edge. `wado query definition` and `references` answer for `Leaf` and `None`; before the fix both returned nothing.
- The two coherence diagnostics quote names with single quotes, as the other trait diagnostics do. Fixture expectations updated.
- `trait_error_cross_trait_blankets.wado`: the header leads with the rule the `__DATA__` pins and labels the current behaviour "Today". `trait_error_local_blanket_ties_foreign.wado` already labels its paragraph that way and is unchanged.
- WEP 2026-07-31: "in scope" is stated as the calling module's scope.

Skipped:

- The three scope-gap fixtures (`trait_error_unimported_*`) stay as red fixtures with `compile_error`. That is the corpus convention on this branch: a desired-missing diagnostic is a red fixture, and `#[TODO]` is for behaviour not yet decided. They are the five intended failures in the E2E jobs (with `trait_error_cross_trait_blankets` and `trait_error_local_blanket_ties_foreign`), all listed under Known gaps in the WEP.

CI: "Test (Wado)" was a real disagreement, fixed in f8e621fb4. The solver refused a blanket whose pack is read off a projection (`impl<S: ReflectStruct<Fields = [..F]>, ..F: Arbitrary>`), where the compiler defers that bound to monomorphization. The solver now defers it too, and leaves the `Reflect*`-bounded blankets of the structural traits out of the program, since `derive` answers for those per declaration as the member walk does. `mise run test-wado` with CI's exclusion compiles all 194 files, and the e2e suite is back to the ten intended failures.

---

gfx:
Review round 2, answered in d95540c53 and e9a007cd1.

Fixed:

- `coherence.rs`: the impl key is canonical. Omitted trait arguments are expanded at the trait's defaults with `Self` read as the target, so `impl Mul for Point` and `impl Mul<Point> for Point` are one pair. Each parameter's bounds and pins are sorted, so `T: A<X = i32> + B` and `T: B + A<X = i32>` are one condition. A default the lowering cannot spell keeps the omission, so an impl writing that argument stays a distinct pair rather than a guessed duplicate. Unit tests `two_blankets_at_reordered_pinned_bounds_are_a_duplicate` and `an_omitted_argument_is_its_default`.
- `holds.rs`: a bound on a parameter that neither the target nor a pin supplies now refuses the impl. Only a pin-supplied parameter (`Fields = [..F]`) defers its bound to monomorphization. `impl<T, U: Alpha> Beta for List<T>` answers for no receiver, which agrees with the compiler rejecting that impl where it is written ("is not constrained by the impl target or the trait reference"). Unit test `a_bound_on_a_parameter_nothing_supplies_answers_for_no_one`.

Verification: 79 solver unit tests, clippy clean, `test-wado` with CI's exclusion at 195 files and 2215 tests green, and the e2e suite at 5108 passed with the ten intended failures.

---

gfx:
Update on the five red fixtures: as of a7471cac0 they carry `#[TODO]` with their `compile_error` expectations in place. The harness reports each as pending, and fails hard once the diagnostic appears, so the pins survive and the E2E jobs should now be green.

---

PR #1950, opened 2026-09-02
fix(lsp,gale): a contextual keyword is whichever the parse read it as

### LINKED ISSUE #1944: Contextual keywords used as names are coloured as keywords, in both highlighters
`Wado.g4`'s `identifier` rule accepts 20 contextual keywords as names. Both
highlighters — the Gale-generated one and the compiler's own classifier — colour
several of them as keywords instead.

The compiler accepts 9 of the 20 in `let` position; 4 of those are wrong today:

| written | Gale | compiler's classifier | should be |
| --- | --- | --- | --- |
| `let type = 1` | keyword | keyword | variable |
| `let from = 1` | keyword | keyword | variable |
| `let of = 1` | keyword | keyword | variable |
| `let flags = 1` | keyword | keyword | variable |
| `let test/do/task/trap/forward = 1` | variable | variable | variable |

`let matches = 1` is a fifth shape: Gale colours it `operator`.

The two sides agree, wrongly, which is why `mise run check-highlight` has never
reported it.

## Gale side

`Wado.highlights.scm` lists 5 of the 20 by hand. # adds
`(rule) @capture`, which expands at generation time to one override per token
the rule accepts, so the fix is one line:

```scm
(identifier) @variable
```

Applying it drops the corpus divergence count from 129,788 to 17,391 — most of
what the query left uncoloured were names it can now place. It cannot be landed
alone, because `check-highlight` then reports the compiler's side as a
disagreement (222x `compiler=keyword gale=ident`, e.g.
`wado-compiler/lib/core/base64_test.wado:75` `flags`).

## Compiler side (`wado-lsp/src/semantic_tokens.rs`)

`classify_token` matches `token.kind.keyword_category()` first, so a keyword
*token* the parser accepted as a name never reaches the identifier path. The
mirror case already works: the contextual keywords that lex as *identifiers*
(`test`, `do`, `resume`, `task`, `trap`, `forward`) are recognised by position
through `AstSpans::contextual`.

Two attempts, both backed out:

1. **Consult `sem_classes` before `classify_keyword`.** No effect — a local
   binding is in neither `iter_symbols` nor `iter_references`, so the map has no
   entry at those spans.
2. **Record `Pattern::Ident` / `Pattern::MutIdent` spans as binding positions.**
   Declaration sites fixed (222 → 205), but `Pattern::MutIdent`'s span starts at
   `mut`, so 19 `mut` tokens were misclassified as variables; and *use* sites
   (`encode_with(&data, flags)`) still classify as keyword.

So it needs:

- [ ] A name span for `Pattern::MutIdent` — either a `name_span` field on the
      AST node, or deriving it from the token after `mut`.
- [ ] Use sites of a keyword-named local. They carry no use→def edge, so either
      the binding-span set has to cover them or they need to go through
      `classify_ident`.
- [ ] Then `(identifier) @variable` in `Wado.highlights.scm`, with
      `check-highlight` green.
- [ ] The same review for SQLite: `SQLite.g4`'s `any_name : IDENTIFIER | keyword
      | …` has the identical shape, and `SQLite.highlights.scm` carries no
      rule-context capture at all, so `SELECT key FROM t` colours `key` as a
      keyword. Pin it with a test.

## Corpus

`wado-compiler/tests/fixtures/contextual_keyword_names.wado` is already in, and
holds every contextual keyword the compiler accepts as a name. The corpus had no
such case before — nothing in the stdlib names a variable `type` — which is why
the class went unnoticed. It is inert while both sides agree and fails as soon
as one is fixed without the other.

## Note on scope

`memberName` looks like the same gap but may not be: it accepts ~35 keywords and
the query overrides 6, so `x.type` colours as a keyword. The existing comment in
the query claims a real keyword stays a keyword on both sides there, and the
compiler agrees today. Whether `x.match` should be `@property` is a separate
editorial decision, not obviously a bug.

gfx:
A contextual keyword is a keyword or a name depending on where it sits, and
only the parse tells the two apart: `type Alias = i32;` and `let type = 1;`
put the same token in the same position of the same shape, and a name binds no
AST node a walk could find it by. So the parse records its reading.

`Module::contextual_keywords` holds the span of every token read against its
lexical class, in source order:

- an identifier read as a keyword — `test "…" { }`, `resume`, `do`,
  `task return`, `..trap`, `..forward`;
- a keyword read as a name — `let type = 1`, `fn from(…)`, `Instant::from(x)`,
  `struct { variant: String }`, and every keyword after a `.` (`x.match`).

`Parser::restore` rolls the record back with the speculative branch that wrote
it, and a template interpolation's sub-parser appends into the enclosing
module, whose file its spans already index.

`wado_lsp::semantic_tokens::classify_token` consults that set before the
keyword category, so a keyword token read as a name takes the identifier path
(`let type = 1` is a variable, `fn from(…)` a function, `x.type` a property)
and an identifier token read as a keyword colours as one. Every other token
keeps the role its lexing gives it, which is what makes a reading the parser
does not record degrade to the token's own class rather than to a name.

`Wado.highlights.scm` captures each word its `identifier` rule accepts as a
name, and gives `methodName` / `fieldName` a whole-rule capture: a member name
is a name whichever of `memberName`'s ~45 words it is. `(identifier) @variable`
cannot stand in for the list, because the IDENTIFIER token it also claims
outranks `typeRef` / `pathSegment` from further in and paints every type name
and enum case a variable. `check-highlight` compares the two highlighters over
the stdlib + fixture corpus and reports no disagreement.

Gale's `(rule) @capture` follows a rule that delegates to a naming rule, which
is what lets `fieldName : memberName` read its token set off the grammar.
`gen_highlight` emits the CST walk for every grammar with a parser: the walk
carries the rule stack an override reads, and a query with no override reaches
the same captures without one.

`SQLite.g4` has the same shape in `any_name : IDENTIFIER | keyword | …`, so its
query captures `(keyword) @variable`: `SELECT key FROM t` names a column and
`PRIMARY KEY` does not.

`Wado.g4`'s name rules accept what the parser accepts — `extends` under
`identifier`, `extends` and `internal` under `memberName` — and
`check-highlight-vocab` holds them there: every keyword in `NAME_KEYWORDS` is
a name in `identifier`, and every keyword is one in `memberName`. The corpus
cannot report that class of gap, since a file the grammar rejects is dropped
from the comparison rather than compared.

Closes #1944

https://claude.ai/code/session_014fVjoLnWWqcPT7UQsDKzMR

---
_Generated by [Claude Code](https://claude.ai/code/session_014fVjoLnWWqcPT7UQsDKzMR)_


gfx:
Both findings from the last review, answered in `eb992ab9e`.

**Fixed — `memberName` excludes `resume` and `self`.** Real, and of the same class as the `extends` / `internal` gap this branch already closed. Both words lex as identifiers for the compiler, so `consume_field_name` takes either as a member name; `wado check` accepts `struct S { resume: i32, self: i32 }` with `s.resume + s.self`, which the grammar rejected. `memberName` spells both now.

The half that let them through was the vocabulary check itself: invariant 5 held `memberName` to `KEYWORDS` alone, and these two live in `CONTEXTUAL_KEYWORDS`. It now covers both registries — removing either word from the rule reports it by name.

`self` keeps the class the compiler reads it as wherever it stands, so `(fieldName "self")` and `(methodName "self")` capture `@constant.builtin` ahead of the whole-rule captures. A `::` segment gained the identifier-lexed contextual keywords for the same reason: `Foo::test` is a name to the compiler and was a keyword to the grammar.

**Skipped — capture every `memberName` literal under `pathSegment`.** `Foo::internal` is not a program the compiler reads: a path segment goes through `consume_ident`, whose `as_ident_name` takes an identifier or one of `flags` / `type` / `of` / `from` / `extends`, and `internal` is none of those. Capturing it would colour a shape that has no compiler reading to agree with. The segment list is the set the compiler accepts there, which is what the comment above it now says.

`check-grammar` (2973 files, 0 gaps) and `check-highlight` (0 disagreements) are green on the merged tree.

---

PR #1952, opened 2026-09-03
feat(core): add clamp and String::split_once, read the stdlib from disk in dev

gfx:
## `clamp`

Every primitive carries `clamp` alongside `min` and `max`, i128 and u128 included. The wide pair compares with `if` rather than `builtin::select`, which would evaluate both sides of a GC struct. An empty range is an assertion, so `clamp(v, 10, 1)` traps rather than silently picking a bound; a NaN bound trips the same assertion. A NaN `x` passes through, since wasm `f64.min` / `f64.max` propagate it.

`core:temporal`'s date `constrain` and fraction-digit resolution clamp through it, as do `Slice::slice`, `Slice::to_array` and the other prelude sites that bound an index.

## `String::split_once` / `rsplit_once`

Both return the parts either side of the first (or last) separator as `Option<[String, String]>` — the shape a key/value or path/extension split wants, where `splitn(2, sep)` yields an iterator the caller must unpack and re-check. An empty separator splits at the near end for `split_once` and the far end for `rsplit_once`, matching `find` / `rfind`.

`core:url` reads through them: the `key=value` of a query part, the `@` and `:` of an authority, the `:` before a port, and the `#` / `?` that peel a fragment and query off a reference.

## A dev build reads the stdlib from disk

`src/stdlib.rs` declares both build modes from one list of `(import path, file)` pairs, so they cannot drift. A dev build reads `lib/` at first use, so editing a stdlib module takes effect on the next `wado` run: touching `lib/core/prelude/string.wado` and running `cargo build -p wado-compiler` finishes in 0.13s with nothing recompiled. A release build embeds the sources with `include_str!`, as does any `wasm32` build, which has no filesystem. `CORE_LIBM_WAT` is embedded in every mode — it is generated bytes, not source anyone edits.

`all_core_modules()` and `all_binding_modules()` are the two tables, and `get_stdlib_module` is how a caller reaches one module's source.

## Verification

`cargo test`, `wado test` (1279 files, 5159 tests), `mise run clippy` with `-D warnings` across the workspace, and `check-grammar` all pass.

https://claude.ai/code/session_01TkXjZLtdn5AXASiehcNczY

---
_Generated by [Claude Code](https://claude.ai/code/session_01TkXjZLtdn5AXASiehcNczY)_


gfx:
@coderabbitai resume

PR #1953, opened 2026-09-03
perf(zlib): rewrite core:zlib's hot paths, and stop field_scalarize over-syncing nested loops

gfx:
`core:zlib` compresses `twitter.json` at 117 MB/s and decompresses at 509 MB/s,
and `nir/field_scalarize` stops emitting sync it can prove nothing observes.

## `core:zlib`

Deflate walks its symbol buffer once. `count_block_freqs` builds the frequency
table, `encoded_data_bits` prices both the fixed and the dynamic encoding by
summing over codes, and `plan_dynamic_block` builds the trees, the RLE'd
code-length header and the block's exact bit cost into the `DynamicPlan` that
`write_dynamic_block` then emits.

Three tables are one `i32` per entry rather than an array of GC references, so a
lookup is one unboxed load:

- the inflate decoding table (`op`, `bits`, `val`),
- the deflate encoding trees (bit-reversed code, length),
- the LZ77 token buffer (symbol, match distance), which also halves the buffer
  and the per-token push.

The LZ77 chain search rejects a candidate on two bytes and starts from the
pending lazy match, as zlib's `longest_match` does. `hash3` is rolling, so the
positions a match skips cost one load each, and `insert_hash_run` buckets them
in one place for all three emit paths: it starts after the position the search
already bucketed, so `prev` holds no self-link and a chain walk reaches every
older candidate rather than spending its `max_chain` budget on one position.

An inflate literal is written where it is decoded, without a mode of its own.
`deflate_stored` emits a BFINAL block for empty input, which RFC 1951 requires
and which makes `deflate_raw(&[], 0)` a stream that inflates.

| Compress               |  Throughput |  ms/iter | vs best |
| ---------------------- | ----------: | -------: | ------- |
| Rust (zlib-rs)         | 327.71 MB/s | 1.927 ms | 1.00x   |
| JavaScript (node:zlib) | 205.94 MB/s | 3.067 ms | 1.59x   |
| C (zlib 1.3.1, Wasm)   | 126.41 MB/s | 4.996 ms | 2.59x   |
| **Wado** (core:zlib)   | 116.98 MB/s | 5.398 ms | 2.80x   |

| Decompress             |  Throughput |  ms/iter | vs best |
| ---------------------- | ----------: | -------: | ------- |
| Rust (zlib-rs)         |   3.24 GB/s | 0.195 ms | 1.00x   |
| JavaScript (node:zlib) |   1.79 GB/s | 0.354 ms | 1.82x   |
| C (zlib 1.3.1, Wasm)   | 840.05 MB/s | 0.752 ms | 3.86x   |
| **Wado** (core:zlib)   | 509.36 MB/s | 1.239 ms | 6.35x   |

`twitter.json` deflates to 51458 bytes, against C zlib's 51494 at the same
level.

## `nir/field_scalarize`

A nested loop is inside the scalarization scope rather than an exit from it.
Only a call can observe a candidate through the struct there — a direct field
access is what the pass rewrites to the scalar, and a local or field whose
address escapes is refused at selection — so `walk_nested_loop` commits only
what a call in the body reaches, and re-reads only that after.

An unlabeled `break` out of a nested loop lands back in the loop body, so it is
recorded and joined with the loop's other exits, and only the sites that diverge
sync. The post-loop `ScalarOnly` demotion to `FieldOnly` covers the
zero-iteration path, which has no slot for convergence sync, so it applies
unless every exit already agrees on `ScalarOnly`.

A candidate is weakened to `ScalarOnly` at a nested loop only when the body
writes it, which is what makes the back edge free. One the body never writes
cannot leave `Both`, so weakening it would cost the next call that wants the
field a write-back and buy nothing.

`core:zlib`'s inflate loop is where this shows: a bit-buffer refill loop reaches
nothing and syncs nothing, worth 6% of decompress. `gale-gen` pays 0.75% for it,
its scalars living across nested loops in a function wide enough to spend that
on spills. Compiling every benchmark under both compilers leaves byte-identical
wasm on all but `zlib`, `json-catalog-v2` and `gale-gen`, so no other row is
affected.


gfx:
Both findings verified against the current code before anything was touched.

## Fixed: the hash chain self-link

Real, and larger than the comment claims. `pos` is bucketed at the end of the
match search, so the skip-insert run starting at `pos` writes `prev[pos] = pos`.
A later search reaching `pos` in that bucket spends its whole `max_chain` budget
on it. The two sibling runs already start at `pos + 1`; only this one had
drifted, and the old bound also dropped the position at the far end.

The proposed range is right. twitter.json now compresses to 51458 bytes instead
of 53096, at 117 MB/s instead of 96 — the chain cost ratio *and* speed. That is
a slightly better ratio than C zlib's 51494, and compress is now within 8% of it
on wasmtime.

Two synthetic inputs failed to expose this, because their matches sit at the
head of the bucket. The third does: text-shaped data whose small vocabulary
recurs at varying distances holds 16.2 KB with the walk running to the end of a
bucket and 17.9 KB when it stops at the head. `a match deep in the hash chain is
still found` bounds it at 17 KB, and fails at exactly 17881 when the fix is
reverted.

Since one of three identical runs drifting is what produced the defect, the
three are now one `insert_hash_run`, so the rule lives in one place.

## Fixed, in part: the sweep snippet

The world argument is a fair point and is now in the snippet, along with
dropping the schema modules from the glob. Every remaining benchmark must
compile, so a failure is reported as `FAILED` rather than skipped. Verified over
the corpus: no failures, and `http_routing` is now compared rather than passed
over.

Not taken: reporting compile diagnostics to classify failures. Once the schema
modules are out of the glob there is no expected failure left to distinguish
from an unexpected one, so parsing diagnostics would add machinery for a
distinction that no longer exists.

Not accurate in the finding: that a change here "can escape comparison". The
skipped rows were printed, and the prose already said a skipped row is uncovered
rather than unchanged. The snippet now removes the manual follow-up that
required, which is the part worth having.

gfx:
Fixed, and the finding was right about the shape even though its specific
failure mode does not occur here.

I checked first: `wado compile` exits 1 and writes no file when it fails, so
"nonzero status while producing an output" was not reachable. But file existence
was standing in for "the compile succeeded", and the exit status is that signal
directly. Using it makes the loop shorter, not longer — the separate emptiness
test and the per-round delete both go, because a failed compile now leaves
before any comparison.

On diagnostics: `wado compile` reports on stderr even on success ("Generated:
…"), so simply letting stderr through buried the sweep's own output under two
lines per benchmark. They are held back and printed with the `FAILED` row
instead, which is where they are wanted. Verified both paths — the corpus sweeps
clean with only the three `DIFFERS` lines, and a forced failure prints the
compiler's error above its `FAILED` row.

The SkillSpector "Tool Parameter Abuse" annotation on the same lines is not
about this code; it fires on the shell block generically.

gfx:
Skipping this one. The fact is right and the state it guards is unreachable.

`cmp` does exit 2 on a missing operand — confirmed locally. But the compare only
runs after both compiles have exited 0, which is the gate the previous round
added, and an exit-0 `wado compile` leaves a readable non-empty file: checked
across all sixteen benchmarks, no exit-0 compile failed to produce one.

The direction also matters. A `cmp` error would land in the `DIFFERS` bucket,
which sends you to measure that row back to back — where a missing file is
immediately visible. The answer this check must never give by accident is
"identical", and `cmp` returns 0 only when both files are readable and equal, so
no error status can produce it.

So the branch would add handling for a state that cannot arise, whose
misclassification is in the harmless direction, in five lines whose job is to
teach the sweep rather than to be exhaustive shell. The two earlier findings on
these lines were worth taking because each made the sweep cover more of the
corpus or made it shorter; this one does neither.

PR #1965, opened 2026-09-04
perf(optimizer): price a splice by what it replaces

gfx:
The optimizer prices two rewrites by what the caller actually receives, and
const-object globalization sees through a borrow the inliner leaves behind.

## Inliner: the price of a splice is net of the call site

`inline_threshold` bounds how much a caller grows, so `net_cost` is the quantity
it judges: the callee's body less the `call_site_size(param_count)` that the
splice replaces. `cold_outline` already makes the same trade in the other
direction, weighing a region against the call that would replace it.

## Inliner: one level of splice lookahead

`weight::CALL` is what a call that *stays* a call costs. A driver — a callee
built out of other inline candidates — reads as a handful of ABI edges, while
what its caller receives is every one of those bodies. `inline_functions` builds
a `spliced` table indexed by `FuncId`, holding what each loop-carrying callee
costs to splice, and `CostWalk::call_price` charges a call by that instead of
the edge. The table reads `splice_barred` and `effective_threshold`, so it
agrees with `classify_callee` on which callees this pass leaves alone.

A site handing the callee any constant is charged the edge regardless, because
the constant may unroll the loop the table priced.

Only a loop-carrying callee is priced this way. A loop is what the model charges
most to splice, being a region of its own in the caller; pricing every inlinable
callee by its whole body judges a driver by its post-inlining size against a
threshold calibrated on as-written ones.

## WIR: flatten a `?`-unwrapped variant slot

`slot_flatten` recognises a `then` arm that is a pure slot copy through a
single-element `Seq`, so the `?`-unwrap shape reaches the rewrite and every
sequence element is unboxed. `drop_tail_result` clears the block result type
along the diverging arm's tail chain, which `rewrite_unwrap_to_guard` leaves
declared once the outer `if` becomes a statement.

The rewrite is charged to the caller: it binds multi-value results into caller
locals, so `MAX_CALLER_LOCALS` declines a caller already carrying more than 128
of them, where the added binds cost more than the boxing they remove.

## NIR: globalization through a constant field borrow

`deref_const_field_borrows` rewrites `&(S { f: v, .. }.f)` to `&v` before
`globalize_const_objects` scans, so a literal borrow that inlining produced
still presents the `InlineRef` shape the candidate scan keys on.
`arena_query::projected_const_field` is the one implementation of that
projection, shared with `const_folding`.

## Results

Release build, arms alternated within one session, best of three per row:

| row              |  gain |
| ---------------- | ----: |
| cbor-canada de   | 20.1% |
| cbor-catalog de  | 19.5% |
| json-catalog ser | 11.9% |
| json-catalog de  |  9.9% |
| json-canada ser  |  7.1% |
| json-canada de   |  7.0% |

`benchmark/README.md` and `wasm-size/README.md` carry the full refreshed
figures. `.claude/skills/wado-performance/dead-ends.md` records the approaches
these benchmarks rejected, and the WEP records why `slot_flatten` still leaves a
sequence element's `Option` boxed.


gfx:
Verified the finding. All three claims hold, and chasing them turned up something
larger than the comment's framing.

**Fixed — the stated pricing rule was the pre-`net_cost` one.** `inline_cost` no
longer prices a callee as written: `CostWalk::call_price` charges the ABI edge
unless the callee carries a loop this pass will splice, and admission compares
that price *less* `call_site_size(param_count)` against the threshold. Both rules
are now stated where the old one was.

**Fixed — the entry's premise was false against this tree.** `uscale` grosses 21
against the -O2 budget of 16, but it takes four parameters, so the call site it
replaces is worth 6 and `net_cost` puts it at 15. It inlines: `wado compile -O2`
on `benchmark/json_canada/json_canada.wado` emits no `uscale` function at all,
while its `fpfmt` neighbours (`unpack64`, `pow10_coarse`, `mul_pow10`) are still
there. The entry read as an open question about how to admit it; the answer is
that the cost model already does, which is the json-canada ser 7% the entry
attributes to raising the threshold to 22. The three arms below it stay — each
fails for a reason that does not turn on how `uscale` is admitted — but they are
now marked as measured while `short` still paid three calls.

**Fixed — `splice_growth`, though not where the comment points.** Line 79 makes
no claim about the growth budget. The `uscale_bounds` bullet does: it said the
inliner takes a single-site candidate because it is "charged nothing". Admission
is the threshold against the candidate's own price; `splice_growth(size, 1) == 0`
only means the growth budget never gets to veto it. The bullet now says which
gate did what.

0cfb3d4

gfx:
Verified and fixed in 9da2339.

The finding is right, and pinning it down took separating two shapes that look
alike. A **value** parameter handed out with `&mut` is boxed by lowering, and
reads then name the box rather than the parameter, so the fold cannot reach it —
`bump(&mut limit, seed)` prices at 26 against a plain 26, no hole. A **`&mut`
parameter passed on** is the real case: it is already a reference, so it carries
no `MutRef` of its own, reaches the next call as a bare place, and only the
argument's `is_mut` says a write happened.

Measured on `pick(s: &mut S, seed: i32)` calling `bump(s, seed)`, with the loop
bound reading `s.n`:

| | plain | folded |
| --- | ---: | ---: |
| before | 25 | 2 |
| after | 25 | 25 |

`written_locals` now reads `ExprKind::Call` and marks the `place_root_local` of
every `is_mut` argument. That is the only node carrying `ArenaCallArg`, and a
method's `&mut self` rides `args[0]`, so one arm covers every call shape.

`inline_folded_cost_mut_arg.wado` is the regression. The bogus fold claims
splicing `pick` deletes a loop, which holds `pick` back from receiving inlining,
so `step` stays a call: 2 call sites without the fix, 0 with it. `bump` carries
`#[inline(never)]` — spliced, it would leave an assignment in `pick` that the
plain scan for writes already catches, and the fixture would pass either way.

`mise run test` and `mise run test-wado` are green (5165 passed, 0 failed); four
goldens shift.

PR #1969, opened 2026-09-05
feat(niri): fold constant string templates, and report the regions that stay

gfx:
## The fold

A template whose interpolations are all constant folds to the literal the source could have written, so the buffer, the `Formatter` and the digit loop leave the binary. Each folded literal is then globalized, built once at instantiation rather than per call.

Two facts together get there.

- The store constant-object globalization leaves at a use site (`{ G = <const>; G }`) does not refuse the region around it, when every mention of that global in the package is one half of such a pair. Folding the region deletes the store and the only read it serves together, so no read is left depending on a store that went away. The condition is that property, not a count of stores: inlining copies the pair, so two sites are as safe as one, while a global with a single store and a distant read is not safe at all.
- A borrow denotes its referent whether or not the binding is mutable. What makes an alias safe is that nothing displaces it, which holds of a `let mut` nothing reassigns, and the executor and the analysis deciding what a frame can track read that from one derivation.

The pair itself is not a region: folding `{ G = <const>; G }` would write the literal back over the naming construct and undo the sharing globalization arranged. Nor is a block that leaves nothing on the Wasm stack — a unit-typed inlined statement call, or the diverging `else` of a `let ... else { panic("…") }`.

Measured at `-Os`, on a program printing one interpolation against one printing the literal (1 997 bytes):

| Interpolation | Cost | Over the literal |
| --- | --- | --- |
| `${"y"}`, `${42}`, `${42:?}` | ~2 015 | ~18 |
| `${7:04}`, `${true}` | ~2 020 | ~24 |
| `${'x'}` | 2 340 | 343 |
| `${255:x}` | 5 197 | 3 200 |
| `${3.5}` | 6 411 | 4 414 |

The first two rows fold to the literal. `char`, `{:x}` and floats do not, and the WEP roadmap says what each needs. `${255:x}` waits on the place-valued field: `Formatter::prepare_int_write` survives as a call taking `&mut Formatter`, whose `buf` is such a field.

## Two designs that kept drawing the same defect

Merging `main` broke the fold, and the two causes were each a design that admits a class rather than one bug.

- `niri/place.rs` held five partial spellings of "what storage does this operand name", each with its own transparent-node set; its own doc named the hazard rather than closing it. One walk resolves an operand to the local it roots at and the steps between, and every question is a predicate over those steps. A node kind the IR gains is taught to that walk alone.
- The `__tmpl` marker lived in a mangled label string that four passes compared by equality and three passes rewrite, so `__inline_<callee>_<n>____tmpl` and `__fused___tmpl` both lost it and the in-loop prune dissolved the block before the fold could run it. A `BlockRole` rides on the labeled-block node instead: lowering reads the label once, and every copy carries the role. One string test remains in the compiler.

## The instrument

A remark names each block that computes a constant at run time and what stopped it: the calls left on its path, or the fact that refused it as a frame. It reads the final IR, never a refusal a pass recorded, so it retires itself as the fold reaches each shape. The fold and the remark ask one derivation for region shape, what a frame would need, and which globals materialize, so a remark cannot blame a store the fold exempts.

A call under a `cold_path` marker is in the block but not on its path — `push` carries `grow` behind its capacity check — so a region filling a pre-sized container names the callee it does reach. Naming nothing is then its own answer: the fold waits on a value the engine cannot represent rather than a body it cannot run, and `WADO_TRACE=ctfe_stmt` names the statement from there. `ctfe_call` names what each declined call wanted that the frame could not give, and `region_seed` every refusal along the fold path.

`mise run report-const-regions` counts the remarks over the benchmark and `wasm-size` corpora and the Wado packages. It counts 1 198 surviving regions across 9 of 22 files:

| Cause | Regions |
| --- | --- |
| no call on its path explains it | 900 |
| it writes a global | 240 |
| `push_encoded_ranges` still runs | 34 |
| it writes a place rooting in no local | 13 |
| `union_char_ranges` still runs | 4 |
| `binary_property_ranges` still runs | 4 |
| `general_category_ranges` still runs | 2 |
| `i32::fmt_decimal` still runs | 1 |

Every callee it names builds a `List<T>` from a constant, and a `List<T>` region is what the no-call row reports, so the census points at one missing capability twice over. Two Gale-generated files hold 1 120 of the regions; what generated code does is call that capability often.

## Documentation

`docs/wep-2026-04-27-nir-interpreter.md` states the design as it stands: where the line falls between the hash-cons pool, the `ValueGraph` and `niri`; what the engine evaluates; how a region is decided; and why the region fold is staged ahead of the call-level fold that mixed templates need. Its roadmap is ordered by the census, and its known gaps carry what closing each would take.

`docs/AGENTS.md`: a WEP may carry Roadmap and Known gaps. They split on commitment rather than size — a roadmap item will be done and says what finishing it means, a known gap is unowned. Moving an item between them is the human's call.

`AGENTS.md`: a failure that shows only in CI is usually the merge with `main`, which the test jobs run on, so sync before suspecting the runner.

https://claude.ai/code/session_01WZ69CfZtkfGeCMEzykzph3

---
_Generated by [Claude Code](https://claude.ai/code/session_01WZ69CfZtkfGeCMEzykzph3)_


gfx:
CI is red and I cannot reproduce it. Recording what I have, because the divergence itself is the finding.

## What fails

`Test (Core)` — `remarks::a_constant_integer_interpolation_folds`:

```
unexpected remarks: ["5:13 remark: this block computes a constant at run time: `i32::fmt_decimal` still runs here"]
```

`Test (E2E O2)` — `niri_const_format_region_fold`, `niri_match_const_fold`, `niri_struct_match_fold`, each `wir_expect:O2 failed: pattern not found`.

Both say the same thing: the constant-template region fold does not fire on the runner.

## The divergence

Same commit `b94ec22a`, same command `cargo test --locked -p wado-compiler --test e2e -- fixture_test_o2`, `niri_struct_match_fold.wado` at `-O2`:

| | `run()` |
| --- | --- |
| runner | `"i32::fmt_decimal$scalar$spec0$scalar"(1013, __local_2)` |
| here | `"core:rt/write_string_to_stream"(tx_21, global:…::__const_obj_1, 1)` |

The match folds on both sides — `1013` is an immediate in the runner's WIR — and only the template around it differs. Deterministic on each side: the failing jobs re-ran and failed identically, and the three fixtures pass here across every configuration I tried.

The fold does work on the runner elsewhere: the golden WIR `wado-bot` regenerated in `chore: tidy` carries all six folded literals from `niri_const_format_region_fold.wado` (`i=42`, `b=true`, `pad=0007`, …). That generator runs one compile per process; the e2e harness runs many in one.

## Ruled out

Build-cache staleness (`CARGO_INCREMENTAL=0` over a full rebuild), stale fixture expansion (`touch tests/e2e.rs`, count checked), hash order (`IndexMap`/`IndexSet` over `FxBuildHasher`), a time-based budget (`step_budget` counts steps and resets per function; the optimizer loop is a fixed iteration count), parallelism inside the compiler (no rayon), env mutation from tests (no `set_var`), toolchain (1.98 pinned both sides), tree state (`lib/` tracked and unmodified), the bundled libm (`lib/core/libm.wat` is checked in), test ordering and thread count (3, default and 16 all pass), and CPU count (`taskset -c 0` still folds).

## What I need

I have spent the one re-run these rules allow, and I am not going to loosen the fixtures to get green — the assertions are what the change is for. Fixing it blind is not something I can validate.

The useful next step is a diagnostic push: a temporary commit that runs the failing fixture under `WADO_TRACE=ctfe_stmt` in CI, so the runner says which statement abandons the frame. That widens the PR with noise, so I would rather have a maintainer's call before doing it — or a second opinion from anyone who can run `cargo test --locked -p wado-compiler --test e2e -- fixture_test_o2` and say which side their machine lands on.

Note that the current head `5fb253f3` has no test signal at all: `Detect changes` gated every test job out of the golden-fixture tidy commit, so the last real run is the one on `b94ec22a`.

---

gfx:
Review response, pushed as 61804676a.

**Fixed**

- `remarks.rs` discarded the compile result, so a source that failed to compile produced no remark and passed every "is empty" expectation for the wrong reason. The assert exposed a real case: `param_used_outside_a_gate_is_not_remarked` declared `run(args: List<String>)`, which the CLI world rejects, so the test had never checked anything. Its source now matches its comment. The assert lives in one helper both entry points share, because the `-D` sibling had the same hole and its own copy of the extraction. I confirmed the guard fires by running a deliberately broken source through it before removing that probe.
- The labeled-block test builders derive the role from the label instead of hardcoding `Plain`, so the two that build a `__tmpl` block stop describing a node lowering never produces.
- Three materialization refusals that returned silently now trace.
- The WEP's `switch` roadmap item was a garbled sentence; rewritten as two plain ones.
- The CI note in `AGENTS.md` is scoped: the test jobs run on the merged tree, and `tidy` is the exception since it checks out the head ref.

**Skipped**

- Shortening "whether or not it will be closed" in `docs/AGENTS.md` inverts the meaning. The phrase says a gap is recorded either way, which is what the following sentence ("with no claim that it will be") relies on.
- Refusing every `writable_reference` seed in `frame.rs`. The scenario needs a writable reference over a non-scalar value seeded into a region whose alias-local write is then dropped. Boxing covers references to primitives, enums, variants and function types, never struct aggregates, so "a boxed mutable aggregate" does not exist here; the shape that does is a `&mut` to a variant, and I probed it along with three other shapes. All are correct at every level, and each region refuses through the outer-write rule or a non-constant seed, never through the scalar test the suggestion targets. The guard as written also predates this behaviour: the previous condition admitted the same non-scalar case, so the suggestion is a new, broader conservatism rather than a restoration, and it would cost the folds this PR adds. The sound closure is the alternative in that comment, resolving a region-local alias to the caller-owned root before the outer-write analysis, which I would rather do as its own change.

Two earlier threads are stale as of this head: the fixture is registered and its O2 test runs, and the CI folding failure is green.

---

gfx:
Second review round, pushed as 56d5a0909.

**Fixed**

- The negative-length finding is real. `Value::Int` holds the sign-extended bit pattern, so a negative `i32` length read back as a huge `u64`, `usize::try_from` accepted it, and the element-count check reported "used &lt;huge&gt; of N elements" where the refusal meant "negative len". The length is decoded at its own width now. Behaviour is unchanged, since both paths refuse; what changes is that the diagnostic tells the truth.
- The `switch` sentence in the WEP reads "later" rather than "then".

**Skipped**

- The focused test for a negative length. A sequence length is a `used` field the compiler itself writes, so a negative one cannot arise from a well-formed program. The branch is a defensive guard, and a fixture would have to construct the value it then asserts on, which passes CI and guards nothing.

Also in this commit, from a distill pass over the branch: the derived materializing-global set now records why it is rebuilt per call, since it reads body contents that the count-keyed fold cache does not track and moving it into that cache would leave a stale set behind.

---

gfx:
@coderabbitai review

PR #1974, opened 2026-09-05
refactor(elaborator): resolve a qualified call from one declaration index

gfx:
A reference site is resolved once, and the consumers take that answer rather
than each deriving one from the name.

## One index behind the qualified spelling

`ImplMethodIndex` holds every method an impl block declares, keyed by the
receiver (`ImplTargetKey`). An `ImplMethodEntry` carries the declaring module,
the inherent visibility, `has_self` and the method's `DefId`, so a lookup that
wants one kind filters the walk instead of reading a second index.

`impl_method_entries` is that walk. Which declarations the spelling can name,
whether a type declares the name at all, whether a newtype answers or its base
does, the signature, the return type, the parameter types, `mut` and defaults,
the visibility and the Component Model binding are all projections of it. The
three spellings reach it alike: `Type::method`, `Type::<T>::method` and
`ns::Type::method`.

The spelling names two kinds of callee — a receiver-less method, and an instance
method whose receiver the call passes as its first argument (`P::add(&p, 1)`).
A struct, a newtype, a `variant`, a `flags` type, a CM resource and an imported
type each answer for both.

## An inherent declaration shadows a trait impl's, within a kind

A type head names the type's own method, so `S::tag` is the inherent `tag`
where one exists, as dot syntax already resolves it; the trait's is
`Trait::tag`. The shadow holds within a kind. An associated function and a
method taking a receiver are not alternatives for one call, because different
argument lists reach them, and shadowing across the two would leave the
associated function no spelling at all: `Trait::tag()` has no receiver to infer
`Self` from.

Which trait impl _block_ the arguments pick is a different question and keeps
its own walk, since a block that declares no method of the name still answers —
`impl Factory for Wrap<T> {}` gives `Wrap::make()` the trait's default. That
selection consults the resolver for the one fact it shares: whether an inherent
associated function shadows the name.

## A call's facts come off the declaration it selected

`CalleeParams::of_signature` derives `param_types`, `param_is_mut` and
`param_defaults` from one signature. Written qualified, an instance method's
receiver is the first argument, so each list carries a leading entry for it:
spelled at the call site and never omitted, hence no default, and `mut` exactly
when the method takes `&mut self`. Visibility, arity and the return type read
the same declaration.

The dot spelling takes its `mut` flags off the declaration dispatch chose, in
place of a second read by name. The two disagree where an inherent method and a
trait impl's share a name, and a parameter wrongly flagged non-`mut` loses the
copy that value semantics owe it: the callee then grows the caller's list.

## A canonical states its own core type

A cancel answers with the `u32` its copy ended on, and `fn cancel_read(&self)`
declares no result. `CanonicalIntrinsic::returns_discarded_result` names those
intrinsics, and one binding in `wir_build` both types the core import with the
result and drops the value, so the module validates.

`CanonicalIntrinsic::stream_op` and `future_op` map a base name to its variant
in the one place a `stream-*` or `future-*` name is spelled. `resource_rewrite`
derives a member's binding from `CompilerItem` and from
`CanonicalIntrinsic::from_import_name`, and its test reads the primitive names
out of the `#[cm(...)]` attributes on the stdlib's `resource` declarations.

## WEP

`docs/wep-2026-05-26-elaborator-rearchitecture.md` records what the qualified
spelling can name, the per-kind shadowing rule, and why trait-impl block
selection keeps a walk of its own.

https://claude.ai/code/session_01298RP7pus2DHCvLUTciQUL


gfx:
The benchmark alert (`zlib/decompress -O2` 1.53, `json/twitter/ser -O2` 1.55) is the host, not this branch.

Bytes settle it without timing anything. Compiling every benchmark under `origin/main`'s compiler (`3ad1adb6b`) and this branch's, 15 of 16 come out byte-identical — `zlib` and `json_twitter` among them. A row whose emitted wasm is identical cannot have moved.

The one that differs is `benchmark/gale_gen/gale_gen.wado`, 289 bytes smaller: two `Slice<T>::internal_new` monomorphizations the base emitted are no longer generated, for `Alternative` and `AttrDecl`. It compiles and runs (210.40 KB/s). Neither alerted row is that one.

---

PR #1978, opened 2026-09-05
feat(compiler): tagged template literals over ReflectTemplate

gfx:
A path written directly before a backtick is a tag: the template denotes a call
of that function on the holes, in their own types, with the literal text around
them as constants.

```wado
let q = sql`SELECT * FROM users WHERE id = ${id} AND name = ${user.name}`;
let s = String::raw`${dir}\bin\run.exe`;
```

A tag is an ordinary function — any effects, any return type — whose first
parameter is bound by `ReflectTemplate`, a sixth reflected kind synthesized per
template shape rather than per declaration. It walks the holes with tuple
`for-of`:

```wado
fn sql<T: ReflectTemplate<Holes = [..V]>, ..V: ToParam>(t: T) -> Query {
    let mut text = "";
    let mut params: List<Param> = [];
    for let h of ReflectTemplate::<T>::members() {
        text.push_str(&h.lit());      // the literal before this hole
        text.push_str(&"?");
        params.push(h.get(&t).to_param());  // the value, typed
    }
    text.push_str(&ReflectTemplate::<T>::tail());
    return Query { text, params };
}
```

A `Hole<T, V>` answers `index()`, `lit()` / `raw()` (the preceding segment,
escapes processed or preserved), `source()` (the expression as written),
`has_spec()`, `get(&t)`, and `fmt(&t, f)` (rendering as the untagged template
renders). Everything but `get` and `fmt` is a constant, so a monomorphized tag
body is one constant append per segment and one typed operation per hole —
`tagged_template_sql.wado` holds it to that with `wir_not_expect:O2`.

`..V: ToParam` is what makes a hole of an unsupported type a bound error at the
call site rather than a missing-method error inside the unrolled body.

## The type

Each tagged template denotes a compiler-synthesized struct, unique to its
static shape (raw segments, specifiers, hole source texts, hole types), holding
one field per hole. Two templates of one shape share the type and every
instantiation of the tag. The type is anonymous and reached only through the
bound. Its name is a SHA-256 digest of a length-prefixed encoding of the shape,
so the name a compiled artifact carries does not move with the Rust toolchain.

A hole's field carries its value without copying and without boxing: a type
whose reference is a bare GC handle is held as `&V`, and one the compiler would
box is held as `V`. `TypeTable::is_boxed_reference_target` is the single
predicate for that, which `lower::plan::boxing` reads too. Whether a hole's
field holds its value by handle is decided per hole, so a read bridge shared by
a newtype and its base derefs only the one behind a handle
(`read_bridge_mixed_newtype_group.wado`).

Holes are evaluated in source order, and the literal is built after the last of
them, so a hole that names storage is read in situ only while nothing after it
can write that storage. That is what holds `` format`${a} ${bump(&mut a)}` `` to
what `` `${a} ${bump(&mut a)}` `` renders.

## The prelude tags

`format` is what an untagged template means — every hole rendered through its
specifier into one buffer — and `String::raw` keeps the escapes as written. A
raw template's escapes must still be ones the lexer knows; only their spelling
is preserved. `tagged_template_format_equiv.wado` holds `format` to the untagged
form across the specifier matrix.

## Elsewhere

- `Expr::TaggedTemplate` in the AST, parsed as a postfix on a path with the
  backtick adjacent by byte span. `wado format` round-trips it, `Wado.g4` has a
  `taggedTemplate` rule, and both the Gale query and the LSP colour the tag as
  the function it names.
- `example/sql_tag.wado` is the worked example as a program that runs.
- `docs/wep-2026-01-10-tagged-template-literals.md` is rewritten around this
  design; `docs/spec.md`, the cheatsheet and `docs/compiler.md` follow it.

## Fixes this carries

- A type-argument list is dense: an effect or `fn`-bound parameter holds no slot
  in one. Bound enforcement pairs against that space, and the associated-type
  projection runs before the check, so one enforcement answers for every call
  kind (`reflect_pack_bound_fn_param.wado`,
  `reflect_pack_bound_free_fn_error.wado`).
- An anonymous struct's mangle is fixed when the shape is interned, so newtype
  erasure cannot re-render it into a name WIR never registered
  (`anon_struct_newtype_field.wado`).
- A read bridge groups its members by their erased value type, and decides the
  dereference per member, so a group holding both a newtype and its base reads
  each field the way that field stores it.
- A resolved argument carries the span it was written at, or the call's where
  the source spells none.
- `T::method(...)` on a type parameter receives its argument.
- The generated stdlib reference walks a re-exported module once, however many
  re-exports name it.

## Known gaps

Recorded in the WEP, and the three that are compiler bugs are pinned as
`#[TODO]` fixtures: a generic static method reached through a namespace import
or answered by a blanket impl, and the missing argument check on the
type-parameter static path. Each fails the same way when spelled as an ordinary
call. A turbofish on a tag (`` tag::<i32>`x` ``) is a parse error: the
speculative turbofish parse commits only on `(`.

https://claude.ai/code/session_01Mo276GhfQeGf4ZnzpF9Z7h


gfx:
Answered in `036e3ec67`. Each finding was reproduced against the current code before anything changed.

## Fixed

**`synthesis/traits.rs` — one group, many value types.** Real, and worse than the report: the bogus deref reaches lowered NIR. A `struct Span { len: Meters, raw: f64 }` (with `type Meters = f64`) emitted

```wado
pub fn "$field_get$…/Span$f64"(v: &Span, index: i32) -> f64 {
    return match index { 0 => v.len, 1 => *v.raw, _ => … };
}
```

and the hole bridge emitted `*v.h1` on a field holding its `f64` by value. `try_boxing_deref` rewrites only the one that is a box, so the other survives as a NIR `Unary(Deref)` on a scalar.

The fix decides per member rather than through the representation head: `ReadBridgeMember` already records both the field type and the value type, and they differ exactly when the field holds its value by handle. `member.field_type == member.value_type` is that test, and the deref answers `member.value_type`. The group's first member still names the helper's return type, which is what the grouping already guarantees. `read_bridge_mixed_newtype_group.wado` covers both bridges; no existing golden changed, so the case was genuinely uncovered.

**`tir.rs` — `DefaultHasher`.** Real. The mangle reaches the Wasm name section and the checked-in WIR goldens, so a Rust upgrade would move `$tmpl$…`, `$hole_get$…` and `$hole_fmt$…` with no source change. Now a SHA-256 (`sha2` is already a compiler dependency) over an explicit encoding: counts as `u64` LE, every string length-prefixed, `spec` tagged so `None` and `Some("")` differ. The first 8 bytes keep the existing 16-hex width.

**`docs/spec.md` and the WEP — `String::raw` example.** Real, and the sharper form of it: `` String::raw`C:\path\${name}` `` does not compile, because `\p` is not an escape the lexer knows (`elaborator/util.rs` `unescape_one`) — and `\$` before the hole would suppress the interpolation besides. Replaced with `` String::raw`${dir}\bin\run.exe` ``, verified to run, and the spec now says escapes must be well-formed even where a tag preserves their spelling. The PR description carried the same example and is updated.

**WEP — "takes one argument".** Real, contradicted by `tagged_template_tag_default_arg.wado`. Now: the first parameter is the one bound by `ReflectTemplate`, the template is the one written argument, and ordinary call resolution fills trailing defaults.

**`wep-2026-09-05-total-reflection.md` — `match type`.** Real: the example is stated to be exhaustive and `TypeInfo::Template` had no arm. Added `template(Holes = [..V])`.

**`tagged_template_bad_tag_error.wado`.** Both tags do diagnose (checked), so `compile_error_count: 2` pins them. Added.

**`dump_test.wado`.** Fair: absence of `Leaf(alt 0)` also holds for a `NoViableAlt`. Ran the test to capture the real text and asserted it — `Ambiguous([alt 0, alt 1]) — opaque rule-ref prefix` — so the test now pins the opaque config surviving into the tournament.

## Skipped

**`reflect_pack_blanket_static_todo.wado` has no `__DATA__`.** Not a violation. A fixture without a `__DATA__` section runs under the test world (`tests/e2e.rs`), which is exactly what a `#[TODO]` module carrying a `test` block wants: the compile failure is caught and reported pending, and the fixture hard-fails if it ever starts passing. Verified — it passes as pending at O0 and O2. Several existing `#[TODO]` fixtures are shaped the same way.

**`tagged_template_not_path_error.wado` message.** The expectation matches what is emitted. `` pick()`x` `` never reaches the resolver: the parser rejects a non-path tag at the backtick with *"a template tag must be a function name or a static method path"* (`parser.rs`). `wado check` on the fixture prints exactly that. The resolver message is a different diagnostic, covered by `tagged_template_bad_tag_error.wado`.

**`substitute_idents` not recursing into `t.tag`.** The scenario is unreachable. A tag naming an earlier parameter is rejected where it is written, before any expansion: `fn f(tag: i32, s: String = tag`x`)` reports *"expression is not callable: type 'i32' is not a function"* at the declaration. So there is no program in which the substitution would change a tag, and the call would be dead code reading as if it mattered.

**`walk_expr` not visiting `t.template.id`.** Nothing looks that id up — the elaborator keys the template's facts on the `TaggedTemplateExpr`'s `AstId`. Visiting it would make things worse: `ast_id_at` prefers the *narrowest* containing span, and the nested literal's span is strictly inside the tagged expression's, so every position on literal text would resolve to an id no phase records a type for.

**Turbofish tags in the LSP and the grammar.** The premise does not hold: `` tag::<i32>`x` `` is a parse error today — the speculative turbofish parse commits only on `(`, so the rollback leaves `tag` and the `::` fails. Both highlighters therefore match what parses. Whether a turbofish should be accepted on a tag is a language decision, not a highlighting gap; it is now recorded under Known gaps for @gfx to rule on.

**`anon_struct_name` / widening comment.** Both facts are already stated at their definitions — `anon_struct_name` says it is the declaration namespace and that `anon_struct_mangle` is what a key is built from, and `struct_head_decl_name` says it is never a lookup key. Repeating them at the widening site would be a third copy.

The duplicate `format` in `docs/stdlib-core-prelude.md` was already fixed in `95080cac7`: the doc generator walked a re-exported module once per re-export, which also duplicated `Eq` and `Display`.

`mise run test` is green on this head.

---

gfx:
Following the `defaults_module` finding to a repro turned up a P0 next to it, fixed in `73110f311`.

## The bug

A static method on an inherent `impl` that lives in a module other than its type's home ICEs the compiler:

```
a call minted an extern stub for a name the package defines:
  Free(… module_source: "./a.wado", name: "./a.wado/S::make")
  vs defined Free(… module_source: "./b.wado", name: "./a.wado/S::make")
```

This is legal Wado — `docs/spec.md` §Inherent Impls makes the coherence unit the *package*, not the module, and `cross_module_inherent_third.wado` already covers the **instance** method on exactly this shape. The static was the uncovered half. A default argument is incidental: a plain `Gadget::of(21)` reproduces it.

## Where it actually was

Not at the line the finding names. That call takes the *qualified* path, `resolve_static_method_call_from_qualified`, whose `StaticMethodRef` fallback used `declaring_module_of(struct)` — the type's home — while the definition is registered under the impl's module. Patching the other site left the failure unchanged, which is what pointed here. The fallback now takes the inherent entry's module where one is indexed, matching what the instance path already did.

`cross_module_inherent_static.wado` (with `sub/cross_module_inherent_static_impl.wado`, reusing the existing `…_third_type.wado`) pins both the plain static and the defaulted one. Both were red before the fix and pass after.

## On `defaults_module` itself

Applied, but I want to be accurate about why. I could not construct a program where `struct_module` and `static_method_module` differ *observably* on that path: the cross-module inherent case goes through the qualified path fixed above, and a trait static's default is not picked up through `Type::method()` at all (`expected 1 arguments, found 0` — a separate gap, not this one).

What justifies the change is consistency, not a reproduction: annotate resolves those defaults in `static_method_module` (the module they were written in, per the comment at the binding) while reify was handed `struct_module`. One value read from two places is the divergence class that becomes the ICE above. If you would rather not carry an unreproduced change, that hunk is separable from the fix.

## Not taken

**`liveness.rs` — `Item::Resource` in `compute_last_uses()`.** `seed_operations()` seeds resource methods, so the asymmetry is real, but the consequence stated is a *missed* copy elision — a pessimization, not a defect, and nothing observable to write a test against. Adding coverage there is its own change with its own measurement, not a rider on this branch.

`mise run test` is green on this head.

---

PR #1979, opened 2026-09-05
chore(hooks): drive the review-response and distill steps from hooks

gfx:
Three hook-driven steps, and the skill rules they point at.

**Answering a review.** A `/code-review` run, typed or called as a skill, draws a reminder to answer its findings with the `code-review-response` skill. That skill states the altitude to answer at: a finding names one site, and the fix belongs where the class of defect is admitted — the missing invariant, the type that allows the state, the call site nobody has to remember. `AGENTS.md` and the skill's own description name a `/code-review` run alongside PR review comments.

**Distilling.** A `Stop` hook asks for a distill pass over the branch, once per session, when the worktree is clean and the branch carries commits against its merge base. A dirty worktree is the environment's commit-and-push hook's to demand, so the two Stop hooks run in sequence rather than both at once. Invoking `distill` records that it ran, under `.git/wado-hooks/`; `stop_hook_active` keeps the nudge from repeating.

**Watching a pull request.** `pull-request` says to subscribe with `subscribe_pr_activity`, to handle every event it delivers, and to answer a review with `code-review-response`.

**The command guard.** `perl`, `setsid` and `disown` are denied wherever the word resolves — nested in `bash -c`, behind `xargs`, `env` or `timeout` — which is the reach `permissions.deny` lacks, matching on a literal prefix.

`skill-invocation.mts` names the skill a hook payload invokes, covering both the `/name` prompt and the `Skill` tool call, so a hook compares a name instead of matching a pattern against one. `mise run test-hooks` covers each hook's decision.

https://claude.ai/code/session_01KQLDnS14EdK8qiSUyWqiRB

### DIFF
--- a/.claude/hooks/code-review-followup.mts
+++ b/.claude/hooks/code-review-followup.mts
@@ -1,49 +1,28 @@
 #!/usr/bin/env node
 // Hands the findings of a /code-review run to the skill that answers them.
-// Two events, because a skill reaches the session two ways: the user types
-// `/code-review` (UserPromptSubmit), or the model calls it (PostToolUse > Skill).
+
+import { type HookPayload, invokesSkill, readPayload } from "./skill-invocation.mts";
 
 const REMINDER =
   "The findings of this review are claims to answer, not a task list. Invoke the" +
   " `code-review-response` skill once the review reports: verify each finding, fix the class" +
   " it names, and report what was skipped and why.";
 
-const SLASH = /^\s*\/code-review(\s|$)/;
-const SKILL = /(^|:)code-review$/;
-
-export function followUpContext(payload: unknown): string | null {
-  const event = payload as {
-    hook_event_name?: string;
-    prompt?: string;
-    tool_name?: string;
-    tool_input?: { skill?: string };
-  };
-  switch (event?.hook_event_name) {
-    case "UserPromptSubmit":
-      return SLASH.test(event.prompt ?? "") ? REMINDER : null;
-    case "PostToolUse":
-      return event.tool_name === "Skill" && SKILL.test(event.tool_input?.skill ?? "")
-        ? REMINDER
-        : null;
-    default:
-      return null;
-  }
+export function followUpContext(payload: HookPayload | null): string | null {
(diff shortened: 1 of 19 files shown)


gfx:
Answered the review. Three fixed, one skipped.

**Fixed — `skill-invocation.mts:13`, a name that runs past its word.** Reproduced: `/code-review/foo` and `/code-review?x` both named `code-review`. The name now runs to the end of its word (`(?=\s|$)`), so a path or query suffix names no skill. Both hooks read the payload through this one function, so the boundary holds for `distill` too.

**Fixed — `skill-invocation.mts:23`, a field of the wrong type.** Reproduced: `{"skill": 7}` threw `TypeError: s.replace is not a function`. A field of any type but string now names no skill. Same class as the finding above — the payload is parsed JSON, and a declared field is a claim about its shape — so both fixes live where the payload is read, with `skill-invocation.test.mts` covering the naming rules that the two hook tests used to duplicate.

**Fixed — `code-review-response/SKILL.md`, a report with no pull request.** The Report section now names the target for a review that arrived outside a pull request: the same list, in the session.

**Skipped — `distill-reminder.mts:20`, intent recorded as completion.** The claim is true: an aborted `/distill` leaves the marker reading `done` and suppresses that session's reminder. The proposed fix needs a completion event that does not exist. `UserPromptSubmit` fires when the prompt is submitted and `PostToolUse` when the `Skill` tool returns the skill's instructions; neither observes the pass finishing, and `Stop` cannot tell an abort from a turn that ends after the work. Recording intent is the strongest signal available, and the case it misses is a person deliberately abandoning a pass they asked for.

`mise run test-hooks`: 16 pass.

---

gfx:
Fixed — `skill-invocation.mts:24`, the coercion the reviewer names.

Reproduced: `String(["/distill"])` is `"/distill"`, so an array prompt named a skill. The prompt is now taken only when it is a string, which is the rule the `skill` field already followed; the `String(...)` call was the one place the boundary leaked. `skill-invocation.test.mts` covers an array on both fields, since that is the value coercion turns into a valid name.

`mise run test-hooks`: 16 pass.

---

PR #1980, opened 2026-09-05
fix(compiler): lower every `i128` / `u128` pattern, on one type-table spelling

### LINKED ISSUE #1970: Retire `PrimitiveType::I128` / `U128`: `i128` has two spellings in the type table
## The duplication

The user-facing `i128` / `u128` are prelude declarations:

```wado
// wado-compiler/lib/core/prelude/int128.wado:57
#[compiler_item("u128")]
pub struct u128 { low: u64, high: u64 }
```

A value of that type carries `ResolvedType::Struct`, which is what reify's cast lowering matches on:

```rust
// wado-compiler/src/elaborator/reify.rs:9481
ResolvedType::Struct { def, .. }
    if self.tysys.type_table.borrow().struct_head_name(def) == "i128" =>
```

`PrimitiveType::I128` / `U128` exist beside them, interned at type-table init as `TypeTable::I128` (`TypeId(4)`) and `TypeTable::U128` (`TypeId(9)`). So one source-level type has two representations in the type table.

Both lower to the same code, so which one a phase holds has never been observable. That is why it has survived.

## Why remove it

Two spellings for one type means every phase asking "what is this type" can get either answer, and nothing states which is correct. It is also about to stop being invisible: a classification over `ResolvedType` — the shape `TypeTable::reflect_kind` already has — would answer `Primitive` where the surface type is a struct, so the same source type could be classified two ways depending on the call site.

## Survey

89 references across 17 files. Almost all of them are match arms; the producers are few:

| Kind | Site |
| --- | --- |
| Interned at table init | `tir.rs:935`, `tir.rs:940` |
| Name → type in builtin signatures | `builtin_registry.rs:178`, `builtin_registry.rs:183` |
| Listed as a primitive name | `tir.rs:318` (`all_primitive_names`) |

The other ~86 sites are defensive arms — `wit_emit`, `const_eval`, `lower/plan/boxing`, `lower/plan/globals`, `synthesis/cm_binding`, `wir_build`, `optimize` — each special-casing "this primitive is composite". If the producers above are the only ones, most of those arms never fire and go with the variants.

`component_model.rs:4392` maps `TypeTable::I128 | U128` to `ValType::I64` with the comment "Default to i64 for now (caller should handle specially)"; it becomes unreachable once the producers are gone.

## One place that needs care first

`builtin.wado` imports the struct:

```wado
use { i128 } from "core:prelude/int128.wado";   // :21
pub fn v128_const(value: i128) -> v128;          // :505
```

But `builtin_registry.rs` resolves a builtin signature's type name by matching the name before considering imports (`:173`, "Otherwise, check for primitive types"), so `"i128"` resolves to `TypeTable::I128` — the primitive — while a caller passes `ResolvedType::Struct`. Those are different `TypeId`s.

This has not been traced end to end. Either something reconciles the two, or `v128_const` is called in ways that hide it. It may be a latent bug rather than dead weight, so it is worth settling before the mechanical removal.

## Suggested order

- [ ] Add an e2e fixture calling `v128_const` with an `i128` value, pinning current behaviour. If it is broken, that is a compiler bug and takes priority over the rest of this issue.
- [ ] Remove `PrimitiveType::I128` / `U128`, the two intern lines, and the two entries in `all_primitive_names`.
- [ ] Delete the arms left unreachable, and make `builtin_registry`'s name resolution reach the prelude struct.
- [ ] Confirm nothing regresses in `wit_emit` / CM binding, where 128-bit types cross the Component Model boundary.

## Context

Found while designing [WEP: Reflection over an Unknown Type](https://github.com/wado-lang/wado/blob/main/docs/wep-2026-09-05-reflect-unknown-types.md), which classifies `i128` / `u128` as `Struct`. That WEP assumes this issue is closed by the time its classification is implemented, so it carries no workaround for the dual spelling.

gfx:
`i128` and `u128` are prelude struct declarations in `core:prelude/int128.wado`.
The type table holds exactly that: `PrimitiveType` spells the scalar integers,
and every phase asking what a wide integer is gets the struct.

`TypeTable::wide_int_item` answers "is this a wide integer" by declaration
identity, so a user type sharing the name is not mistaken for one.
`primitive_by_name` is the only by-name primitive resolution, and it lists no
wide integer — `builtin_registry` resolves the `i128` that `core:builtin`
imports for `v128_const` through the declaration path instead, matching the
`ResolvedType::Struct` its callers pass.

Two predicates carry the consequences. `is_integer` answers `false` for a wide
integer: no Wasm integer instruction takes one. `is_boxed_reference_target`
answers `false` through its struct arm, so `&i128` stays the value's own GC
handle rather than a `Box` cell.

Retiring the duplicate spelling is what exposed the six compiler bugs this
fixes; [#1970 records each one](https://github.com/wado-lang/wado/issues/1970#issuecomment-5555701919),
with the acceptance criteria they add.

## Wide-integer patterns

A wide-int pattern lowers to the `Eq` / `Ord` call the equivalent comparison
does, in every pattern position: a literal or a range, at the top of a `match`
or nested inside a tuple, under `if let` or `matches`, with or without a guard.

Pattern lowering emits scalar comparisons — it holds the type table immutably
and cannot intern the prelude calls itself — so `convert_expr` rewrites an
integer literal of a wide-int type into its constructor call and a wide-int
comparison into the trait call. One site, which every pattern position passes
through whatever nests it. `lower::wide_int_literal` builds both, and reads a
constant back out of either shape a literal takes: the constructor call, or the
struct literal inlining leaves in its place.

## Range endpoints

One decoder answers what integer a range endpoint denotes, and it takes the
scrutinee's signedness: `util::range_endpoint_to_i128`, over
`util::parse_int_bits`. Read unsigned, a bound above `i128::MAX` keeps its bit
pattern. `range_endpoints_ordered` orders two such endpoints the same way, so
the reversed / empty-range diagnostic and exhaustiveness agree with the code
generated for the arm.

`is_unsigned_int` follows the newtype chain, as `is_integer` beside it does, so a
range pattern on a newtype over `u32` compares unsigned.

## Component Model

A wide integer crosses the boundary as its own record, `record i128 { low: u64,
high: s64 }`, pinned by `wit::wide_int_export_emits_its_prelude_record`.
`docs/spec.md`'s type-correspondence table records that.

Closes #1970

https://claude.ai/code/session_01NChadteAv6dzLgWtDR1b3f


gfx:
On CodeRabbit's **Out of Scope Changes** warning: the observation is right — the wide-integer pattern-lowering and range-endpoint work is not in #1970's checklist. Taking the second of the two suggested resolutions rather than the first: I have recorded the scope and acceptance criteria on [#1970](https://github.com/wado-lang/wado/issues/1970#issuecomment-5555701919), and the work stays here.

Three reasons splitting it out would be worse:

1. `AGENTS.md` makes a compiler bug P0: "The instant you suspect one, stop all other work, and as the top priority write a minimal reproducible e2e fixture and fix it. A workaround that lets the current task proceed is never a reason to skip or defer any of these." Four of these are compiler bugs — one ICE on every `builtin::v128_const` call, two silent wrong-branch cases, one ICE on a wide-int literal nested in a tuple pattern.
2. #1970's own first checklist item anticipates exactly this: "Add an e2e fixture calling `v128_const` with an `i128` value, pinning current behaviour. If it is broken, that is a compiler bug and takes priority over the rest of this issue." It was broken.
3. The fixes are not separable in practice. They hang off `TypeTable::wide_int_item`, which this PR introduces, and the pattern bugs are what retiring the dual spelling exposed — the issue predicted as much ("It is also about to stop being invisible"). A split would stack on this branch, not stand apart from it, and would mean landing the refactor with known wrong-code bugs it had just surfaced.

---

PR #1982, opened 2026-09-06
feat(gale): compose `import S;` against the supplied inputs

gfx:
An ANTLR4 grammar import resolves, so a `.g4` can be built from more than one file: a dialect extending a base grammar, or a host nesting another language inside itself.

## Resolution

A Kiln generator reads nothing — its inputs arrive by value — so ANTLR4's `-lib` filename lookup has no analogue. `import S;` binds to whichever supplied input declares `grammar S` in its header:

```wado
use { Parser } from "./MiniHtml.g4" with {
    generator: { module: "wado-lang:gale@0.1", inputs: ["./MiniCss.g4", "./MiniJs.g4"] },
};
```

Resolution by name is what the corpus requires: `DelegatorInvokesFirstVersionOfDelegateRule` writes `import S,T;` where the extracted files are `.slave1` (grammar `T`) and `.slave2` (grammar `S`).

A name identifies exactly one input, and every way of failing that is loud. An import naming no supplied input, two inputs claiming one name, and an input that reaches the composite neither as a master nor through an import are each an error naming the grammar. The clause itself needs at least one name and its `;`, so a mistyped one is an error where it is written. Nothing is dropped in silence.

## Composition

`compose_grammars` partitions the inputs in one pass. A grammar another input's import list names is a delegate; everything else is a split half of the master and concatenates through `merge_grammars` as before, so a grammar with no import anywhere in the set takes that path untouched.

Delegates fold in depth-first in import order, the first version of a rule winning, per `vendor/antlr4/doc/grammars.md`. Master rules come ahead of imported ones, which is what makes lexer precedence and token numbering come out as ANTLR4's. A `tokens { }` placeholder yields its slot to a real rule from anywhere in the composite, so the whole thing has one token space. An overridden rule takes its own references with it. Named actions merge; options, `superClass` and `language` among them, do not. A mode an override emptied is discarded, and a mode a surviving `pushMode` targets is kept.

`import Foo = Bar;` is a loud error. Its alias names the delegate for the qualified action references Gale has no counterpart for, and accepting the form while dropping what it means is worse than refusing it. This is the one place Gale rejects a grammar ANTLR4 compiles; claim (a) in `antlr4-compatibility.md` carves it out, and `TODO.md` carries what closing it takes.

Contract and the embedding recipe: `package-gale/import.md`.

## Descriptor corpus

All 17 `CompositeLexers` / `CompositeParsers` descriptors are under the ordinary eligibility rules: 15 claim (b) parse-accepts, 2 claim (d) token dumps that pin the composite's token numbering, and 11 Stage C output-compares, none triaged. `[type]` normalises `CompositeParser` / `CompositeLexer` to the recognizers they are, the slaves ride into the emitted `use` as `inputs`, and a composite's claim (a) asks `assemble_grammar` so the composition itself is what the claim asserts. `TestChunkSpec.slave_count` has no default anywhere, so a claim added later cannot hand the generator the master alone.

Stage B′ still excludes composites: `antlr4-oracle.sh` invokes the jar on one grammar file with no `-lib` slave lookup.

## Example

`package-gale/example/` gains three grammars that compose into one recognizer — one lexer with three modes, one parser, one tree — with three highlight queries riding in beside them. What it demonstrates is one line of `MiniJs.highlights.scm`:

```scheme
(params (JS_IDENT) @variable.parameter)
```

`arrow` and `group` both open with `(`, so whether an identifier between the parentheses is a parameter is settled by the `=>` after the closing one:

```js
let add = (a, b = (1 + 2)) => a + b;   // a, b are parameters
let one = ((a));                       // a is a variable
```

A lexer classifies the identifier before that token exists. A regex highlighter looks ahead, but a default value nests parentheses, so `\(([^)]*)\)\s*=>` stops at the inner `)`; reaching the `=>` means matching brackets. The parser matches them and the query reads the answer off the rule stack. `highlight_test.wado` pins the page's whole HTML.

Embedding needs no feature beyond composition, given four conventions in a delegate: its lexer rules live in a mode of its own, its token names are prefixed, its parser rules name their tokens rather than spelling them, and no token of its own spans the host's terminator. That last one is why a delegate's scanning rules stop at `<`: a composite has one lexer, so a `JS_COMMENT : '//' ~[\r\n]*` would run straight through `</script>` and leave the rest of the document lexed as JavaScript, and a mode cannot express HTML's rule that a closing tag ends raw text whatever the embedded language thinks. The host declares the same mode name for the token that leaves it, so it owns both boundaries and neither delegate names its host. The trade is that an embedded delegate does not lex on its own; `TODO.md` records what a grammar usable both ways would need.

https://claude.ai/code/session_01Q2Kw2ucQ6fpgwK1KH8CzLC

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q2Kw2ucQ6fpgwK1KH8CzLC)_


gfx:
Answered the review in e52575f47. Three of the four findings reproduced; one did not.

**Fixed — a delegate token hiding the host's terminator (`MiniCss.g4`, `MiniJs.g4`).** Real, and worse than the finding says: I reproduced it on ordinary valid input. `<script>// note </script><p>x</p>` had `JS_COMMENT` swallow the closing tag, so the mode never popped and `<p>x</p>` came out lexed as JavaScript. `JS_STRING` and a terminated `CSS_COMMENT` did the same; only an *unterminated* `/* ...` recovered, because `.*?` has no match without its `*/`.

The class is not those three rules. A composite has one lexer, so inside the mode a delegate rule is matched like any other, and a mode cannot express HTML's rule that a closing tag ends raw text whatever the embedded language thinks — deciding that needs lookahead a mode does not have. So it is a fourth embedding convention in `import.md`: a delegate's scanning rules stop at `<`. The `<` that then cannot appear inside a string or comment is listed with the other deliberate omissions in each grammar, alongside tests pinning that each of the three inputs ends its block and highlights the rest as HTML.

**Fixed — the `import` clause accepted a malformed list (`src/g4/parser.wado`).** `import ;` recorded no delegates and returned `Ok`. `import S, ;` and `import S T;` also left the clause silently, surfacing later as a missing rule. The clause now requires at least one name and its `;`, with a test for each of the three.

**Fixed — the compatibility promise (`README.md`, `AGENTS.md`).** Both stated it without qualification; both now point at the exception list under claim (a) in `antlr4-compatibility.md`, which `import Foo = Bar;` joined.

**Skipped — numeric channel ids in `rebase_lexer_rules` (`src/ir.wado`).** Not real, in the case given. A `channels { }` block *is* the grammar's numbering, so in a delegate declaring `channels { Y }` the ids `2` and `Y` name the same channel: `channel(2)` and `channel(Y)` are indistinguishable because there is nothing to distinguish. Remapping that `2` onto the composite's `Y` preserves the meaning. The remap only touches ids inside the delegate's declared range, which is exactly the range where the two coincide; outside it the literal is left alone, as `merge_grammars tolerates a numeric channel id with no channels block` already pins. Two delegates and their channels were untested, though, so the suggested regression test landed anyway: distinct names take distinct slots (2/3/4), and two delegates naming one channel share one.

`wado test package-gale`: 2964 passed, 0 failed, 20 todo.

---

gfx:
`Test (Gale O3)` was red on `cb0d9ab` and `e52575f`. It was a compiler bug, not a Gale one, and it is fixed in `8df91a1`.

**What it was.** Variant-return SROA flattens a `Result` return to `[tag, payload slots…]` and re-mints each match arm's payload binding as a `let`. It decided whether the binding's local could hold the slot read by comparing the local's declared type against the *pattern's*. `lower::plan::boxing` rewrites both when the binding is address-taken — `if let Err(err) = &result` with `${err:?}` promotes `err`'s local to `Box<AssembleError>` and retypes the pattern to match — so the two agree while saying nothing about the value arriving, and the raw payload landed in a `Box` local: `type mismatch: expected (ref null $type), found (ref $type)` at codegen. The composite corpus is what reached the shape; a 35-line program reproduces it at `-O1` and up.

**The fix.** The question is what the slot read produces, so `Rebind::rebound` takes that type. `Layout::payload_read_type` names it, `arm_is_one_level` asks with it, and `bind_payload_before` reads it off the emitted node under a `debug_assert` tying the two together. The pattern's `type_id` reaches neither path now. With the real type in hand the `Boxed` arm fires and wraps the read as lowering did, so the site is still rewritten rather than declined.

Regression: `wado-compiler/tests/fixtures/bug_sroa_variant_return_boxed_payload_binding.wado`, exercising both error cases at runtime. Red without the fix (ICE at `-O2`), green with it at `-O0`/`-O2`/`-O3`. `mise run test` is green: 39 suites, 0 failures.

---

On the second review round:

**Fixed — `AGENTS.md` dangling referent.** "the action-body one below" pointed at nothing in that file. It names the exception directly now (`cd8d9a2`).

**Skipped — reject reachable import cycles.** A reachable cycle terminating on the visited set is ANTLR4's depth-first walk, and `src/ir_test.wado:300` pins it deliberately: "A cycle the master does reach is not an error: the walk cuts it on the second visit." Neither `import.md` nor `vendor/antlr4/doc/grammars.md` requires cyclic inputs to fail, and the suggested change would fail that test. An *unreachable* cycle is already an error, two lines above.

**Skipped — assert `result.ok()` in Stage C.** Two reasons. Stage A claim (b) already asserts `result.ok()` on the same input for the same descriptor, so a recovered parse whose emit buffer matched could not pass unnoticed. And Stage C covers `ParserErrors` and `LexerErrors`, whose parses are meant to fail while still emitting output — asserting `ok()` there would fail them.

**Skipped — numeric channel ids, re-raised.** Answered above: a `channels { }` block is the grammar's own numbering, so in a delegate declaring `channels { Y }` the ids `2` and `Y` name one channel and the rebase preserves that. The remap only touches ids inside the declared range, which is exactly where the two coincide.

---

PR #1984, opened 2026-09-06
feat: gate inline `crate::` / `super::` paths, and widen Gale's static prediction

gfx:
## The rule

A `crate::` or `super::` path belongs in a `use` item at the top of the module. Where the item is read, write its name; alias a collision (`use crate::ast::Ty as AstTy;`) rather than spelling the path out. `AGENTS.md` and the `rust` skill state it.

Clippy cannot gate it. `absolute_paths` sees only the `crate::` half of the rule, and it flags paths into other crates unless every dependency is named in an allow-list nothing derives. So the gate is `scripts/rust-inline-paths.mjs`: it blanks comments and string / raw-string / char literals while preserving every byte offset, telling a char literal from a lifetime, then reports each `crate::` / `super::` outside a `use` item. `$crate` is macro hygiene and `pub(crate)` is a visibility, so neither is a hit.

`scripts/rust-inline-paths.json` records how many each file still carries — 4,915 across 266 files — and `--check` fails only on a file that exceeds its own entry. The ratchet lets the corpus migrate file by file while CI stays green.

- `mise run check-rust-paths` runs in the CI `check` job.
- `mise run update-rust-paths-baseline` runs in `on-task-done`, and only ever descends.
- A `PreToolUse` hook refuses an edit that writes one inline, naming the line and the fix.
- `mise run test-hooks` covers the scanner and the hook.

This is a stepping stone. The rule is meant to be enforced by a Rust lint written in Wado, on Gale over `package-gale/tests/grammars/RustParser.g4`, which turns the gate into a product with auto-fix. The rest of the branch is the prediction work that stands between Gale and this repository's Rust.

## Gale prediction

Six static fixes and one routed to the lexer ATN. Each has a fixture holding a hit case and a rejection case, as `package-gale/AGENTS.md` § *Failed approaches* requires of any prediction repair.

- **A greedy optional's entry signature reaches past its fixed-token prefix** into the first set of the first non-fixed, non-nullable element (`ll_opt_prefix_ruleref.g4`).
- **A caller loop's iteration prefix reaches the callee as its FOLLOW.** `FollowArg::AltMask(id)` says the continuation is the mask *or* what the current rule was called with, since a loop offers another iteration as an alternative to what follows it, not as a prefix (`ll_opt_tail_caller_loop.g4`, `ll_loop_yields_closer.g4`).
- **A greedy loop body's tail call carries the loop's own iteration prefix**, which the body's element list does not mention. Parse side and scan side are retargeted together, so the scan cannot report an end the parse walks past.
- **An LR overlap sub-dispatch checks a second token only where it separates** the candidates. Where two candidates admit the same token second, the fallback suffix scan decides, so `e DOT ID LPAREN RPAREN` no longer swallows every `e DOT ID` (`lr_suffix_prefix_overlap.g4`).
- **A multi-alt `RuleRef` contributes its first set as one exact position, then halts.** Past the head the alts diverge and a per-depth union would admit alt-mixing inputs; a nullable rule contributes nothing, since the position may be the next element's.
- **The lexer simulator takes a greedy rule's longest match**, and the earliest accept only where a non-greedy subrule makes that the answer. `scan_undecidable_arms` routes a greedy loop whose body arms overlap and reach different lengths — `STR : '"' (~["] | ESC)* '"'` — to the lexer ATN, because no static policy over the arms decides it (`lexer_loop_arm_longest.g4`).

## A continuation that runs past its group

`compute_follow_arg` read the element list it was given, and inside a group that list is the group's own alternative — so a position there was told its continuation ended at the group. `block : (OR names? OR) body` never mentioned `body`; that is `RustParser.g4`'s `closureExpression`, and it is why `g(|e| e)` did not parse.

`lower_group_alt` walks a group alternative on both sides at once and carries the enclosing alternative's remaining elements. It replaces the two near-identical walks that were in `lower_single_alt_group` and `lower_general_group`.

Three defects that change exposed, each fixed at its own level:

- **A suffix that can vanish entirely is no continuation to gate on.** Yielding on its first token hands back one the caller may take none of — otherwise `htmlElement` reads as followed by `htmlChardata?` and nothing else, and `<p>` gives its text to the loop that called it.
- **A prefix composes with the caller's continuation only where it spans the suffix.** The forwarded half lands one position past the prefix, so a prefix halted by a multi-token element asserts the caller's next token at a depth the input does not have it — `if EXPR blockExpression`.
- **An alternation boundary is only placeable at the call that wrote it.** It marks a reading that begins where the alternation is written; composing a prefix in front moves the branches without moving the mark. `let [{ x, y }] = p` stopped parsing because the field list revived the enclosing list's comma reading one position early and handed over its own separator. `ll_nested_comma_loops.g4` pins it, with the rejection case and the two levels of nesting it takes to show.

That last one was caught by `check-grammar` over the 2,901-file Wado corpus, which the Gale suite does not reach.

Six of these cases are pinned against the published jar rather than hand-written trees. One of them corrected a fixture instead of the compiler: `ll_nullable_suffix` asserted `(r (a X) (b Y))` from reasoning, and the jar prints `(r (a X Y) <EOF>)` — both readings complete, so `Y?` is a greedy subrule over an ambiguity and takes the token.

## What the FOLLOW gate costs

The gate threads a continuation through every rule that reads one, so its representation is a per-frame cost on every grammar that has a single caller-FOLLOW repair anywhere.

**A continuation is one `i64`.** The gate reads at most `FOLLOW_MAX_DEPTH` positions per reading, and a position is an index into `FOLLOW_SETS`, so seven slots of eight bits hold the whole thing — a 1-based position id and a bit saying it opens a fresh reading. Composition is `MASK | follow << <positions * 8>`, whose shift is settled at generation:

```wado
pos = scan_name(tokens, pos, (FOLLOW_MASK_2 | (follow | FOLLOW_OPENS_BIT) << 16) & FOLLOW_CONT_MASK);
```

**Only the rules that read it carry it.** `emit_follow` was one flag for the whole grammar. Asked per rule it is a least fixpoint: the rules that gate or probe seed it, and a call handing its own continuation to one of them is an *edge*, because such a call spells `follow…` only where the callee declares it. In the generated SQLite parser that took 317 of 510 functions down to 220.

**`FOLLOW_SETS` is a bitset**, built once from the emitted `FOLLOW_KINDS`; SQLite has three position sets of 130+ kinds each, which the gate used to scan linearly.

Measured with `benchmark/ab.ts`, two runs per arm, interleaved so a drift in host speed lands on both. The Rust row is the control — a moving control means the Wado row cannot be read.

```
+0.5%  noise   2.348 -> 2.337 ms   [2.348-2.533] [2.337-2.454]   Wado (Gale)
+0.8%  noise   2.069 -> 2.052 ms                                 Rust sqlparser-rs (control)
```

Indistinguishable from `main`.

## Tools

`package-gale/tools/rust_corpus_check.wado` parses a list of `.rs` files with the generated Rust parser and prints one `ok` / `ng` line each, carrying the first diagnostic's **rule stack**. Recovery restarts at an item boundary, so nearly every failure reports the same message and only the stack names the rule. Repository-wide it is the one number that says whether a prediction change helped.

## LR dispatch

`LrOverlapSite` names how a site reads the next two tokens, how it calls a suffix scan, and which emitter its branches commit to, so `emit_lr_group_chain` and `emit_lr_overlap` in `parser_gen.wado` emit both the scan twin and the parse loop.

https://claude.ai/code/session_017TqdCVVpMDaG2UduERKrRf


gfx:
Review answered. Every finding was reproduced against the current code before anything changed; the skips carry the same evidence as the fixes.

## Fixed

| Finding | Reproduced as | Commit |
|---|---|---|
| Hook scanned a replacement as if it were a file | renaming inside the `use` that imports it was refused | `25a68af` |
| `use` capture-list lookahead capped at 5 chars | `impl Sized + use     <'a> { crate::g() }` reported no hit | `25a68af` |
| ASCII identifier boundaries | `αcrate::T` a false hit, and `let αuse = crate::g();` **lost** a real one | `b422ae6` |
| `--update` could raise the baseline | `on-task-done` runs it after edits, licensing what `--check` refuses | `b422ae6` |
| `$use` read as an import | `macro_rules! m { ($use:ident) => { crate::g() } }` reported no hit | `182e580` |
| `find_fragment` in the `longest` decision | `axx` → "missing B" | `260daa8` |
| `is_nullable` in the optional entry signature | `a:` → "extraneous input `:`" | `260daa8` |
| `lr_second_token_discriminates` on an unmeasurable peer | by inspection; it now reads the first sets | `260daa8` |
| `--paths-from` read failure counted as clean | — | `260daa8` |

Two went further than suggested, because the suggestion had its own defect:

- **The hook** — reconstructing the post-edit file and scanning it would refuse *every* edit to the 267 files the baseline already covers. It reconstructs the file **and** compares against what the file held before, which is the per-file ratchet `--check` applies.
- **`find_lexer_rule`** — the same asymmetry exists in `latn.wado`'s `scan_undecidable_arms`, which returned `false` for every `RuleRef`. Fixing one site would have left the class in the tree, so both walk the graph the ATN is built from, under a cycle guard.

New fixtures, each carrying a rejection case as `AGENTS.md` § *Failed approaches* requires: `lexer_ref_rule_non_greedy.g4`, `ll_opt_sig_nullable_ruleref.g4`. Suite: 1123 files, 2995 tests, 0 failed.

## Not fixed

**`greedy_loop_prefix` should require `consumed >= body_elements.len()`** — wrong, and I have the failure to show for it. I applied it on the argument alone and it broke three cases in `driver_cst_ll_opt_tail_caller_loop_test`: `a::b` began reporting "extraneous input `b`", which is exactly what `AltMask` exists to repair. A `(SEP seg)` body's second element is the rule the loop re-enters, so demanding the walk span the body rejects every loop of that shape. Reverted, with the fixture named in the comment so it is not re-applied. The existing `prefix.len() < 2` guard is the ≥2-position evidence soundness invariant 1 asks for.

**`latn_closure` must clear `open` before resuming at `ret`** — not reproducible. `"a\"b\"c"` lexes as one token. A closure consumes nothing and `ESC` is not nullable, so two escaped quotes are two independent closures at two input positions and never share an `open`. The case is now a permanent test in `driver_cst_lexer_loop_arm_longest_test` rather than an argument.

**Hook should compare occurrences, not counts** — deliberate. The hook mirrors `--check`'s per-file ratchet exactly; making it stricter would refuse edits CI accepts, and an edit-time guard that disagrees with the gate is worse than one that is permissive.

**Share `RustBase` / `RustPBase` between the corpus tool and the driver test** — declined. It would make `tools/` depend on `tests/`, and the ports are twelve lines of test-double that must track `RustLexerBase` rather than each other.

**Bind each LR fallback scan after its precedence gate** — real, and out of scope here. It is a performance change to emit structure on a PR already carrying a benchmark regression I am fixing; it belongs in its own change where it can be measured.

**Open a tracking issue for the grouped-follow gap** — this repo's convention is that the `#[TODO]` tests are the record. The three in `driver_cst_ll_group_tail_outer_suffix_test` name the cause, the shape, and the `RustParser.g4` rule it costs.

## Still open

`Benchmark` is red and it is this PR's. `compute_follow_arg` returning `Combined` on the non-nullable path turned ~197 call sites into per-call continuation composition, and value semantics deep-copied SQLite's 130-token position sets at each one. `528a5b6` makes a continuation a flat `List<i32>` of indices into one `FOLLOW_SETS` table: **1.38 → 3.51 MB/s** on `sqlite_parse (-O2)`, against a 5.47 base on the same machine. That is 1.56×, still over the 1.50 threshold, so the next step is narrowing where the composition is emitted at all rather than making it cheaper. Not merge-ready until that is under the gate.

---

gfx:
Correcting my previous comment: **`Benchmark` passes.** All 19 checks are green on `260daa88c`.

That comment's "Still open" section said the check was red and the PR not merge-ready. `528a5b6` — a continuation as a flat `List<i32>` of indices into one `FOLLOW_SETS` table — was enough to clear the threshold on CI's hardware. My local measurement (3.51 MB/s against a 5.47 base, 1.56×) is more pessimistic than CI's own before/after comparison, and CI's is what the gate reads.

One measurement worth recording while it is fresh, since it is a real cost rather than a blocker. `compute_follow_arg` returning `Mask` instead of `Combined` for a **non-nullable** local suffix, measured on `sqlite_parse (-O2)` against a 5.47 base:

| | MB/s | vs base | corpus /492 | `\|a\|b\| c` | `\|a\| b` in a group |
|---|---|---|---|---|---|
| `Combined` (this branch) | 3.51 | 1.56× | 448 | passes | `#[TODO]` |
| `Mask` | 5.08 | 1.08× | 449 | fails | passes |

The two spellings of the closure-parameter shape want opposite answers, and the corpus barely moves either way, so this is a trade rather than a fix — left as it is, and noted here so the next person reaching for it has the numbers instead of the intuition.

---

gfx:
## The FOLLOW gate: both halves, measured

### Correctness

`compute_follow_arg` read the element list it was given, and inside a group that list is the group's own alternative — so a position there was told its continuation ended at the group. `(OR names? OR) body` never mentioned `body`; that is `RustParser.g4`'s `closureExpression` shape. `lower_group_alt` now walks a group alternative on both sides at once and carries the enclosing alternative's remaining elements, and it replaces the two near-identical walks that were in `lower_single_alt_group` and `lower_general_group`.

Three defects surfaced by that change, each fixed at its own level rather than at the site:

- A suffix that can vanish entirely is no continuation to gate on — yielding on its first token hands back one the caller may take none of. Otherwise `htmlElement` reads as followed by `htmlChardata?` and nothing else, and `<p>` gives its text to the loop that called it.
- `sequenced_follow_arg` composed even where the mask stopped short of the suffix. The forwarded half lands one position past the mask, so a mask halted by a multi-token element asserts the caller's next token at a depth the input does not have it — `if EXPR blockExpression`.
- The optional-head guard now reads the extended suffix, keeping ANTLR4's greedy-innermost answer for nested subrules.

Four `#[TODO]`s resolve (three in `ll_group_tail_outer_suffix`, one in `open_ended_nullable`) and none is added.

`ll_nullable_suffix` looked at first like a policy that had to lose — its call site is the same static evidence as `ll_greedy_optional_cross_rule` with the opposite expected answer. The jar says otherwise: `antlr4-oracle.sh ll_nullable_suffix.g4 r` on `X Y` prints `(r (a X Y) <EOF>)`. The fixture asserted its expectation from reasoning, not from the oracle. Both readings complete, so `Y?` is a greedy subrule over an ambiguity and takes the token, exactly as the `else` case does. Six cases are now pinned against the jar and all six match.

### Speed

Measured with `benchmark/ab.ts`, two runs per arm, `sqlite-parse`. The verdict is range overlap, and the Rust row is the control — a moving control means the Wado row cannot be read.

| stage | vs `main` |
|---|---|
| group extension + gate early-out + FOLLOW bitsets | **−14.2%**, ranges disjoint |
| \+ FOLLOW slots only where a rule reads them | **−10.6%**, ranges disjoint |
| \+ continuation packed into one `i64` | **noise** — `[2.375-2.411]` vs `[2.401-2.431]` |

Two structural changes got there:

**Per-rule plumbing.** `emit_follow` was one flag for the whole grammar: a single gate anywhere gave every parse and scan function five extra parameters. The same question per rule is a least fixpoint — the rules that gate or probe seed it, and a call handing its own continuation to one of them is an *edge*, because such a call spells `follow0…` only where the callee declares the slots. In SQLite that took 317 of 510 functions down to 220.

**One `i64`.** The gate reads at most `FOLLOW_MAX_DEPTH` positions per reading, and a position is an index into `FOLLOW_SETS`, so a whole continuation fits in seven 8-bit slots — a 1-based position id plus a bit saying it opens a fresh reading. Composition becomes `MASK | follow << <positions * 8>`, whose shift is settled at generation time:

```wado
pos = scan_name(tokens, pos, (FOLLOW_MASK_2 | (follow | FOLLOW_OPENS_BIT) << 16) & FOLLOW_CONT_MASK);
```

Nothing is allocated, copied, or spread across slots; `EMPTY_FOLLOW` and the slot walk are gone.

Full suite: 1123 files compiled, 2999 tests passed, 0 failed.

---

gfx:
## Two red checks, one of them mine

### `check-grammar` — a real regression in this branch

0 gaps at the merge base, 5 with this branch's `package-gale`. The Gale suite reaches 1127 files; `check-grammar` parses 2901, and that is where it showed.

Minimal case: `let [{ x, y }] = p` — `let [{ x }] = p` is fine, and so is the paren form.

```
patternElements → pattern:          AltMask   → M41 | (follow|OPENS) << 16
patternPrimary  → patternFieldList: Combined  → M8  | that << 8
```

`AltMask` marks the forwarded continuation as opening a fresh reading, and the gate restarts its depth there. That is right where the alternation is written — both branches begin at that position — but composing `}` in front moves the branches without moving the mark. At the comma in `{ x, y }` the gate revived the loop-iteration reading one position early, matched `, y`, and handed the field list's own separator to the enclosing `[…]`.

**An alternation boundary is only placeable at the call that wrote it.** `follow_composed` now masks the forwarded half down to its positions. Dropping a reading only ever costs a yield, which is the greedy answer this repair started from — the safe direction.

`ll_nested_comma_loops` pins it with its rejection case. One level of nesting does not reproduce it (the revived reading dies on the token after the comma), so the grammar carries the bar loop `Wado.g4` has between its two comma loops.

### `check-rust-paths` — not mine, but my gate

PR jobs run the branch merged with `main`, and `main` had added inline paths after the baseline snapshot. Neither side fails alone. Imported them with `use` as the gate asks (`harvest.rs`, `inline.rs`, `parser.rs`); baseline 4925 → 4915.

### Review findings

All three that were still open verified as real:

- `rust_corpus_check` rendered the rule stack innermost-first with its own separator, against the contract `runtime_test` pins and the `rule_path` that already existed. The rendering is one `rule_path_of` now, called by `ParseError` and `Diagnostic` both.
- `--paths-from` as the last argument fell through to the `.rs` arm and reported zero files cleanly.
- `latn_closure` grew `stack` and `open` together and popped only `stack`, so a rule called twice in a row was refused the second time as recursion. `lexer_seq_nullable_ref` pins it — red without the symmetric pop.

### Speed, re-measured after the fix

Arms interleaved (base, head, base, head) so a drift in host speed lands on both. The previous attempt was invalid: its control row moved 8.8%, which by `ab.ts`'s own rule makes the Wado row unreadable.

```
+0.5%  noise   2.348 -> 2.337 ms   [2.348-2.533] [2.337-2.454]   Wado (Gale)
+0.8%  noise   2.069 -> 2.052 ms                                 Rust sqlparser-rs (control)
```

Both noise, so the host was steady and the row reads: indistinguishable from `main`.

Suite: 1127 files compiled, 3008 tests passed, 0 failed. `check-grammar`: 0 gaps over 2901 files.

---

PR #1985, opened 2026-09-06
feat(nir): write a constant aggregate back as a literal, and give every block a label

gfx:
Three changes to the NIR layer, each resting on the one before it.

## The aggregate exit

`niri` holds an aggregate value whether or not it can spell one, and
`commit_fold` now writes any of them back as the literal tree the lower phase
emits for that shape: a scalar to a pooled operand, a byte sequence to
`PackedArray`, another sequence to `ArrayLiteral`, a tuple to `TupleLiteral`,
another struct to `StructLiteral`, a variant to `VariantConstruct`. So a pure
callee that fills a `List<T>` and returns it reaches the IR as the list, and one
returning a struct as the struct literal, nested `String` and all.

Reading a literal into a `Value` drops what writing one back needs — a field's
name and declared type, a case's index and payload type. `AggregateShapes`
indexes both out of `NirPackage` per pass and rides on `ProgramFacts`, where an
absent fact costs folds rather than correctness.

The writer decides whether it changed anything, by comparing each node against
the node it would put there and reusing what already spells its part of the
value. Nothing settles that question separately, so nothing can drift from it,
and the worklist settles.

Three gates bound it. A literal is a fresh object, so the exit writes only over
a node that yields an object nothing else reaches — a call, a region, a literal,
or a local the region bound itself — and refuses a reference-shaped leaf
wherever it sits. A backing array is storage rather than a value, so it reaches
the IR as its container's field, cut to the length. And the writer charges
itself one per operand it would place, refusing past `MAX_MATERIALIZED_LEAVES`;
over the benchmark and `wasm-size` corpora every value written comes in under
16, so the ceiling is a safety valve rather than a tuning knob.

## The value graph knows what it allocated

`array_new(n)` names its allocation in the graph and records the length, and
`array_len` over it — through the borrow and the `as Array<T>` the lower phase
writes — reads that back. The map is `niri`'s own `CtfeBuiltinMap`, so both
layers answer from one source.

Around that, three facts the graph was missing:

- A local bound to an aggregate literal keeps its value across a call. The local
  is the storage, so a callee writing through a `&mut` of it changes fields,
  never which object it names, and the heap version already says the fields may
  have moved.
- A write through a borrow resolves its receiver the way a read of one does, so
  `*r.f = v` seeds the slot `x.f` reads.
- A struct-literal field built from `&x` reaches `x` the way `let r = &x` does,
  recorded only where no `Assign` in the body writes that field index.

The scoped walk takes the same per-call verdicts as the whole-body build:
withholding them does not merely forgo a forward, since a call the walk cannot
call pure bumps every escaped local's heap version and a field read then matches
no store.

Forwarding, folding and pruning run in one fixpoint, because it is pruning a
capacity guard that removes the call whose heap bump hides the next one. A run
of appends into a `String::with_capacity` buffer folds its guards together, and
`WADO_TRACE=vg_field` reports what a field read forwarded to, what a literal
seeded, and which local a `&mut` field borrows.

## Every block carries a label

`LabeledBlock` is NIR's only block expression. A block no `break` names carries
a label all the same, so every block in the IR is one a dump, a trace or a pass
can say the name of, and `wir_build` emits a Wasm `block` only for a label some
`break` targets.

A synthetic label wears `$`, which begins no Wado identifier. The source cannot
spell one whatever it tries, so the parser needs no reservation rule and `__` is
a label like any other.

What a label does not mean is the rest of the change. It is not a barrier: what
ends a fact region in `condition_implication` is a `break` into the block, so a
grouping block's facts flow like any other statement sequence. It is not a
promise that the block's value is its tail either — `Body::block_yield` answers
that once, for `condition_implication`, `licm`, `tmpl_hoist` and `niri`'s
lattice alike, and `labeled_block_fusion` refuses a block whose value falls out
of its tail, since it rebuilds one `break L:` exit at a time.

`Body::breaks_to` stops at a block that re-uses the label, so a `break` is read
against the nearest block wearing it.


gfx:
All three findings were real and are fixed in 3ef07c7dc0. Full suite green (4908 e2e), clippy clean.

### `charge_leaves` did not charge a composite's own slot — `niri/rewrite.rs`

Confirmed. The doc promised a count of "the operands it would place", and the walk billed only the scalar leaves: a `StructLiteral`, `ArrayLiteral` or `VariantConstruct` node is itself an operand in its parent and cost nothing, so the ceiling did not bound what the writer emitted.

Each value is now charged the operand it becomes, stated once at the top of the walk, with a container charged the two nodes `write_container` adds beside the struct (its length and its backing array). `charge_elements` is left holding only the "a byte array packs into one `PackedArray` however long it is" rule.

I did not take the suggested shape of charging "each non-root child operand" — the root is an operand in its own parent too, and counting from the node the writer emits rather than from the recursion's position is what makes the number mean what the doc says.

### The block yield type — `const_branch_prune.rs`

Confirmed, and the consequence is concrete: `sroa_variant_return` reads `result_type` per label and asserts the labels agree (`sroa_variant_return.rs:312-322`), then rewrites it (`:413`). Feeding it the type of the position the node sits in is feeding it the wrong contract. Adopted as proposed.

I checked the other four `ExprKind::plain_block` construction sites — `lower/translate.rs`, `field_scalarize.rs` (x2 as a `sync`/`prefix` wrapper) and `niri/rewrite.rs`'s branch splice. Each wraps an expression in a block whose value is that expression, so the block's result type is that expression's type and they are already right. C3 is the one site that rebuilds a block which already had a contract to keep.

### A sequence length from a breakable labeled block — `const_folding.rs`

Confirmed, with two holes rather than one. `StmtKind::Break { value: Some(v), .. }` did not check the label, so a trailing `break OUTER: v` was read as this block's value; and the leading-statement filter admits a `Let` whose own value contains a `break` to this label, which the tail then ignores.

Both are what `Body::block_yield` answers — the block's one value, or nothing where more than one point produces it — so `const_seq_len` asks it. That is the fifth caller; `condition_implication`, `licm`, `tmpl_hoist` and `niri`'s lattice were moved onto it earlier in the branch, and this site is the one I had judged pre-existing and skipped. It is pre-existing, and that is not a reason to leave it.

I could not construct a program that reaches the miscompile from any of the three — the shapes want an optimizer-built IR rather than something the source can spell — so no fixture went in. An assertion that holds by construction reads as coverage without being any.

gfx:
Fixed in 6571e7bd1b. Full suite green (4914 e2e), clippy clean.

### The finding is real

`borrowed_root` read the block's trailing statement, so a borrow leaving through `break LABEL: &place` was answered `NotABorrow`. That is not a lost forward — it is unsound. `record_ref_root` reads `NotABorrow` as "this binding is not a reference" and *drops* the `ref_roots` entry, and `invalidate_place` reads an absent entry as a reference bound outside the frame ("a parameter above all", as the comment there puts it). A write through the binding then lands on the handle and the referent keeps its constant.

### Not the suggested fix

`Body::block_yield` is the wrong tool here, and I did not use it. It returns `None` where more than one exit produces the value, which lands on `NotABorrow` — the unsound answer again, now reached deliberately. Falling back to the safe-looking case is not safe when the lattice's bottom is the dangerous end.

Both helpers join over every exit instead: one root where they all name it, `Unrooted` where they disagree or a rooted borrow sits beside something that is not one, and `NotABorrow` only where no exit is a borrow. That is strictly more precise than before — it is the first version to see a break-carried borrow at all — and correct where the exits differ. The `Call` arm directly beside it already joined its arguments the same way.

### The class

`Body::block_exits` now answers where a block's value comes from — what each `break LABEL:` carries, plus the trailing statement — and `block_yield` is the case of exactly one exit. Nothing reads a block's value by hand any more:

- the value graph's struct-literal peel was `block_yield` written out at length, down from 40 lines to 6;
- `tmpl_hoist`'s tail unwrap now sees the single-break block it used to walk past;
- `sroa_variant_return`'s call-site collector is left alone: it visits every node, so its sibling `Stmt` arm already takes each `break` by label, and routing the `Expr` arm through `block_yield` would count a break twice.

`Body::falls_through` holds the terminator test `labeled_block_fusion` had spelled for itself. The value graph's `block_falls_through` stays — it is a reachability walk through `If` arms and self-breaks, a different question — but its doc claimed to mirror a `const_folding` function that no longer exists, so it now says what the difference is.

### Reachability, and no fixture

A probe over the fixture corpus (2 457 programs) counts **27 579** sites where a `break` carries a rooted borrow that `borrowed_root` was answering `NotABorrow` for. The shape is not rare.

I could not construct a program where it is observable, and I am not adding a fixture that passes either way. Two attempts:

- one exit — `const_branch_prune`'s C3 rewrites `break L: v` into a tail before `const_fold` sees it, so the old code was already reading the right operand;
- two exits — passes too. `ref_roots` is the fallback for a borrow taken over an accessor result; a borrow over a field is a place the frame can spell, so `frame_place_of` resolves it before `ref_roots` is consulted.

So the remaining shape is a borrow over an accessor result carried out by a break. If that turns out to be constructible I will add it; a fixture that never takes the branch reads as coverage without being any.

gfx:
Fixed in c8c0e1d861. Full suite green (4914 e2e), clippy clean, and `redundant_bce` still eliminates what it did before.

Real, and I had seen it. Reviewing the first round I noticed that `fresh_region`'s `Stmt` arm matched only `StmtKind::Loop` while its `Expr` arm handled labeled blocks, and I let it stand as pre-existing. `wado-compiler`'s rules say a pre-existing defect is fixed when found, so that was the wrong call, and the finding is the same defect a round later.

Both arms now ask the same question of the same predicate — a loop, or a block some `break` names:

```rust
NodeRef::Stmt(s) => match &engine.body.stmts[s].kind {
    StmtKind::Loop { .. } => true,
    StmtKind::LabeledBlock { block, label, .. } => {
        engine.body.breaks_to(NodeRef::Block(*block), label)
    }
    _ => false,
},
```

A block nothing breaks to stays a plain sequence, so facts still flow through the grouping blocks every index expression expands into. The `Expr` arm asked this through `unbroken_block(e).is_none()`; it asks `breaks_to` directly now, so the two read as one rule.

No fixture. I wrote one — an indexed read inside a statement block establishing `i < len`, a `break` skipping it, and a second read after the block that has to keep its guard — and it traps correctly with and without the fix. A fixture that passes either way reads as coverage without being any. Both `redundant_bce` tests still pass, which is the check that this costs no elimination it used to make.

Issue #1998, opened 2026-09-07
A binding read through a reference-typed field keeps sharing the referent, so a later write to it is visible


gfx:
```wado
struct Cell { r: &mut List<i32> }

export fn run() with Stdout {
    let b: List<i32> = [1, 2];
    let mut xs = builtin::black_box(b);
    let c = Cell { r: &mut xs };
    let v = *c.r;
    xs.push(999);
    println(`${v:?} ${xs:?}`);   // [1, 2, 999] [1, 2, 999], want [1, 2] [1, 2, 999]
}
```

`v` is bound from `*c.r`, which lands on `xs`. Value semantics make `v` an independent value, so the later `xs.push(999)` must not be visible through it. Reproduced at `-O0` and `-O2`, and the same through a function (`fn peel(c: &Cell) -> List<i32> { return *c.r; }`, then `let v = peel(&c);`).

### Where it is not

A `/code-review` run attributed this to `is_owned_value`'s `Deref` arm (`wado-compiler/src/lower/plan/value_copy/analyze.rs:312`), which inherits freshness through the operand and so judges `*c.r` fresh when `c` is. That reading is right about the arm being over-broad, but narrowing it does **not** fix either repro — I tried it (allow the inherit only through an explicit `&`/`&mut` of an owned value, i.e. `*&v`) and both still print `[1, 2, 999]`. So the arm is not what suppresses the copy here, and the change was reverted rather than committed as an unverified fix.

The freshness verdict already says "not owned" for the function form once #1999 lands: `peel`'s body returns `*c.r`, which `is_projection_of_param` does not count as a projection of `c`. The copy is dropped further down — the read-only-share / last-use side, where `c` being dead after the read is what lets the binding take the storage. `tests/fixtures/value_copy_share_root_escape.wado` pins the adjacent case (a projection of a reference root mutated through a callee or an alias); this shape escapes through a reference-typed *field* instead of an index projection.

### Why it matters now

Nothing in the language had a reference-typed struct field until tagged templates: `hole_field_type` holds each hole as `&V`, so `$hole_get`'s `*v.h0` is the first synthesized function of this shape. User code can declare one too — `struct Cell { r: &mut List<i32> }` compiles — so the shape is reachable from source, not only from synthesis.

Worth locating the real suppression point before patching; the fix belongs wherever the share is granted, not at the freshness arm.

Issue #1997, opened 2026-09-07
A `&mut self` call on a non-place receiver writes through to whatever the receiver aliases


gfx:
`convert_receiver_arg` (`wado-compiler/src/lower/translate.rs:2289`) never calls `should_wrap_value_copy`:

```rust
fn convert_receiver_arg(&self, receiver: &TirExpr, is_mut: bool) -> ArenaCallArg {
    ArenaCallArg { expr: self.convert_operand(receiver), is_mut }
}
```

Its doc gives the reason, and the reason is right for a *place* receiver: copying `builder.push_str(...)`'s receiver would append to a throwaway. But the rule is applied to every receiver, including ones that are not places, and a call result that aliases its argument is one:

```wado
struct Holder { xs: List<i32> }
trait Bump { fn bump(&mut self); }
impl Bump for List<i32> { fn bump(&mut self) { self.push(999); } }

fn get_field(h: &Holder) -> List<i32> { return h.xs; }

export fn run() with Stdout {
    let a: List<i32> = [1, 2];
    let h = Holder { xs: builtin::black_box(a) };
    get_field(&h).bump();
    println(`${h.xs:?}`);   // [1, 2, 999], want [1, 2]
}
```

Reproduced at `-O0` and `-O2`.

`get_field(&h)` is a value the caller cannot name again. Mutating it must not reach `h`. Bound first it behaves correctly — `let mut v = get_field(&h); v.bump();` copies — so the defect is exactly the receiver position.

This is the same class as the `StructField::get` / `Hole::get` / `VariantCase::extract` leak fixed in #1999, which covers a `let` binding. `tests/fixtures/reflect_member_read_copies.wado` would leak again if its `let mut v = f.get(t); v.bump();` were written `f.get(t).bump();`.

Fix direction: keep the no-copy rule for a place receiver, and put a non-place receiver of a `&mut self` call through the same `should_wrap_value_copy` every other by-value argument takes. That changes which receivers the lowering materializes, so it wants `mise run test-wado` behind it, and a fixture per shape (call result, block value, match value).

PR #1999, opened 2026-09-07
fix(value-copy): a member read hands out the storage of the argument it names

gfx:
A reflect member read hands out the container's storage, so binding its result and mutating the binding copies, as every other read does. `StructField::get`, `Hole::get` and `VariantCase::extract` each bridge through a bodyless `builtin::` marker, which is what puts the question to the value-copy plan.

Underneath, where a call's result comes from is one fact: a parameter, plus a projection into it. A body is read for that fact; a `core:builtin` declares it.

## The parameter is part of the path

`place::ReturnPath` records which parameter its projection starts at. An accessor need not hand out its receiver — `StructField::get(&self, v: &T) -> F` returns a field of `v` — and a path read as the receiver's is no path at all, leaving the call a place the walk cannot name and every caller a defensive copy. `core:serde` walks that path once per field, so a struct serializer copied every field of every value it wrote.

`is_projection_of_param`, `is_receiver_projection` and the caller-side application of a callee's path all follow the recorded parameter.

## What a builtin declares

`#[returns(owned)]` says the result is a fresh place. `#[returns(part_of(p))]` says it names a component of parameter `p`, so it lives as long as that argument's storage and a place walk projects one selector further in.

The declaration is mandatory. Only a reference argument can carry storage out, since a by-value one is already deep-copied at the call, so the obligation is checkable from the signature alone: link asserts that a bodyless `core:builtin` reading through a reference and returning storage carries a `#[returns(...)]`. Silence therefore means "allocates" rather than "guess", and a builtin whose storage behaviour is known to its author cannot reach the plan phase without saying so. `core:builtin` is compiler-owned, which is why this is an assertion rather than a diagnostic.

Link also snapshots the declarations into `FlatPackage::builtin_return_conventions`: monomorphization drops the generic and materializes an instance only for the few a later phase rewrites, so the plan phase has none left to read.

The eight declarations this covers are the whole population — `array_get_value`, `array_get_ref`, `array_get_ref_mut`, `struct_field_get`, `hole_get` and `variant_case_extract` name their argument; `array_clone` and `array_clone_prefix` allocate.

## Attribute arguments

`AttrArg::Call` carries an applied name and its identifiers, so `#[returns(part_of(arr))]` names a parameter. The parser, the unparser and `AttrArg::as_str` handle it alongside the other shapes, and `Wado.g4` accepts it — the corpus check reads that grammar as a second parser over the same sources, so a shape only the Rust parser takes is a grammar gap.

## Elsewhere

`ModuleSource::is_builtin` is the one spelling of "compiler intrinsic rather than compiled source", called by the lowering plan, the optimizer passes and NIR alike.

The NIR and optimizer files in the diff name the types they read in a `use` item instead of spelling a path inline. Behaviour is unchanged; `scripts/rust-inline-paths.json` records the lower budget.

Fixtures: `reflect_member_read_copies.wado` pins that a written binding of a reflect member read copies; `reflect_member_read_shares.wado` pins that a read-only one does not; `builtin_returns_owned.wado` pins `#[returns(owned)]` with a `wir_not_expect:O0` that fires the moment the attribute stops being read.

Two defects in the same subsystem are reported rather than fixed here, both reproduced at `-O0` and `-O2`: #1997 (a `&mut self` call on a non-place receiver) and #1998 (a binding read through a reference-typed field). Neither is reachable from the reads this branch covers.

`confine::raise_builtin_sides` still infers which by-value argument a builtin stores into a `&mut` one rather than reading a declaration. Unlike the result's origin, that has no signature-level predicate to check an omission against, so it cannot be made mandatory the same way; it is a known gap in WEP 2026-05-21.

https://claude.ai/code/session_01PxZeHqUjKrs5JFGiS2YC4m


gfx:
`Check (deps + wasm32 + hooks)` failed at `mise run check-rust-paths`, and it is not this PR's failure — `origin/main` fails the same check, with the same six files and the same counts:

```
$ git worktree add -d /tmp/wt origin/main && cd /tmp/wt
$ node scripts/rust-inline-paths.mjs --check
error: inline `crate::` / `super::` paths added; import them with `use`:
  wado-compiler/src/nir_engine.rs: 42 -> 49
  wado-compiler/src/nir_value_graph/builder.rs: 83 -> 98
  wado-compiler/src/niri/shapes.rs: 0 -> 1
  wado-compiler/src/optimize/copy_prop.rs: 7 -> 9
  wado-compiler/src/optimize/sroa.rs: 14 -> 16
  wado-compiler/src/optimize/store_load_forward.rs: 16 -> 18
```

The gate landed in #1984 while #1985 added paths to those files; the two merged without either seeing the other, so `scripts/rust-inline-paths.json` never recorded them. Every PR opened since inherits the failure, and no fix PR is open.

Fixed here rather than waited on, so this PR goes green and `main` is repaired when it merges (`0a7c5dd`). The baseline only ever ratchets down — `--update` refuses a file that grew, and says why: "Recording a file that grew would license the paths `--check` is there to refuse" — so importing the paths is the one remedy the gate admits. `ValueId`, `ValueKind`, `FuncId`, `NirUnaryOp`, `TypeTable` and `IndexSet` move into `use` items in the six files, and the census is re-recorded 95 lower.

No behaviour changes: the imports name the same items the code already spelled. `cargo check --workspace`, `mise run clippy` and `mise run test` are green locally.


---

gfx:
Both CodeRabbit findings, verified.

**`builtin_conventions` doc comment** — real, fixed in 8ac548e. The field carries `part_of` alongside `owned`, so the comment now names where a result comes from rather than one of the two arms.

**`part_of` naming an unresolved parameter** — not silent, so no assertion added. The name failing to resolve leaves `declared_return_convention` as `None`, and the mandatory-declaration assert in `link.rs` fires on the next build:

```
builtin `variant_case_extract` reads through a reference and returns storage:
declare #[returns(part_of(p))] or #[returns(owned)]
```

That assert exists for exactly this condition and names the builtin. A `debug_assert!` in `reify` would check the same thing again, earlier, with less to say about it.

Also in 8ac548e: `Wado.g4` accepts an attribute argument that applies a name. The corpus check reads that grammar as a second parser over the same sources, so `#[returns(part_of(arr))]` parsing only in Rust was a grammar gap (`corpus: 2910 files | grammar gaps: 0` locally now).

---

Issue #2000, opened 2026-09-08
P0: calling a `task return` export directly generates invalid core Wasm at -O0


gfx:
## Bug

Calling an `export async fn` that returns via `task return` from ordinary code lowers the call site as if it left a value on the stack. It does not — the value leaves through `task.return` — so the next operator underflows and the module fails validation.

`-O0` only. `-O1`, `-O2` and `-O3` all produce a valid module, so some pass at `-O1` happens to rewrite the bad call site rather than the unoptimized lowering being correct.

## Repro

```wado
export async fn handle(n: i32) -> Result<i32, String> {
    task return Result::<i32, String>::Ok(n);
}

test "discards the result of a task-return export" {
    handle(1);
}
```

```sh
wado compile -O0 --world test -o /tmp/m.wasm ice_min.wado
```

```
Internal compiler error: WIR pipeline generated invalid core Wasm module
Validation error: type mismatch: expected a type but nothing on stack (at offset 0x116)
Offending function: func #3 ice_min.wado/__test_0_discards_the_result_of_a_task_return_export
Failure at body offset +0x0005 (`>>` marks the failing operator):
       +0x0001  I32Const { value: 1 }
       +0x0003  Call { function_index: 2 }
    >> +0x0005  Drop
       +0x0006  End
```

The call pushes nothing and the `Drop` has nothing to drop.

## What was bisected

| variation | `-O0` |
| --- | --- |
| `handle(1);` (value discarded) | ICE |
| `let r = handle(1);` (value bound) | ICE |
| same export declared `export fn` (sync) | ok |
| `-O1` / `-O2` / `-O3`, either form | ok |

So the trigger is the direct call of a `task return` export, not the discard: binding the result fails the same way, with `LocalSet` underflowing instead of `Drop`.

## Where it surfaced

`example/http_mock.wado`, whose test calls the module's own `export async fn handle`:

```sh
wado compile -O0 --world test example/http_mock.wado
# Offending function: func #33 http_mock.wado/__test_1_the_exported_handle_routes_through_to_the_upstream
#     +0x0067  Call { function_index: 30 }
#     +0x0069  Call { function_index: 29 }
#  >> +0x006b  LocalSet { local_index: 9 }
```

[...]

PR #2001, opened 2026-09-08
fix(cli): one implementation of directory expansion, shared by test, format and query

gfx:
## What holds after this

A path argument reaches `*.wado` files through one implementation. `discover.rs` owns it: `files_in_dir` for a directory argument, `discover_tree` for a project-wide walk. A subcommand supplies only a `Filters` callback — `filters_at` naming its manifest section, or `no_filters` where a package's excludes must not apply — so the three cannot answer the same question differently.

Directory expansion roots the walk at the package enclosing the argument, then narrows back to it. `exclude` / `include` globs are written relative to a package root, so this is what makes them match as authored from any subdirectory, and what picks up the `.gitignore` files above the argument.

`wado format wado-compiler/tests` therefore formats nothing: `wado-compiler/wado.toml` declares `[format] exclude = ["tests/**"]`, and the e2e fixtures keep the hand-authored layouts that are part of the test. A directory that matches nothing is an error naming it, so an over-eager filter is distinguishable from an empty tree.

Naming a *file* still bypasses the filters. `scripts/update-golden-format-fixtures.sh` rewrites excluded fixtures that way, so the bypass is load-bearing; only directory expansion is filtered.

`wado query references` walks the same code but names no manifest section: a fixture `[test]` or `[format]` excludes is still a place a symbol is used. What prunes it is what hides a file from the repository rather than from one subcommand — `.gitignore`, submodules, dot-prefixed entries. The queried module is imported whatever the walk returns, so naming a file directly always resolves it, gitignored or not.

## Two defects the shared path exposed

A manifest root is a directory callers walk, not only a prefix they extend. `manifest::discover` returns one that opens: `Path::pop` and `Path::parent` both yield the empty path, which joins like `.` but opens like nothing. `openable_dir` is the single answer, used by manifest discovery, `load_nearest_manifest`, and `workspace_anchor`.

A symlink pointing nowhere names no file to discover, so the walker steps past it rather than refusing the whole walk — no tests run, no files formatted, no query answered, over one dead link. Only that case is swallowed: `symlink_metadata` succeeding where `metadata` failed is what separates it from a permission or a racing delete, which still surface with their path.

## Notes

`docs/wep-2026-05-02-test-discovery.md` records how `format` and `query` relate to the model it already specified for `test`. `docs/formatter.md` and the `wado-cli` skill state which arguments the filters apply to.

`wado test example` reaches all 34 files now, and one of them, `example/http_mock.wado`, fails to compile at `-O0` — a pre-existing compiler bug filed as #2000, untouched here. `mise run test-wado` pins `-O2` and so never saw it.

https://claude.ai/code/session_01TYEVhPatVP5arSkzALazwt

---
_Generated by [Claude Code](https://claude.ai/code/session_01TYEVhPatVP5arSkzALazwt)_


gfx:
All three findings verified against the code and fixed in `ab4a95a5d`, each at the class it belongs to rather than the line named.

**`query_adapter.rs:385` — real.** `p.file_name() != QUERY_ENTRY` dropped any file so named at any depth, though the synthetic entry exists only at `root/__wado_query__.wado`; a real `pkg/__wado_query__.wado` lost its references. Now compared against `root.join(QUERY_ENTRY)`, with the nested case added to `the_synthetic_entry_is_not_imported` (red before the fix, green after).

Swept `file_name()` comparisons across the tree for the same shape. The other two are name checks by intent and stay: `registry.rs` asks whether a path is named `wado.toml`, `clean.rs` matches `.worktrees` and `.git` at any depth. This was the only site where a name stood in for an identity.

**`wado-cli/AGENTS.md:39` — real.** An explicit file reaches neither entry point. It now distinguishes `files_in_dir` for a directory, `discover_tree` for a project-wide walk, and a file named directly being taken as given.

**`docs/formatter.md:34-36` — real, and it held in three more places.** The built-in `**/generated/**` and `**/build/**` skips were in the code and in no document. Every statement of the contract now names them and `[format] include`: `wado format --help` (which users read first, and which nothing flagged), `docs/formatter.md`, the `wado-cli` skill, and `wado-cli/AGENTS.md`.

**Skipped, deliberately:** the root `AGENTS.md` formatter bullet. It states one behaviour agents trip on — that `wado-compiler/tests/**` is excluded — and does not claim to state the file-selection contract, so listing the built-ins there would add noise without correcting anything.

Validated before pushing: `mise run format`, `cargo clippy -D warnings`, and a full `mise run test` (39 targets, 0 failures) on a clean run. The `--help` output was read back rather than assumed.

---

PR #2002, opened 2026-09-08
feat(value-copy): a builtin declares what it keeps and where its result comes from

gfx:
## Declarations, not inference

A bodyless `core:builtin` states two storage facts in its own signature, and the value-copy analysis reads them instead of guessing from shape:

- `#[returns(owned)]` — the result is a fresh place.
- `#[returns(part_of = p)]` — the result names a component of parameter `p`, so it lives as long as that argument's storage does.
- `with stores[p]` — the call keeps `p` beyond it. `array_set` and `array_fill` store their `value`; nothing else a builtin is handed outlives the call.

`FlatPackage::builtin_declarations` snapshots both at link, because monomorphization drops the generic bodyless declarations the plan phase would otherwise read.

### The obligation is enforced

Link asserts that a bodyless `core:builtin` reading through a reference and returning storage carries a `#[returns(...)]` row. It asks that at the declaration, where the return type is still a type parameter, so a generic reflect builtin is judged on the widest instantiation of `T` rather than on the bare parameter it reads as there. What is left undeclared cannot hand storage out at all, which makes reading it as fresh a fact rather than an assumption.

A malformed `#[returns(...)]` is a `RETURNS_ATTR` diagnostic: no argument, an unknown convention, a `part_of` naming no parameter, or one quoted where an identifier belongs. Silence means "allocates" — the reading that elides a copy — so a typo must not reach it.

### The self-projection verdict names its parameter

`returns_self_projection` records *which* parameter a function's not-owned returns project, and the caller tests that argument. `VariantCase::extract(&self, w)` returns a component of `w`, so `w` decides the call's freshness. A `core:builtin` answers the same question from `#[returns(part_of = p)]`.

## Closure `with`

`with` after a closure's parameter list, or after its return type, is always the closure's own effect and stores row. The compiler carries no such row yet and says so, and `Wado.g4` accepts the shape for the same reason.

The keyword never starts a handler-expression body in that position, so a handler reaches the body through a block or a pair of parentheses:

```wado
let f = || (with Log => &mut sink do { Log::emit(`hi`); });
let g = || { with Log => &mut sink do { Log::emit(`hi`); } };
```

The formatter keeps those parentheses. `closure_body_needs_parens` follows the body's leftmost spine, so a body starting with `with` re-parses as what was written.

## Attribute shape

`#[returns(part_of = p)]` is a `key = ident` pair, the shape `#[cm(type = "...")]` and `#[param(name = "...")]` already have. `AttrArg::KeyIdent` holds it and the unparser leaves the value unquoted. `Attribute::has_arg` asks `AttrArg::name()`, which answers with the key of a `key = ...` pair and with the word itself otherwise.

https://claude.ai/code/session_01PxZeHqUjKrs5JFGiS2YC4m

---
_Generated by [Claude Code](https://claude.ai/code/session_01PxZeHqUjKrs5JFGiS2YC4m)_


gfx:
Verified each finding against the code. Fixed in 6d9c7f80d.

## Fixed

**`unparse.rs` — the comparison-chain paren rule.** Real, and reproduced: `wado format` turned `|| (with … do { 1 }) < 2 < 3` into `|| with … do { 1 } < 2 < 3`, which then fails to parse.

The class is wider than the one construct reported. `matches_scrutinee_needs_parens` also returns false for a handler, so `|| (with … do { 1 }) matches { 1 }` lost its parens the same way. Both arms are now in `closure_body_needs_parens`, each guarded by the predicate its own unparser uses.

I enumerated `Expr` to check the class is closed rather than trusting the cases I happened to find: every other variant either leads with a keyword, delimiter or operator, or puts its leftmost operand in a postfix/cast slot where `binds_tighter_than` already parenthesizes a handler. `TaggedTemplate` and `Assign` do print a leading sub-expression bare, but neither can hold a handler there from source. So `Binary`, `Range`, `ComparisonChain` and `Matches` are the complete constructible set, and the format test now runs all four.

**`reify.rs` — surplus `#[returns]` arguments.** Real. `[arg] = attr.args.as_slice()` instead of `.first()`, with a fixture case for `#[returns(owned, bogus)]`.

**`docs/spec.md` — the return-type position.** Correct, the paragraph named only the parameter list.

**Formatter coverage for `part_of = ident`.** Added, pinning that the identifier stays unquoted while `#[param(name = "…")]` keeps its quotes.

## Skipped

**`confine.rs` — unmatched wasm-asset calls.** Not reachable. The finding needs an argument satisfying `needs_value_copy` at a wasm-asset call, and `WasmCoreValType::from_wasmparser` rejects `ValType::Ref(_)`, so those parameters are only i32/i64/f32/f64/v128. `taint_of` short-circuits on `!carries_identity`, so `raise_side` is a no-op for every such operand whichever `Kind` the callee gets — `Kind::Builtin` and `Kind::Opaque` are indistinguishable here.

## Not in this PR

**`ownership.rs` — resolving projected callees in `projection_param`.** The premise is real: `VariantCase::extract`, `Hole::get` and `StructField::get` in `prelude/traits.wado` each return a bare call to a `part_of` builtin and so get no self-projection verdict. Resolving the callee would be sound and would name the right parameter.

It is a missed optimization rather than a defect, and it changes what the verdict means across the whole corpus, so `wado-compiler/AGENTS.md` applies: measure what a conservatism buys before dropping it. That belongs in a follow-up with a golden-diff measurement, not in a PR already under review.

While sweeping for the same shape I also found `extract_inline_hint_attr`, `extract_export_name_attr` and `extract_allocator_tag_attr` reading `args.first()` and ignoring surplus arguments, with `#[inline]` additionally mapping an unknown argument to `Auto` silently. Pre-existing, and giving each a diagnostic channel is a separate change to attributes this PR does not touch.

---

gfx:
Verified each finding against the code. Four fixed, one skipped as unconstructible, one measured and deliberately not landed.

## Fixed

**Closure body parentheses** (`unparse.rs`) — real, and wider than reported. The chain case reproduced: `|| (with … do { 1 }) < 2 < 3` formatted to `|| with … do { 1 } < 2 < 3`, which then fails to parse. Probing the same question found `matches`, `?`, `=` and `+=` break identically — `TryOp`, `Assign` and `CompoundAssign` print their own leading operand bare, which the operand-slot table does not govern.

Rather than add four arms, `closure_body_needs_parens` no longer ends in a wildcard. The match is exhaustive over `Expr`, per the repo's Rust rules, so each variant is a decision and a new one will not compile until someone makes it. That wildcard is what let the class escape twice. `tests/format.rs` covers all eight body shapes.

**Surplus `#[returns]` arguments** (`reify.rs`) — real. `attr.args.first()` accepted `#[returns(owned, bogus)]` silently; now `[arg]`, with a fixture case.

**`docs/spec.md`** — the `-> Type` position now named alongside the parameter list.

**Formatter coverage for `KeyIdent`** — added: `part_of = xs` stays unquoted while `name = "build.id"` keeps its quotes.

## Skipped — the case cannot be constructed

**`confine.rs` wasm-asset confinement.** The claim needs an argument satisfying `needs_value_copy` at a wasm-asset call. `WasmCoreValType::from_wasmparser` rejects `ValType::Ref(_)`, so those parameters are only `i32`/`i64`/`f32`/`f64`/`v128`. And `taint_of` returns early on `!carries_identity`, so `raise_side` is a no-op for every such operand whichever `Kind` the callee is classified as. No defensive copy can be skipped there.

## Measured, not landed

**Resolving projected callees in `projection_param`.** The premise is right: the wrappers exist, and `VariantCase::extract`, `Hole::get` and `StructField::get` each `return builtin::<part_of builtin>(v, …)` with no verdict. I implemented the resolution — it is sound and does give `extract` the verdict "projects `v`".

It buys nothing on its own. `is_owned_value` has no `TirUnaryOp::Ref` arm, so `&fresh_local` is never owned, and the caller that would cash the verdict asks exactly that about its argument. Two measurements agree: all 1768 WIR goldens byte-identical, and a case written to exercise the new arm — `wrap(h: &Holder) -> List<i32> { return get_items(h); }` called with a fresh `Holder` — byte-identical at `-O0` too.

`wado-compiler/AGENTS.md` says to measure what a precision change buys and drop what buys nothing, so the change is reverted and the finding is recorded as a known gap in the ownership WEP with those numbers, stated as what it is: the call resolution and the `&fresh` reading have to move together, not separately.

---

PR #2003, opened 2026-09-08
feat(compiler): one resolution behind every `Type::method()` spelling

gfx:
`Type::method(…)` names one declaration, in every spelling it can be written: bare, `ns::Type::method()`, `Type::<A>::method()`, `Trait::<T>::method()`, `Self::method()`, and `T::method()` through a bound. Which spelling you reach for no longer decides whether the call compiles, which body it reaches, or whether its arity is checked.

## What resolves

**The whole argument list selects.** Two impls of one trait that agree on the first argument and differ after it are told apart by the argument that differs, and a literal anywhere in the list is shaped by the impl the selection picks.

```wado
trait Later<T> {
    fn later(tag: i32, v: T) -> i32;
}

impl Later<i32> for M { … }
impl Later<String> for M { … }

M::later(0, 5)     // the i32 impl
M::later(0, "x")   // the String impl
```

Candidates are compared by declaration, so two distinct types that print one name stay two candidates. A parameter still holding type parameters is compared by shape: `fn(T) -> i32` takes a `fn(i32) -> i32`, and `Holder<T, T>` does not take a `Holder<i32, String>`.

**Type arguments are inferred at the call** — the receiver's and the static's own alike — and a type parameter the arguments do not pin takes the default its trait declared.

```wado
Box3::<i32>::build(42)      // the method's own <T> solved from the argument
Cell::wrap(7)               // the receiver's <T> solved from the impl the argument picks
```

**Instance methods and inherited default bodies answer the type spelling.** `Type::method(&recv, …)` reaches a trait's instance method, and a block that overrides nothing still answers `Type::method()` with the trait's body — including through a namespace.

**`Trait::<T>::method(…)`** supplies `Self` for a receiver-less declaration, reaching past a variant case of the same name.

**A `type` alias or a newtype reaches its base's statics**, and does not launder a private one.

## What is reported rather than miscompiled

- Two traits declaring one method name on one receiver — for statics, instance methods, inherited default bodies, namespaced spellings and value blankets alike, naming each trait once.
- A static's arity, whether its parameter list comes from the trait, from a block that inherits the body, or through a turbofish. Declared defaults are padded first.
- An argument list no impl declares, saying what the receiver does take.
- An `impl` method that changes the trait's shape: a receiver the trait does not declare, or a different number of type parameters.
- A receiver type argument nothing pins — at the call site, not as an unimplemented trait during WIR build.

## Language rules

**An impl declares the type parameters its head uses.**

```wado
impl<T> Trait for Type<T> { … }
impl<A> Codec for Box3<A> { … }
```

**An impl restates the trait's parameter lists** — the same value and type parameters in the same order, with the defaults omitted. Defaults belong to the trait, resolve in the trait's module, and every spelling of the call fills them from there.

`docs/spec.md` and `docs/cheatsheet.md` state both.

## How

`resolve_static_callee` (`wado-compiler/src/elaborator/static_call.rs`) is the single walk behind every spelling. A call site takes its callee identity, its parameter lists and its return type from that one answer, so no two of them can describe different declarations. The rungs it walks — the receiver's own declarations, its resource statics, a trait impl declaring the method, one inheriting the trait's default body, the auto-derived `Default`, the newtype base — are each reached through the declaration rather than a bare name, which is why a namespaced or aliased receiver behaves like a locally declared one.

The design is recorded in [WEP: Static Call Resolution — One Walk, Four Answers](https://github.com/wado-lang/wado/blob/claude/trait-static-param-defaults-z0mkzx/docs/wep-2026-09-06-static-call-resolution.md), added here.

`wado-compiler/tests/fixtures/` carries one file per topic, each holding every case of that topic under one set of declarations; error fixtures assert over the full diagnostic list through `compile_errors_contains`.

https://claude.ai/code/session_01QQH85JCUDURXrKc1BED1Ka


gfx:
Answered in 2c645b126. Each finding was reproduced against the current code before anything changed; three turned out to be compiler bugs.

## Fixed

**`class_admits` compared rendered names** (`synth.rs`). Reproduced: `impl<T> Mk<fn(T) -> i32> for X` refused a matching function with *"no impl of 'Mk' for 'X' takes an argument of type `fn(i32) -> i32`; 'X::mk' is available for `fn(T) -> i32`"*. A base name embeds a function type's own parameters and a tuple's elements, so those never spell alike.

The fix is structural and reuses `infer::unify`, which already recurses through function types: solve the parameter's slots from the argument, substitute, and compare the result. That is strictly more precise than the name test — it also rejects `Holder<T, T>` against `Holder<i32, String>`, which the name test admitted. Covered by a new case in `static_overload_by_arguments.wado`.

**A type-parameter count mismatch was accepted silently** (`orchestration.rs`). Reproduced: `impl` declaring `fn boxed<T, U>` for a trait declaring `fn boxed<T>` compiled. `inherit_trait_param_defaults` then `zip`s the two lists and truncates one. The whole-program pass already compared the value list and the receiver on the same reasoning — a call is built to the trait's shape — so the type list is the same rule, not a new one; `docs/spec.md` already says the implementation restates the trait's list. `TraitMethodArityMismatch` now carries which of the two lists disagrees rather than gaining a near-duplicate variant. Covered in `error_trait_impl_redeclares.wado`.

**A trait resolved by rendered name** (`method_lookup.rs`, and the same shape in `elaborator.rs` where default-method bodies are synthesised). `wado-compiler/CLAUDE.md` requires a declaration to be identified by its `DefId`; both sites had the resolved declaration in hand. Fixing both left `trait_sig_by_name` with no callers, so it is deleted — no site can look a trait up by spelling any more.

**`build_declared_type_params` re-admitted the elision this branch bans** — it added any undeclared name found in the impl head, so `impl Tr for Box<T>` treated `T` as free. Removed; the declared list is now the only way in, matching `is_impl_target_param`.

**Stale symbols in `docs/wep-2026-07-31-overload-resolution.md`.** `conversion_preselect`, `conversion_impl_survey` and `AmbiguousConversionArgument` no longer exist — `from` is a trait static and takes the ordinary static path. The stale names are at lines 439 and 574; line 34, which the comment pointed at, names `static_arg_preselect` and `select_candidate` and is correct as written.

**The `Both::both` fixture assertion.** Real, and it predates the fixture consolidation: both blocks returned `i32` and both bodies returned `5`, so the assertion could not tell which was selected. It returns `T` now and asserts `"x"`.

## Skipped

**"Run `mise run format` after the documentation edits."** Ran it: exit 0, no changes. The tree was already formatted.

## Verification

`cargo clippy --workspace --all-targets -- -D warnings` clean, `mise run test` green across all 39 targets, and `mise run update-golden-fixtures` regenerated with one expected golden moving.

---

gfx:
Answered in `ff4b6e4c0`, `19e3da034`, `1a34e6354`.

## Fixed

**The declared-slot set read a global name union.** `build_declared_type_params` filtered its own declaration list through `known_type_names_cache`, which this file's own documentation calls a global union: `module_visible_types` sits beside it precisely because that union "is **not** polluted by type names from unrelated modules", and its example is this very question. But the filter can only ever *remove* a declared parameter, and a declared parameter is a slot however many modules name a type that — which is the rule `is_impl_target_param` already implements with no filter. So the filter is deleted rather than re-pointed at the per-module set: keeping it would mean two functions answering one question two ways.

I could not construct a program whose behaviour changes. Three probes — a single clashing parameter, two parameters with one clashing, and a clash on a name certainly in the cache — all compile and pass, because `verify_impl_type_compatibility` returns `true` on an empty set and its only callers are the arithmetic and indexing projections. No fixture was added for that reason: one passing identically before and after would read as coverage while guarding nothing.

**The rest of that class.** Two more sites asked "is this name a type parameter?" of the same global union — `is_blanket_type_param` and the blanket-name branch in `method_lookup.rs`, both deciding whether `impl<T> Trait for T` is a blanket. A type named `T` declared anywhere would have stopped it reading as one. Both ask the block's declaration list now.

**An open parameter was held to a stricter rule than a closed one.** `unify` binds `&T` against `&mut U` — its comment calls this "the one direction assignability allows" — and `ArgClass::Exact`'s fast path allows the same coercion, but my post-substitution comparison used bare `TypeId` equality. Both go through one `param_takes` now. The same function also dropped `TypePack` bindings when converting `unify`'s map to indices, though `substitute_type_params` fills packs by index too; a dropped binding leaves that slot spelled as written, so the comparison could never succeed. Currently latent — `&T` peels to a bare slot in the early return — but it stops being latent under the other finding below.

## Skipped

**Narrowing `&T` admission** (`peel_refs` in the early return). Mechanically accurate: `&T` admits every argument class. But over-admission is the direction this module is built to err in — its header says an inexact premise "widens rather than guessing" — and the cost is a spurious ambiguity report, where under-admission wrongly rejects a valid call. The other `ArgClass` arms reach `representation_head`, so `&T` parameters would start being rejected for integer, string and head-classed arguments. No defect was demonstrated, so I would want a reproduction before changing selection breadth on a green branch. Recorded here rather than silently dropped.

## Verification

`cargo clippy --workspace --all-targets -- -D warnings` clean and `mise run test` green across all 39 targets, on this exact tree. A distill pass over the branch went in with the fixes rather than after them: it caught a `class_admits` doc that still described the pre-branch behaviour, and the golden fixtures were regenerated.

---

PR #2005, opened 2026-09-08
fix(cm-binding): a Wado call of an `export async fn` gets what `task return` delivered

### LINKED ISSUE #2000: P0: calling a `task return` export directly generates invalid core Wasm at -O0

gfx:
`task return` names an `export async fn`'s result. Where that value goes depends on who entered the function: the Component Model runtime when the export binding did, the caller when Wado code calls it.

## The two entries

`task.return` hands a resource result to the host, so one body cannot both deliver and return.

`split_task_entry` clones the user's function as `__cm_task_entry__<name>`, `expand_task_returns_in_func` rewrites the copy's `task return` into the canonical calls, and the export binding calls the copy. A `FunctionRef` resolves by name, so the rename is the whole redirection, and DCE drops whichever copy nothing reaches. The CM path through that copy is the one that already shipped.

The user's own function carries its declared return type. Each `task return value` binds into an `Option` slot, and the body ends returning what it bound — trapping if it reaches the end having delivered nothing, which for the task entry is a CM protocol error either way. A bare `return` takes that same exit, so delivering an error and leaving early still fills the Wasm return. A unit result takes no slot, leaving only the operand. Neither rewrite crosses into a closure, which is a function of its own.

The delivery point stays inside the body in both copies. A binding that waited for the body to return before delivering would buffer a streamed response body in full, which is the deadlock `task return` exists to avoid.

## What it buys

A test can call an exported handler and assert on its own result. `example/http_mock.wado` checks the status and headers of the `Response` that `handle` delivered, alongside the upstream's record of what it was asked for.

`wado wit` is unaffected: `handle: async func(request: request) -> result<response, error-code>`.

Closes #2000. With the call site and the callee agreeing on the result type, the `Drop` / `LocalSet` over an empty stack has nothing to arise from. The issue reports `-O0`, but any *use* of the result fails at every optimization level; its repro discarded the value, which is why the optimizer appeared to save it.

## Fixtures

`async_export_called_directly`, `async_export_result_reaches_the_caller`, `async_export_task_return_evaluates`, `async_export_bare_return_carries_result`, `async_export_unit_result_evaluates`, `async_export_closure_return_is_its_own`, `async_export_without_delivery_traps`, `async_export_declared_result_survives`.

https://claude.ai/code/session_01TYEVhPatVP5arSkzALazwt

---
_Generated by [Claude Code](https://claude.ai/code/session_01TYEVhPatVP5arSkzALazwt)_


gfx:
Both findings verified against the code. One is skipped, one is real but predates this PR and is filed as #2006.

## `elaborator/item.rs` — path-sensitive delivery validation: skipped

The check as described rejects a correct program that is already in the corpus. `wado-compiler/tests/fixtures/http_task_return_in_let_block.wado`:

```wado
let _emitted: i32 = match selector {
    1 => {
        task return Result::<Response, ErrorCode>::Ok(response);
        1
    },
    _ => 0,
};
```

`selector` is 1, so the handler always delivers, but a path-sensitive walk sees the `_` arm reaching the end without a `task return`. "Every reachable path delivers" is false statically and true dynamically, so the check would fail a fixture that expects status 200.

Delivering nothing is handled at runtime instead: `take_task_result`'s `None` arm traps, so the Wado caller gets a trap at the cause rather than a zeroed value. A static check that admits the fixture above needs value-range reasoning over the scrutinee, which is a larger change than this PR carries.

## `synthesis/cm_binding.rs` — void async exports: real, filed as #2006

Confirmed with a `--lib` package exporting `export async fn ping()`: the component builds and contains zero `task.return` occurrences, so the host task never completes.

Not introduced here — the `if let Some(return_type) = &export.return_type` gate is identical on `main`; this PR changes what happens inside it, not the gate.

Passing the unit type and an empty flat-slot list through the same path does not close it. It reaches the import signature instead:

```
Validation error: type mismatch for export `task-return:ping` of module instantiation
  argument `wasi` expected: (func) found: (func (param i32))
```

`Package::task_return_flat_params` is one signature per world, and `record_task_return_flat_params` asserts every export flattens alike — "the shared `task_return` import cannot represent both". A void export needs `(func)` while a value-returning one needs `(func (param i32))`, so the fix is to make that import per-export, as `CanonicalIntrinsic::TaskReturn(name)` already is on the call side. That mechanism is untouched by this PR, and a speculative patch turned a silent non-completion into an ICE, so it is #2006 rather than a widening of this branch.

---

PR #2010, opened 2026-09-09
fix: the empty tuple is not unit, on either side of the compiler

gfx:
`()` and `[]` are distinct types, and the compiler treats them as two throughout: in the spelling it manufactures for its own unit, in the impls it registers, and in the boundary paths that carry neither.

## The compiler spells its own unit `()`

`type_id_to_ast_type` is the bridge from a `TypeId` back to the AST, and it renders `ResolvedType::Unit` as the named `()`. Every AST-level predicate asks `Type::is_unit`, which only that spelling answers, so the compiler's own unit reads as unit wherever it lands.

`cm_payload_type_from_ast` therefore classifies `Result<(), E>` as carrying no `Ok` payload, and its future lift materializes no slot for one:

```
__result_val = struct.new "…Result<(),…ErrorCode>" { 0 };
```

`lift.rs` and `lower.rs` each name unit among the types they load and store: it occupies no memory, so there is nothing to read from an address and nothing to write to one. Without that, the catch-all `Type::Named` arm takes `()` for a resource handle.

## An `impl` on `[]` is not an `impl` on `()`

```wado
impl Marker for [] {
    fn tag(&self) -> i32 { return 7; }
}

export fn run() {
    let u: () = ();
    let n: i32 = u.tag();   // error: no method 'tag' found on type '()'
}
```

A written `[]` keys its impl by the tuple head, as `FqTypeName::builtin`'s doc requires — "one spelling, whichever side asks". The general tuple path serves an empty element list: lengths of zero match, an empty zip satisfies `all`, and `SolverType::Tuple` and `FqTypeName::tuple` each carry an empty argument list. `resolve_type` makes `[]` an empty tuple, and the type check, method lookup and impl key now agree on that.

## No boundary path carries the empty tuple

`check_cm_boundary_representable` rejects a written `[]` at an export, and unit reaches synthesis spelled `()`. The import side follows from the same fact by another route: an import adapter runs only for a registry-known CM interface, and those come from WIT, which has no `tuple<>`.

Six paths state this with `unreachable!` against one shared message, so a path that does reach the empty tuple announces itself rather than lowering it as something it is not.

## One predicate

`Type::is_unit` is the only place the question is answered. `wit_emit` and codegen ask it rather than matching a name, and the name-to-string arms spell `()` the way every other named type is spelled.

The spec records what follows from `()` and `[]` being distinct: separate `impl`s, and no Component Model representation, so `[]` cannot cross a component boundary.

## Known gap

A `#[cm]` attribute naming an interface the registry does not know panics in WIR translation (`[WIR] unresolved Call`) rather than reporting a diagnostic. Unrelated to the empty tuple — the same panic occurs with an `i32` parameter — and found while probing the import path here.

https://claude.ai/code/session_015ePhqkD7uCobpVaFu4z1Ci

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1954,12 +1954,11 @@ let unit: () = ();    // Unit type/value
 let empty: [] = [];   // Empty tuple (rarely used)
 ```
 
-Being distinct types, they take separate `impl`s: a method defined on one is not
-found on the other.
+They take separate `impl`s, so a method defined on one is not found on the other.
 
-`[]` has no Component Model representation — a `tuple` carries at least one type,
-and `()` is the type that carries none — so it cannot cross a component boundary.
-An export naming one is rejected at compile time.
+`[]` cannot cross a component boundary. It has no Component Model representation:
+a `tuple` carries at least one type, and `()` is the type that carries none. An
+export naming one is rejected at compile time.
 
 ##### The `never` type (`!`) — bottom type
 
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -1954,6 +1954,13 @@ let unit: () = ();    // Unit type/value
 let empty: [] = [];   // Empty tuple (rarely used)
 ```
 
+Being distinct types, they take separate `impl`s: a method defined on one is not
+found on the other.
+
+`[]` has no Component Model representation — a `tuple` carries at least one type,
+and `()` is the type that carries none — so it cannot cross a component boundary.
+An export naming one is rejected at compile time.
+
 ##### The `never` type (`!`) — bottom type
 
 `never` is the bottom type: it is a subtype of every type. An expression of type `never` never returns — it always diverges (traps). `panic()` and `unreachable()` both return `!`.
--- a/wado-compiler/src/codegen/component.rs
+++ b/wado-compiler/src/codegen/component.rs
@@ -539,9 +539,7 @@ fn emit_cm_val_type(
                 None
             } else {
                 let ok = &g.args[0];
-                if let Type::Named(named) = ok
-                    && named.name == "()"
-                {
+                if ok.is_unit() {
                     None
                 } else if let Type::Named(named) = ok
                     && own_resource_type_indices.contains_key(&named.name)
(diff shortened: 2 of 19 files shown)


gfx:
Fixed in 1af9a00bf — both predicates now send `[]` to the catch-all, the parameter side through a guard on the vacuous `all` and the return side by dropping the early `true`. The comment that called the empty tuple "the unit type, which is always supported" went with it.

One correction to the finding's reachability, for the record. The panic is not reachable from a Component Model import today: `emits_function` gates on the registry, registry interfaces come from WIT, and WIT has no `tuple<>` to name. The export side is closed too — `empty_tuple_boundary` compiles `export fn f() -> []` and asserts the boundary rejects it, both directly and behind a `future` payload.

The fix is still right, and worth more than the reachability: these predicates answer whether a type has a Component Model representation, and `[]` has none. With them saying so, the `unreachable!` on `ast_type_to_cm`'s empty-tuple arm holds by construction rather than resting on an argument about what WIT can spell.

---

gfx:
Confirmed and fixed in 4501eb6de, though not the way the patch proposes.

The gap is real. On the return side `Result`, `List`, `Option` and `AsyncCall` already recurse; `Stream` and `Future` alone answer `true` without looking at the payload. On the parameter side the generic arm judges nothing but the head, so `List<[]>` and `Result<[], E>` pass as well.

What I did not take is the full recursion. The parameter predicate's `Named` arm accepts a name only if the registry sets carry it, so recursing through arguments turns away a `List<T>` whose `T` is an unregistered struct — nothing to do with the empty tuple. An unsupported signature is one `emits_function` drops, and a silently missing import is a worse failure than the one being fixed. That looseness is deliberate, and widening it is not this PR's business.

Instead `mentions_empty_tuple` recurses through generics, tuples and references, and the two arms that accept a generic without judging its arguments ask it:

```rust
) && !generic.args.iter().any(mentions_empty_tuple)
```

Only `[]` is turned away, at any depth, which is the whole of the finding.

---

PR #2011, opened 2026-09-09
perf(nir): 7 % off the parser's debug compile, and six miscompiles out of the operand bridge

gfx:
Cuts a warm `-O2` compile of the Gale-generated SQLite parser by 6.9 % and
`json_twitter`'s by 4.2 %, fixes six miscompiles in the operand bridge, and
gives the NIR optimizer WEP the measurements its open items are sized against.

## Arena walks allocate nothing per node

A buffer exists only to break `for_each_child`'s borrow, which a walk cannot
hold while it mutates. Three cases, and only the last needs one:

- A read-only walk recurses inside the callback. That includes
  `Body::walk_nodes_under`, so the arena's own traversal has no explicit stack —
  the arena is a tree, and the call stack is the traversal stack.
- `Engine`'s parent map and worklist sit beside `Body` rather than inside it, so
  `alloc_expr` / `alloc_stmt` / `alloc_pat` / `replace_expr_kind` split the
  borrow and write through the callback.
- Only `arena_opt_walk` and `rbce_walk` mutate the body as they recurse. They
  thread one `Vec`: each level appends its children at the tail, indexes them out
  by value, and truncates back, so the peak is the deepest path and the walk
  costs one allocation.

No walk buffer is thread-local. About forty hand-rolled stacks across the
optimizer, `niri` and `remarks` now go through the shared walk API, and siblings
are visited in source order, matching `for_each_child`.

## One set of rules per pure value

A promoted operand is the same computation as the skeleton node it replaced, so
it answers to the same rules. The two tiers spell that computation differently,
so a rule stated once per tier drifts. `niri`'s arithmetic folding is written
against `Operand`, over a view that resolves the tier at its leaves: short
circuits, trap-to-`NonConst`, and the cast target are stated once and reached
from either side. The `place_aliases` refusal on a local read is likewise one
rule, `local_lattice`.

Six places where a pooled operand did not get what the skeleton got:

- `scalar_forward` counted only skeleton reads, so forwarding a temp dropped a
  binding a pooled read still named.
- `copy_prop` forwards a promoted binding only when it is constant.
- `dae` asks `is_pure_nontrapping_operand_typed`: a promoted `100 / zero` is
  pure and still traps, so the parameter stays alive.
- `niri` evaluates through a promoted operand instead of stopping at the pool.
- `ValuePool::binary_folded` carries the short-circuit identities.
- A pooled `Opaque(Local)` read refuses an aliased local.

`optimize::run_pass` audits the invariant behind the first of these — every
local a promoted operand reads still has a definition — at end-of-optimize, and
at every pass boundary under `WADO_TRACE=promoted_reads`.

## The promoted-read census is maintained, not re-derived

The census is an upper bound maintained per edit: an operand becoming reachable
adds its locals and one leaving is never subtracted. Over-counting keeps a
binding alive, which costs an elision and never correctness, and the session-end
`debug_assert` checks the bound. It walks what a fresh census walks, pool
extraction sources included, so it can only over-count.

`Body::for_each_live_node_under` visits each node once, which rests on no two
promoted values naming the same extraction source and on a sourced node not
being skeleton-reachable. Both are assertions.

## `lower` interns all-literal arithmetic

Arithmetic whose leaves are all literals is interned as one pool value instead of
skeleton nodes, folded through the same helpers the optimizer uses. Sound this
early because such a value names nothing outside the pool, so no later pass can
re-contextualize it. Only a tree that folds all the way to a constant is taken:
`1 / 0` is a computation, and the node holding it has to survive for the elider
to see the trap it would erase.

## Measurement

`WADO_TRACE=arena_census` reports the node mix, arena bloat, operand-bridge ratio
and how much of the skeleton `lower` could intern with the flow resolved.
`WADO_TRACE=freeze_refusals` reports why the freeze leaves each pure kind in the
skeleton. Both cost nothing with the target off.

The WEP records what those numbers say, including the split of the code-size cost
the early freeze's local-naming refusal is holding back, and why that refusal is
not next in line.


gfx:
Both findings were real. One of them was larger than the suggested fix, and the fix I first wrote for the other was slower than the defect.

## `param_spec.rs` — the walk, not the name

The suggestion was to rename `for_each_reachable` to `for_each_skeleton_node`, on the ground that an `OpaqueSource::Expr` only ever records a constant-length `array_new`. That is true of how a source is recorded, but it does not follow that param_spec may skip one: `mark_unclassified_opaque` marks a local opaque only for nodes it actually visits, so a use the walk never reaches leaves the local looking cleanly classified. A rename would have made the name agree with a walk whose safety nobody can state.

So the walk moved instead. `Body::for_each_reachable_node` at all six sites; the wrapper is gone.

Measured before committing to it:

| | skeleton | live |
| --- | --- | --- |
| WIR goldens (1768 fixtures) | — | byte-identical |
| `sqlite_parse` `-O2` wasm | — | byte-identical |
| `wado compile -O2` (debug build) | 5.53 / 5.56 / 5.70 s | 5.53 / 5.70 / 5.75 s |

Correct and free, so there is nothing to trade.

## `nir_arena.rs` — the assertion

Real: `find_in_nodes_under` re-walked the body once per extraction source.

The suggested shape — collect the skeleton's `ExprId`s during the first walk — costs an insert per node on **every** live walk, including the bodies that have no source at all. Measured, that is worse than the defect: 5.53 s → 5.92 s, +7 % on the same compile. The set is now built lazily, in one walk, only once `sourced` is non-empty. Back-to-back pairs: 5.54–5.82 s before, 5.62–5.74 s after, and the assertion is O(body) instead of O(sources × body).

The two assertions also became one. Both stated the same invariant — the live walk covers each node once — with the same consequence, and a single "already covered" set catches a case neither caught before: a source sitting inside another source's subtree. It does not fire on the corpus.

## Also fixed

`ValuePool::for_each_opaque_expr` documented its results as "skeleton expressions", which is what the assertion in `for_each_live_node_under` says cannot happen.

## Not done here

The sweep found ~24 other call sites that walk `for_each_node_under(NodeRef::Block(body.root), …)`, and a good number of them are safety censuses where missing a node is unsound in the same direction — `clone_forward::collect_usage`, `const_object_globalization::mutated_locals`, `dae::remap_locals`, `sroa::collect_candidates` among them. Auditing and converting those changes a phase contract across the optimizer, so it is going to the author as a proposal rather than into this PR.

PR #2014, opened 2026-09-10
fix(ownership): a binding read through a reference field keeps its own copy

### LINKED ISSUE #1998: A binding read through a reference-typed field keeps sharing the referent, so a later write to it is visible

gfx:
A binding read through a reference-typed field owns what it holds: a later write to the referent, whether made at the referent itself or back through the field, is not visible in it. Nor does a read of one reference field share storage a sibling field writes. The shape is declarable from source (`struct Cell { r: &mut List<i32> }`), and tagged templates synthesize it for every hole — `hole_field_type` holds each hole as `&V`, so `$hole_get`'s `*v.h0` is a borrowed path the compiler writes for itself.

## Why the comparison needed the flag

The read-only-share rule decides whether a binding may alias the storage it was read out of, by asking whether any write can reach that storage while the binding is live. It answers by root: a write is compared against a read only when the two paths share a root local, and then by which selectors they split at.

A reference-typed field ends a path at storage its root only borrows. The root names the holder; the storage is reached under whatever name its lender goes by. So for such a path neither half of that comparison holds — the root is the wrong local, and the selectors describe the way into the holder rather than into the referent.

## What honours it

`Place::through_borrow` marks exactly this chain, set by `Resolver::names` as it walks: through a local binding, and through a callee's `ReturnPath`, which a second walk over the syntax would follow neither of. Three rules read it:

- **`share_eligible`** refuses a share whose source path is borrowed. The write that reaches its storage is rooted at whoever lent it, where the scan never looks. Two reference fields of one root may also name one storage, so splitting a read and a write at them is no disjointness either.
- **`write_cannot_reach`** answers "no" for a borrowed write, which may be the read's own storage under another name. This predicate owns the root check as well, so one function decides whether a write reaches a read and `disjoint` keeps the common root its comparison assumes.
- **`compute_ref_targets`** records no referent for a borrowed binding. Such a place names the slot holding the reference, not what it points at, so the immutable-root rule abstains rather than answering about the holder's immutability.

The Sharing section of [WEP 2026-05-21](docs/wep-2026-05-21-resource-ownership.md) states both rules. The e2e fixture pins five shapes: a read through the field, the same read behind a callee, a read split from a write at two sibling reference fields, a write made through the field, and a read off a reference bound out of one.

## Cost

`core:args`'s `next_field` binds an option token out of `self.de.tokens[...]`, where `de` is a `&mut` field, and so copies that token. Argv parsing runs once per process over a handful of tokens. No other generated output moves.

Closes #1998

https://claude.ai/code/session_01PwwBgwHZh2YV3EnQt5KYud


gfx:
CodeRabbit's one finding — `docs/wep-2026-05-21-resource-ownership.md:409-410`, "the scan does not inspect a mutation at another root" — is skipped, not fixed.

The line says the write is "rooted elsewhere, where the root-keyed scan never looks". That is a claim about the comparison, not about which mutations the loop visits: `write_cannot_reach` answers "cannot reach" for every write at another root, so a root-keyed comparison has no way to see this conflict. That is why a borrowed read must copy, and it is what the paragraph states.

Rewriting it to name `share_eligible`, `write_cannot_reach` and `path.through_borrow` would put the implementation's shape into the WEP, which `docs/AGENTS.md` keeps out of documentation. The rule itself is already in the sentence before: a path through a reference-typed field places nothing, because its root names the holder rather than the storage.

---

PR #2016, opened 2026-09-10
perf(json): 13% faster json-catalog deserialize, and byte literals that type-check like integers

gfx:
`core:json` deserializes json-catalog 13% faster, at 4.702 ms against the
5.318 ms on record, closing `vs best` from 3.23x to 2.87x. json-twitter moves
with it. A byte literal also type-checks wherever a decimal literal does:
`b'0' <= b`, `(b - b'0') as u32`, `pick(x, b'A')`, `let x: f64 = b'A'` and
`let x: i128 = b'A'` all compile, and a comparison against one still resolves
to `u8` inside an `if`, a field or an argument.

## The whitespace scan

`peek_after_whitespace` returns the byte it stops on, so a caller reading the
token's first byte pays one bounds check and load instead of two. The number
parsers take that byte as their lead; `read_number_raw` and the `parse_*`
family take it on the same terms.

Its run scan reads four bytes per bounds check. wasmtime re-derives an array's
base, its length and a null check for every `array.get` and shares that work
only across gets sitting in one block, so a byte-at-a-time loop pays the whole
preamble per byte. The width is four because eight measures slightly worse, and
the batching pays in proportion to the run it covers: two neighbouring scans
with shorter runs lose it, which `dead-ends.md` records with the run lengths
that decide it.

`is_json_ws` is one bitset test behind a `b > b' '` short-circuit that every
call site spells out. That compare settles the common case without the shift
and keeps the shift count under 64, which the function states as
`assert b < 64`.

## One shape for a numeric literal

`classify_numeric_literal` is the single enumeration of the shapes implicit
conversion reaches. Every walk that asks whether an operand takes its type from
the other one reads it: binary operands in the elaborator and in reify, range
endpoints, block-tail retargeting, and type-arg inference. A shape cannot be
coercible to one walk and already-typed to another.

`LiteralPairOrder` answers the pair the walks cannot settle from context, and
`LiteralPairOrder::resolve` runs it, taking the walk's own resolve function.
An expected type is a hint only where a literal can be that type, so a
comparison's `bool` types neither side. Failing a usable hint, a byte literal
anchors the pair, since `b'A'` carries a type where a decimal literal does not.
`range_endpoint_order` answers the same question for `start..end`, which both
walks read.

A byte literal reaches every target an integer literal does, by its decimal
spelling, and `reify_numeric_literal` is the single construction site for the
TIR either produces.

## One container-entry head

`entry_head` opens a container entry: past whitespace, stop at the closing
byte, consume the comma before every entry but the first. Five of `core:json`
and `core:json_nsd`'s six entry points call it, which measures flat.

`next_field` keeps the shape inline because it reads the byte the scan stopped
on to check the key's opening quote. Returning that byte from the helper makes
every seq element and map key re-read one after the comma, and costs 2.0% over
six alternating rounds against a flat reading for the five. `dead-ends.md`
carries the measurement.


gfx:
Verified each finding against the code before acting. Three were real and are fixed in e6f784c; one is not a regression and its suggestion is a design change I am declining; the procedure one is fixed.

**Fixed — byte literal at `i128`/`u128` (critical).** Confirmed: `let dec: i128 = 65` reifies to `i128::from_i64(65)` while `let byt: i128 = b'A'` reified to a bare `65` carrying the `i128` struct type. Fixed by having `try_reify_int128_coercion` read `classify_numeric_literal` rather than enumerating the shapes a second time — that duplicated enumeration is the class this branch exists to remove, and the suggested `Byte` arm would have extended it.

No behavioural test catches this: every byte value fits the low 64 bits, so the pipeline absorbs it and equality, shifts and `Display` all agree before and after. `tests/integration/int128_literal_coercion.rs` reads the TIR instead, which is where the difference is. It fails without the fix.

**Fixed — EOF after an object member comma (both sites).** Confirmed in `skip_object_body` and in `next_field`. The two also disagreed on the error kind, `malformed` against `unexpected_type`, for the same condition. Both now call `key_quote_error`, which reports `eof` for a negative peek and `unexpected_type` otherwise; the hot path keeps its single compare. Covered by two cases in `serde_json_error_kind.wado`, one per walk, which fail without the fix.

**Declining — range endpoint inference.** Not a regression: `origin/main` resolves `start` first and passes its type as the `end` hint for every case that is not "start is a numeric literal and end is not", which is exactly what `LeftAnchors` does. `git show origin/main:wado-compiler/src/elaborator/expr.rs` at `resolve_range` shows the same `else` branch. So the branch changes nothing here.

The narrowing itself is a language change, not a bug fix: it would stop a range endpoint taking its type from the other endpoint, which is the rule literals follow everywhere else in Wado. Keeping the existing behaviour, decided with the repository owner.

**Fixed — A/B procedure.** Correct that `cp` of the differing files leaves behind anything an arm deletes or renames. The skill now replaces the whole of `lib/` per arm from the arm's ref.

PR #2020, opened 2026-09-11
fix(variant): three wrong-code defects in variant mutation, and a dispatch fold worth 5% of the fixture corpus' WIR

gfx:
A mutation through a variant lands where the source says it lands. A `match` or `if let` binding no longer takes its payload out of a scrutinee that is still live, a `&mut self` method on a variant receiver writes back to the receiver, and a method that replaces a `variant` element of a container is refused rather than compiled into a discarded write.

A variant whose case is a compile-time fact now decides its own dispatch, which removes 35,236 lines of WIR (5.1%) from the 214 golden fixtures it touches.

## A destructured binding is left to the value-copy planner

Two analyses answer whether a read may move instead of copy, and `is_last_use_move` takes their union. The TIR-level one in `lower/plan/value_copy/last_use.rs` combines last use with ownership; the source-level one in `elaborator/liveness.rs` answers last use and knows nothing of ownership, so on a destructured binding the weaker answer decided.

A match arm, `if let`, `while let`, or `matches` binding names storage the scrutinee still owns, so handing it to a new owner is a move out of the scrutinee — which the binding's own last use does not license, and a receiver's field is never dead. `EligibilityPass` now excludes those bindings, leaving the decision to the planner, which has the ownership facts to make it.

`opt_value_copy_match_binding_live_scrutinee.wado` covers the four shapes that reach it — a `match`, an `if let`, `match self.p` through `&self`, and `match h.p` through a `&H` parameter — alongside a by-value scrutinee, where the payload may still move.

## A `&mut self` variant receiver

A `&mut self` method may replace the whole variant, and `*self = P::B(..)` changes the case: no payload the caller shares can carry that write. A variant receiver therefore takes the same address-taken marking as an enum one, on the declared type and on a generic instance alike, so the boxing pass has a slot to write back to.

`TypeTable::is_boxed_reference_target` is the one predicate behind this and behind boxing itself, so the two cannot drift; this closes gap D2 in the reference-representation WEP.

## `xs[i].m()` on a variant element

`xs[i]` hands a `variant` element over as a copy, because `IndexRefMut` declines it — a variant's reference is a detached box — while `IndexRef` still aliases it. A `&mut self` receiver has to be the alias, so the call records the aliasing subscript, which is the borrow `&mut xs[i]` takes, and the write-back pass decides. A method that writes through the payload lands; one that replaces the whole value has nowhere to write it back to and is refused, with the diagnostic pinned by a fixture. A newtype container is refused on the same ground, since `[]` dispatches through its base.

## `known_case`

A local whose only definition is a variant construction has a statically known case. The first arm that case can take collapses to the payload binding it makes, so lowering emits no case test, and the arms the case refuses go. A `VariantTest` / `VariantTag` over such a value folds to a constant.

It runs ahead of the folding rules in the peephole engine: the arms it drops are what leaves a constant condition and a dead branch for them to reclaim. Where a binding is declared at the reference type rather than the payload type — match ergonomics — it declines the collapse and only prunes arms.

The largest effect is on the CM task-return path, where a `Result::Ok` known at the construction site makes the whole `ErrorCode` canonical-ABI flattening dead.

## Shared pattern walk

`ast::for_each_pattern_binding` is the single all-bindings pattern walk, built on `AstVisitor` so a pattern shape added later cannot be a binder that one copy misses. `elaborator/liveness.rs` and `resource_move_check.rs` both call it.

https://claude.ai/code/session_01Q6S3wGFw5SvKG492UxD9W2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q6S3wGFw5SvKG492UxD9W2)_


gfx:
Answered the three findings. Two were real and are fixed in 110a6df; one did not reproduce.

## Fixed: the trait-dispatch case guarded nothing

Right, and it was the sharper version of the problem: the e2e runner reports only the first refusal in a file, so `xs[0].clear()` sitting after `xs[0].replace()` was unreported whether or not it was refused. The fixture read as coverage for trait dispatch while pinning only the inherent path.

Split into `variant_element_mut_self_receiver_trait.wado`, and each half now asserts the method its own diagnostic names rather than the shared prefix, so neither can pass on the other's error:

- inherent → `...Payload::replace' replaces it`
- trait → `...Payload^...Clear::clear' replaces it`

Confirmed each in isolation before splitting — both dispatch paths are refused, with distinct messages.

## Fixed: no coverage for the `VariantTag` fold

The arm was exercised only incidentally, through CM binding and reflection, where the fold is entangled with const-folding — a fixture there would pass on the wrong reason. Added unit tests instead, in an engine holding `KnownCaseRule` alone, so the constant left behind is that rule's and nothing else's: `a_known_case_tag_folds_to_its_index` and `a_known_case_test_folds_to_its_answer`, the latter covering both the matching and non-matching case index.

## Not reproduced: duplicate diagnostics from repeated index resolution

The re-resolution is real, but the consequence isn't. I checked with a subscript carrying its own error (`xs[one(1, 2)].replace()`, arity mismatch) and with the refusal path itself: each diagnostic is emitted exactly once, never twice.

On the annotations: `resolve_index_access(.., Mutable)` overwriting the shared annotation is the mechanism of the fix, not a side effect of it. Recording the aliasing subscript is the whole point — the write-back pass then writes it back or refuses it.

What's left is wasted work — the container and index resolve up to three times. That's worth noting, but it's the elaborator's existing shape (`try_resolve_index_mut_method_call` already re-resolves on the pre-existing path), the path is reached only for a variant element receiver with a `&mut self` method, and removing it means caching resolution results across these entry points — an elaborator change that belongs in its own PR, not smuggled into a wrong-code fix.

CI is green on `fa77140` across all 19 checks, including `Tidy` and all five optimization levels.

---

PR #2040, opened 2026-09-12
perf(core:json): 12% faster canada deserialize, and resolve defaults in their author's module

gfx:
`core:json` holds one `ScannedNumber` per deserializer and scans every number
token through it, so a token costs no allocation — about 12% on json-canada
deserialize. Alongside it, a parameter or field default resolves every name in
the module that wrote it: that module's private items, its import aliases, the
vantage its visibility is judged from, and the binders in scope for it.

## Defaults resolve where they were written

A default expression is written at a declaration and re-elaborated wherever the
default is taken. `TraitEnv` records which module each `AstIdSpace` was parsed
from, so `home_module(node)` answers where any node was written however far its
AST has travelled. One module is one space, asserted where the table is built.

That one question settles what a default asks:

- **Names.** `resolve_ident` asks the writing module for a module-level name, so
  a same-named item at the call site cannot answer for it.
- **Aliases.** Both the alias lookup and the `ns::member` → `ns$member`
  canonicalization read the writing module's `use ns from "..."` table, so a
  default may spell `nsb::make()` or `nsb::SOME` in a caller that declares no
  `nsb`.
- **Visibility.** `visibility_vantage` judges from the writing module, so a
  default may call that module's private function or read its private field.
- **Binders.** `with_caller_bindings_hidden` puts the surrounding frame's
  bindings out of scope while keeping local allocation, so a caller's `let` is
  invisible to a default that spells the same name. A default naming an earlier
  parameter is answered from `Scope::default_arg_types`, consulted below the
  binder tiers, so a binder the default opens itself still shadows it.

Free functions, methods, static methods, field defaults and associated
constants take the same walk. Each pass has one implementation of it —
`fill_trailing_defaults` in annotate, `reify_apply_param_defaults` in reify — so
the type a default is given and the code generated for it read the same binding.

A function-typed global named by a default is reached on the value path rather
than the named-function one, and that path asks the same question. A default
calling its own module's `MAKE()` compiles.

## Compile-time literals

`#include_str`, `#include_bytes` and `#data` key their tables by the module that
wrote them, which is the only module whose paths and `__DATA__` section the
loader read. `#file` / `#line` / `#function` go the other way on purpose and
report the call site.

## Diagnostics

A default naming the callee's own type parameter (`fields: T = T::default()`) is
rejected with "unknown function `T::default`" and nothing else. The compiler-bug
tripwire that guards a default resolving to `()` fires only when no diagnostic
was reported, so it no longer replaces a correct error with a stack trace. WEP
2026-04-11 gains the `Known gaps` section recording what closing that gap would
take; it is not scheduled.

## Benchmarks

`benchmark/README.md` carries a fresh best-of-three `benchmark-all` on an idle
host plus a pinned `benchmark-http-routing` at both worker shapes. The reference
rows moved up a few percent against the previous table, so read the Wado rows
against them. json-canada deserialize lands at 271.80 MB/s, 1.41x off the
fastest row. `wasm-size/README.md` matches a fresh `report-wasm-size` and is
unchanged.

The SWAR digit-parsing attempt that did not pay is recorded in the
`wado-performance` skill's `dead-ends.md`, with the isolated 953 vs 709 MB/s
measurement and the correction it forces to the per-`array.get` cost figure.


gfx:
## Review response

### Fixed

**Finding 1 — a caller's locals hijacking a travelled default.** Real, and wider than the site named. The cause was that a default expression is re-elaborated wherever it is taken, while the elaborator carries "which module am I in" as ambient state. Fixed at the mechanism rather than the site:

- `with_caller_bindings_hidden` — already in the tree, used once — now covers every default site on both sides of the walk (6 in annotate, 4 in reify). It hides the surrounding frame's bindings while keeping local allocation.
- A default naming an earlier parameter is answered from `Scope::default_arg_types`, consulted *below* the binder tiers, so a binder the default opens itself still shadows the parameter.
- The module-dependent questions — namespace aliases, globals, `#include_str` / `#data` keys, match-pattern qualifiers — ask `home_module(node)`, which reads the module off the node's own `AstIdSpace` rather than off ambient state.

That removed the need to splice caller argument ASTs into defaults, so `Expr::substitute_idents` and the `scope_floor` apparatus that existed to compensate for splicing are both deleted.

**A pre-existing P0 found while sweeping the class.** A method's default resolved in the *caller's* module: a caller spelling the same name got its own value (`b.plus()` → 500 where the declaration says 7), and a caller spelling nothing panicked the compiler in `lower::translate`. Reproduced on `origin/main`, so it predates this branch. `MethodDispatch` now carries `defaults_module`, and the method path pads through the same `reify_apply_param_defaults` the free-function path uses.

The existing `default_arg_caller_local_not_in_scope.wado` covers that exact shape but spells `SECRET` as 7 in *both* modules, so it reads identically whichever end answers. `default_arg_xmod_method.wado` gives the two ends different values.

### Skipped

**A full module-perspective swap in annotate**, which would make all 83 ambient reads correct by construction rather than only the ones routed through `home_module`. Tried and reverted: annotate can swap `current_module_source` and the namespace aliases, but `sem.imports` / `sem.decls` stay the caller's, and the inconsistent pair stops resolving globals and match-pattern qualifiers. The callee's `ModuleImports` / `ModuleDecls` live in `state.module_semantics`, which the body walk `swap_remove`s and the `Elaborator` holds no handle to; reify can do the swap only because it has `all_module_semantics` to read. `home_module(node)` is the mechanism that generalizes here, and a `debug_assert!` in `resolve_expr` now checks the invariant the remaining ambient reads rest on — that a travelled walk only ever sees its own module's nodes.

### Gates

`mise run test` and `mise run test-wado` (5412 passed, 0 failed) green, clippy clean, goldens regenerated.

gfx:
## Review response (round 2)

### Fixed

**`T::default()` in a value default panics the compiler.** Real, and it reproduces. But the panic fires *after* the compiler has already reported `error: unknown function 'T::default'` — so a correct diagnostic reached the user as a stack trace. The tripwire now fires only when nothing was reported, which is the case it exists for. Pinned by `default_arg_names_callee_type_param.wado`.

I did **not** register the callee's generic parameters at the call site, as suggested. That would let the default resolve against `T` as a *binder*, but nothing downstream substitutes the call site's concrete type for it, so `info::<i32>("hi")` would produce a wrong value instead of an error — trading a loud failure for a quiet one. Closing the gap properly means resolving value defaults per monomorphization, which changes a phase's contract, so it is a decision for the repo owner rather than something to adopt from a review.

The WEP had no `Known gaps` section at all, despite the panic message citing it for deferred work. It has one now: what the gap is, why the narrow fix is insufficient, and that the choice is open. (`docs/AGENTS.md` requires the section regardless.)

### Skipped

**Passing `dispatch.param_is_mut` when padding static-method defaults.** Not real, and the patch would introduce a defect rather than fix one.

`Param::is_mut` is the `mut` *before the parameter name* (`fn f(mut b: i32)`) — the callee's own local binding. `CallArg::is_mut` means something else, and `lower/translate.rs` says so directly:

> `ArenaCallArg::is_mut` means "the callee may write the caller's storage through this slot", which is `is_mut_ref` […] A declared-`mut` binding is the callee's own local and says nothing about the caller's.

So the suggestion feeds `CallArg::is_mut` from the source the codebase documents as wrong. Checked both shapes — `mut b: List<i32> = […]` as a static method and as an instance method (where reify passes the flag today) — and they produce identical results; a `&mut H = &mut SHARED` default also writes through correctly either way.

Fixing the class instead of the site: a synthesized default is a value built at the call site, so **there is no caller storage behind it** and `false` is correct by construction everywhere. The `param_is_mut` parameter is removed — including from the method path that had been passing it — leaving one behaviour at all four sites.

### Gates

`mise run test` and `mise run test-wado` (5412 passed, 0 failed) green, clippy clean, goldens regenerated with no diff.

gfx:
## Review response (round 3)

The three findings are instances of two classes.

### Class 1 — a module-dependent question answered from ambient state rather than from the node

This is the class the branch exists to close, and the review found two doors still open.

**`ns::member` canonicalization (Major) — fixed.** `resolve_ident` rewrites `ns::member` to the `ns$member` key the registries hold, and it read the alias map of whichever module the walk stood in. A default written as `nsb::SOME` and taken by a module with no `nsb` alias stayed uncanonicalized, so the global lookup missed and a valid program was rejected with "unknown identifier `nsb::SOME`".

Fixed at the mechanism, not the site: one `namespace_imports_at(node)` accessor answers "which aliases are in scope here" from the module that wrote the node, and both the alias lookup and the canonicalization take it from there.

`default_arg_ns_global.wado` pins it — red on the parent commit for both the parameter default and the field default, green now. The reviewer was right that the existing `nsb::make()` fixtures cover a namespaced *function* and not a namespaced *global*.

**The use→def edge (Minor) — half fixed.** `record_item_reference_by_name` searched the walk's own module for a name spelled elsewhere, so a travelled expression could record an edge to a *same-named item of the module that took it*. It asks the writing module now. That is the sink every caller goes through, so it closes the class rather than the default path alone.

The other half I could not substantiate. The claim that `spelled_references` omits the value from liveness and unused-item analysis does not reproduce: a module-private `global` and a module-private `fn` reached only through a travelled default produce no "never used" warning and the right value. Go-to-definition on a plain foreign global already lands on the declaring module. I added the extra recording, failed to make it change any observable behaviour, and reverted it — an addition that guards nothing reads as coverage.

### Class 2 — a node-keyed query asked about the wrong node

**`decl_key_at(call_id, …)` (Minor) — skipped.** The claim is structurally right: the name is spelled by the receiver type's head, not by the call expression. But that fallback runs only when the resolution *and* the inherent-entry index both come up empty, and a static call through a renamed import (`use { Real as Alias }`, then `Alias::make()`) resolves correctly today. Without a case that reaches the fallback, changing the `AstId` there is a change whose effect nothing demonstrates. Reopen it with a program that reaches it and it gets fixed.

### Unrelated, found on the way

Go-to-definition on a namespaced global (`ns::GLOBAL`) returns "No definition", and `references` misses the use — in ordinary code, with no default involved. Pre-existing and outside this branch; recorded here so it is not lost.

### Gates

`mise run test` and `mise run test-wado` (5412 passed, 0 failed) green, clippy clean, goldens regenerated.

PR #2041, opened 2026-09-12
feat(core:fs): whole-file I/O in one call, and a read that fails instead of truncating

### LINKED ISSUE #2036: Reading and writing one file from a Wado tool is 54 lines of `wasi:filesystem`, and two packages that both need it cannot share it
Friction from writing `package-gale/tools/rust_inline_paths.wado` and `package-gale/tools/rust_corpus.wado`. Facts only.

- `read_file` is 26 lines (`rust_corpus.wado`, lines 74-99): `open_at` with four arguments through `.wait()`, a `match` on the result, `read_via_stream`, `read_to_end`, `read_stream.drop()`, then `String::from_utf8` with a second `match`. The whole body exists to turn a path into a `String`.
- `write_file` is 28 lines (`rust_inline_paths.wado`, lines 123-150): `open_at` with `OpenFlags::Create | OpenFlags::Truncate`, `content.bytes().collect()`, `Stream::<u8>::new()`, `write_via_stream`, `tx.write_all`, `tx.drop()`, `sub.read()`, `sub.drop()`, and a check on what `sub.read()` answered.
- Three of those lines are hand-written `drop()` calls in a garbage-collected language. Nothing in the type system says which of the four values (`file`, `rx`, `tx`, `sub`) still needs one, so each was worked out by reading `wasi:filesystem`.
- No `core:*` module wraps any of this. `wado-compiler/lib/core/` holds 27 modules (`json`, `cbor`, `url`, `zlib`, `temporal`, …) and none of them touches files.
- The same helpers exist twice. `package-gale-highlight-wado/tools/corpus.wado` (67 lines) holds `corpus_root`, `collect_paths` and `read_file`, near-identical to the copies in `package-gale/tools/rust_corpus.wado`. The only channel between the two packages is Gale's published API.
- That duplication has already cost something: the hardening for an unreadable `--paths-from` list (`rust_corpus.wado`, line 107) had to be ported into the other package by hand rather than shared.

gfx:
`core:fs` reads and writes a whole file against the preopened directory tree in one call, and 23 files reach the filesystem through it — the benchmarks, `example/tree.wado`, `package-marl`, `package-gale` and `package-gale-highlight-wado`. Reading one file is `fs::read_to_string(&path)?`.

The module is also the one place that knows a transfer's completion future has to be read. `read_via_stream` returns a stream and a future, and a mid-transfer I/O error is reported only in the future — a stream ends on an error the same way it ends on the last byte. `finish()` reads it on every path, so a truncated file is an `Err` rather than a short read that looks whole.

Closes #2036.

## The API

```wado
use fs from "core:fs";

pub fn root() -> Result<Descriptor, FsError> with Preopens;

pub fn read(path: &String) -> Result<ByteList, FsError> with Preopens;
pub fn read_to_string(path: &String) -> Result<String, FsError> with Preopens;
pub fn write<T: AsByteSlice>(path: &String, data: &T) -> Result<(), FsError> with Preopens;
pub fn remove_file(path: &String) -> Result<(), FsError> with Preopens;

pub fn read_dir(path: &String) -> Result<List<DirEntry>, FsError> with Preopens;
pub fn create_dir(path: &String) -> Result<(), FsError> with Preopens;
pub fn create_dir_all(path: &String) -> Result<(), FsError> with Preopens;
```

A path resolves against the first preopened directory, which `wado run` grants as the current one; `""` and `"."` name it, so a recursive walk has a root case. `FsError` carries the path and `Display` renders `path: message`, so ``eprintln(`error: ${e}`)`` is the whole error path. The design is in `docs/wep-2026-09-12-core-fs.md`.

## Failures the signature has to keep

`read` refuses a path that names anything but a regular file. No open flag says "a regular file", and `read_via_stream` traps on anything else, which aborts the program instead of returning the `Result` the signature promises. The other operations state what they expect in their open flags, so the host makes that check for them.

`create_dir_all` refuses an absolute path. It is the one operation that splits the path rather than handing it to the host, so it makes that check itself. It also refuses a path a regular file occupies: the host answers `Exist` for whatever is there, which alone would report a directory that does not exist. `create_dir` is the single-level `mkdir` beside it, and hands the host's answer straight back, as Rust's `create_dir` does.

## What stays on `wasi:filesystem`

`root()` hands over the descriptor for everything the module leaves out: append, rename, symlink, times, metadata, and streaming.

`example/cat.wado` keeps its raw WASI code on purpose. It connects a file's read stream straight to stdout and never holds the file in memory, which is the shape `core:fs` would break rather than shorten.

`package-gale/src/main.wado` keeps its own opener. It searches every grant and names the ones it searched, which is what makes `wado run --dir` legible there; `core:fs` resolves against the first preopen only.

## Generated highlighters

`package-gale/src/highlight_gen.wado` writes the file-reading `run()` into every highlighter it emits, so that boilerplate is `fs::read_to_string` with its error reported. The 734 `.kiln.json` files in the diff carry the generator's source hash, which every grammar's invocation cache records; only the two highlight corpora's generated parsers change with it.

## Shared corpus helpers

`lib:gale` publishes `collect_paths` and `line_of` alongside `run_cli`, so the Rust and Wado corpus tools in two packages share one implementation of the `--paths-from` command line and one of the byte-offset-to-line mapping.

https://claude.ai/code/session_013ULBkkCkorcHoTaTwxpfFM

---
_Generated by [Claude Code](https://claude.ai/code/session_013ULBkkCkorcHoTaTwxpfFM)_


gfx:
Both review findings were real and are fixed in `cd81d8ae7`, with the third (the cheatsheet) alongside them.

**`create_dir_all` accepted `Exist` without checking what exists.** `create_directory_at` answers `Exist` for whatever occupies the path, so a regular file there left the call reporting success for a directory that does not exist. It now stats the entry and accepts only a directory; a file is `NotDirectory`. A test writes a file and then asks for that path as a directory — it returned `Ok(())` before the fix.

**`package-marl`'s walk dropped a `read_dir` failure.** `collect_md_files` returns a `Result` and propagates it through `resolve_paths`, and `run` exits non-zero, so `--check` cannot pass on a subtree it never saw. Swept the rest of the migrated tools for the same class: every other one already records the failure (`failed += 1`, `exit(1)`, a reported `F` line, or the unreadable path pushed back into the list), so marl was the only site.

**The cheatsheet's `core:fs` example referenced an unbound `e`.** It now binds it in the `if let Err(e)` that produces it, which also shows what `FsError` renders as.

Unrelated to the review: the module's tests moved out of `wado-compiler/tests/fixtures/core_fs.wado` and into `wado-compiler/lib/core/fs_test.wado`, where every other `core:` module keeps its own. `wado test` preopens the working directory rather than a scratch one, so each test owns a directory under the gitignored `target/` and reads only what it wrote there.

---

PR #2042, opened 2026-09-12
feat!: a `wado` tag that escapes each hole by its template context, and a prelude scope limited to `core:prelude`'s exports

gfx:
`wado` is a tagged template that emits Wado source and escapes each hole for the
lexical context the template's own text puts it in. `` wado`"${s}"` `` is a valid
string literal whatever `s` holds, `` wado`'${c}'` `` a valid char literal, and a
hole that lands in code is written as it renders. The context is a property of
the literal text, so a generator no longer wraps values to say what they are
about to become, and the tag costs nothing at run time: the scan folds to a
constant at `-O2`, leaving only the escape of the hole's own rendering.

The prelude scope is `core:prelude`'s export surface. A name reaches every module
without a `use` only if it reaches outside `core:` on its own, so what a prelude
implementation module declares for its siblings is nobody's to write unqualified.

## The `wado` tag

`Ctx` carries the compiler lexer's own states — code, string, char, template,
line and block comment — with a pending `\`, `/` or `*` a state of its own. The
tag walks each literal segment through it and escapes the hole's rendering for
the delimiter that state names: `"`, `'`, or `` ` `` and the `$` that would
otherwise open an interpolation in the generated code. Control characters follow
`Inspect for String`; every other byte passes through, so printable non-ASCII
survives intact.

`step`, `open` and `advance` are separate functions so each fits the inliner's
budget. At `-O2` none of them survives in the WIR — only `escape_in_place` over
the hole's own text, which returns after one scan when there is nothing to
escape.

Call sites in `package-gale` write the delimiters themselves: `str_lit(x)` is
`` wado`"${x}"` `` and `char_lit(c)` is `` wado`'${c}'` ``. `StrLit`, `CharLit`,
`str_lit` and `char_lit` are gone from `wadopoet.wado`.

## The prelude scope

`Scopes::build` gates both prelude loops on `Visibility::reachable_from`: a
declaration and a re-export each enter the prelude tier only when it reaches
another package. Builtin types enter on their own footing — `i32` names the same
thing in a module that imports nothing, `#[no_prelude]` included, so where it
was declared says nothing about who may write it.

A fixture pins the consequence: `UnpackResult`, declared in
`core:prelude/fpfmt.wado` and never re-exported, is a compile error unqualified
and still reachable through `use { UnpackResult } from "core:prelude/fpfmt.wado"`.

## Auto-imported traits

`Default`, `Inspect`, `From`, `TryFrom`, `FromStr`, `LenientFromStr`,
`LiteralSpread`, `Ref`, `RefMut`, `Sequence`, `AsSlice`, `Member` and the whole
`Reflect` family are `pub` and re-exported from `core:prelude`. These are names a
module writes rather than calls — a bound or an `impl` head — which is the
footing `Add`, `Display` and `IndexRef` were already on: the compiler resolves
the behaviour by `#[compiler_item(...)]`, and the name is needed only to state
it. The member handles (`StructField`, `VariantCase`, `EnumCase`, `FlagsBit`,
`Hole`) are `pub` but stay out of the prelude, where the name would collide with
a user type of the same spelling.

## Along the way

A brace is literal in a Wado template, so the `\{` and `\}` escapes across the
generator corpus and `core:` tests are gone. `docs/spec.md`'s escape table now
names `` \` `` and `\$` as template-only, which is where they are accepted.

https://claude.ai/code/session_01NrsnuLbJkCXhYCF4FybdNb

---
_Generated by [Claude Code](https://claude.ai/code/session_01NrsnuLbJkCXhYCF4FybdNb)_


gfx:
Answered the review in c188a2a.

## Fixed

`tagged_template_lit_scan_fold.wado` — real, and the claim was the bug. `step`/`scan` now carry the pending escape and the pending slash as states of their own (`Lex`, seven states plus a nested `open`), so they recognize the language `big` does rather than a weaker one. A third template pins the difference: an escaped quote closes nothing, so the segment ends inside the char literal it opened, and a split that dropped the pending escape prints no `#`. `wado dump --wir -O2` confirms the fixture's point survives the larger automaton — `big` is the only one of the five left in the WIR, `step`, `open` and `scan` fold away — so the budget claim now holds at seven states plus a call level rather than four.

`resolve.rs` scope-order comment — real. The same stale statement was in layer 4 of [WEP: Declaration Identity](https://github.com/wado-lang/wado/blob/main/docs/wep-2026-08-12-declaration-identity.md), which the finding did not name; both now say the prelude tier is what `core:prelude` exports, plus the builtin types on their own footing.

## Skipped

`## Implementation` → `## Roadmap` in WEP: Visibility. `docs/AGENTS.md` defines Roadmap as work that *will* be done, ordered, and splits it from Known gaps on commitment. That section records landed work, so the rename would misstate it — the uncommitted API narrowing is already under `## Known gaps`, which is where the rule puts it. The repo spells this section `## Implementation` in 37 WEPs and `## Roadmap` in 5.

---

gfx:
Answered the second review in e4476dc.

## Fixed

`wadopoet.wado` — real, and reproduced before fixing. On `` wado`let s = "a\\${x}"; let t = ${y};` `` the tag emitted:

```
let s = "a\q"; let t = \"str\";
```

The escape state survived the hole, so the second hole was escaped for a string it was no longer inside and `\"` landed in code position.

The suggested repair — rescanning the rendered hole and advancing `ctx` — cannot be right, and would cost the property the tag is built on. It cannot be right because the pending backslash binds the hole's *first emitted byte*, and where that lands depends on the value: a plain `z` stays inside the string, while an escaped `"` renders as `\"`, pairs the backslash into `\\`, and closes it. It would cost the fold because the scan would then read the hole's value instead of the template's own constants, and `tagged_template_lit_scan_fold.wado` and [the optimizer guide](https://github.com/wado-lang/wado/blob/main/docs/optimizer.md#not-yet-implemented) both pin that the whole walk folds at `-O2`.

A literal ending mid-escape is a malformed template, so the tag asserts. The condition reads only the scan's state, which the fold makes constant, so `-O2` leaves no trace of it — verified: no `step`, `open`, `advance`, `hole_quote` or `needs_escape` survives in the WIR, and neither does the assert. Two tests cover it: a completed escape before a hole still carries the state correctly (the hole after the closing quote is written raw, as code), and the dangling one is `#[expect_trap]`.

## Skipped

The `Roadmap` section, again, and now with the WEP's contents checked rather than its headings. `docs/AGENTS.md` splits Roadmap from Known gaps on commitment: a roadmap item "will be done", ordered, each entry saying what finishing it means. That WEP has no unfinished item — every entry under `## Implementation` is `- [x]` — and its one open question is the stdlib `pub` narrowing, already under `## Known gaps` and marked the human's call there, which is exactly where the rule puts an unowned gap. So a `Roadmap` section could only say "nothing is committed", which the absence of roadmap items alongside a populated Known gaps already says.

Whether `## Implementation` is the right spelling for landed work is a corpus-wide question rather than this branch's: 37 WEPs use it, 5 use `## Roadmap`. That is the repository owner's call, so I have left it and raised it rather than changing one file out of 37 in a branch about something else.

---

PR #2045, opened 2026-09-13
feat(elaborator): a default argument may name its callee's own type parameter

gfx:
A value default may name a type parameter of the declaration that wrote it, at
every site that takes one: a free function, an instance or static method, a
trait method, a struct field, and across a module boundary. `T` there stands for
the type argument the taking site settled on, so `fn zero<T: Default>(v: T =
T::default())` resolves `T::default()` against `i32` at `zero::<i32>()` and
against `String` at `zero::<String>()`. A variadic pack is a type parameter like
any other, so `t: [..T] = [..T::default()]` expands to one element per type the
call supplied.

Three ways a type position could reach `unknown` without reporting anything now
report at the site instead, which is what the `interface`-in-a-turbofish case
turned into a monomorphization panic.

## What settles a type parameter for a default

A site settles each parameter of the declaration it calls, and the walk of the
default runs under those bindings. `SettledAs` is what one settles to: a type,
or a pack. A pack carries no type there, because nothing could be handed one —
`[..T::default()]` means the pack's elements, and the tuple would spell
`[i32, String]::default`, a callee nothing declares. The pack a name enters
scope as is minted in `with_callee_type_bindings` and nowhere else.

Each taking site keeps its own walk of the one AST default node, in
`BodyFacts::default_overlays` keyed by the site. `omits_a_default` is the single
predicate for whether a call leaves a defaulted parameter for a walk to fill;
annotate walks exactly when it holds and reify pads exactly when it holds, which
lets reify assert that annotate left it a walk.

## Packs

`SubstitutionContext::substitute` splices a pack into the tuple that holds it:
`[..T]` with `T = [i32, bool]` is `[i32, bool]`, never `[[i32, bool]]`. That one
arm is what made a method's own pack work — `fn spread<..U>(&self, u: [..U])` on
an `impl<T> Box2<T>` — by turbofish and by inference alike.

Monomorphization expands a settled pack before it collects instantiation sites,
so `[..T::default()]` in a default reaches codegen as the tuple the call
settled.

## An interface or a trait is not a type

Both are declared in the type namespace: one name reaches one declaration
wherever it is written. Neither denotes a type, so a type position naming one is
an error saying so — `` `Counter` is an interface, not a type: it names a set of
operations, and no value has it as its type ``. A `resource` stays a type.

The check runs on the two routes a written type takes: the annotation walk and
`resolve_turbofish_args`. It does not run on an `impl` head, where a trait name
is exactly what belongs.

## Two calls that resolved to `unknown` in silence

`ns::Trait::method(&recv)` names the same declaration `Trait::method` does: the
trait is the segment before the method, whatever leads up to it. And
`ns::Missing::method()` where the namespace declares no `Missing` reports an
unknown function at the call.


gfx:
All four findings verified against the code. Three are real, and they are three classes rather than four instances — fixed in 0941c2d.

**A pack's turbofish was grouped after things that count slots** (`method_call.rs`, `call.rs` — covers the `call.rs` and `method_call.rs` findings)

A method's own turbofish spells one argument per pack element, so it has to be grouped into slots before anything counts them. It was grouped once, late, and on the instance path only. Grouping now precedes every count, on all three spellings:

- `Box2::<i32>::spread::<i32, bool>()` compiled and returned `"0"` instead of `"0false"` — the pack default took only the first argument.
- `b.mixed::<i32, bool, String>()` on `fn mixed<U: Default, ..V: Default>(u: U = U::default(), …)` reported `unknown function 'U::default'`: the mismatched count made the binding discard `U = i32`.

The `param_types` → `checked` suggestion was the instance you saw of this, not the cause; with the list grouped, that dispatch stores the right thing. I did not take the diff.

**A settled pack was expanded in functions only** (`func_inst.rs` finding — real)

`global G: [i32, bool] = made::<i32, bool>();` where `made` carries `t: [..T] = [..T::default()]` hit `TirExprKind::TypePackExpansion should be expanded by monomorphize before lower::translate runs`. The pass now walks each global initializer with its own locals.

**A namespaced head never reached the check** (`type_resolution.rs` finding — real)

`mk::<m::Greet>()` resolved to `unknown` and panicked in monomorphize. `walk_type_heads` now reports a namespaced head like any other, and leaves the projection forms `Self::Assoc` and `T::Assoc` to the projection, as you noted. A namespaced name whose local module declares a trait of that same name still resolves correctly — the recorded declaration answers before the spelling fallback.

Fixtures: `default_arg_callee_type_param.wado` gains a static with its own pack, a scalar ahead of a pack, and a global initializer taking a pack default; `interface_trait_not_a_type_error.wado` gains `mk::<m::Greet>()`.

PR #2047, opened 2026-09-13
feat(elaborator): a variadic pack nothing settles is the empty pack

gfx:
A call that leaves nothing over for a trailing variadic pack settles it to the empty pack. `headed::<i32>()` on `fn headed<A: Default, ..T: Default>(a: A = A::default(), t: [..T] = [..T::default()])` compiles and returns `"0"`, as does `packed("e:")` on a pack-only declaration — the one spelling of an empty pack, since `::<>` is a parse error. The rule holds on every call kind: free function, instance method, static method, and a global initializer taking the default.

## What settles it

A pack stands for the type arguments left over. Two shapes leave none: a turbofish naming every parameter ahead of the pack and stopping there, and a call whose written arguments pin nothing for the pack. `settle_empty_pack_of` answers both, and runs at each point where a slot can still be open — the free-function path after associated-type projection, `bind_method_type_args`, the value-default walk's `known` bindings, and the static path once its own solve has run. On the static path a pack no longer counts as an unanswered slot, so `UninferredStaticTypeArg` stays for what a call really must spell.

The walk's bindings matter for a scalar ahead of the pack: settling `T` there is what lets `A::default()` in `bi.headed::<i32>()` resolve against `i32`.

## What it leaves alone

A caller forwarding a pack of its own. Such a pack interns to the same `TypeId` as the callee's, so a scope holding a pack leaves the slot as it found it and `forwarding([1, "x"])` still passes its tuple through.


gfx:
## Review response

One finding, real, one class — fixed in 3ea313e23ba.

**Fixed: `real_type_params` counted the wrong space.** It filtered on `!p.is_effect`, while the dense type-argument space is `GenericParam::is_real_type_param()`, which also excludes a non-pack `fn`-bound parameter. The three consumers CodeRabbit listed — the slot count `turbofish_leaves_slot` reads, `group_variadic_type_args_of`'s pack position, and `settle_empty_pack_of`'s comparison — all index off that count, so the class closes at the one filter. The suggested diff was right and I took it.

Reproduced first, on `fn applied<F: fn(i32) -> i32, ..T: Default>(f: F, t: [..T] = [..T::default()])`:

- `applied(|x| x + 1)` — `unknown function 'T::default'`, then `cannot infer type parameter 'T'`
- `applied::<i32, bool>(|x| x + 1)` — `type mismatch: expected '[i32]', found '[()]'`, the slot shifted one along

Both are assertions in `default_arg_callee_type_param.wado` now.

**Swept, nothing else in the class.** The tree's other three `!p.is_effect` filters are not the slot space: `register_generic_params` visits an `fn`-bound parameter deliberately without advancing the index, and `item.rs`'s operation scope and `bind_trait_type_params_from_impl` index surface-written arguments. While there, `defer_or_report_uninferred_fn_type_args` now calls `real_type_params` rather than re-rolling the filter.

**Skipped: nothing.** No other finding was raised.

PR #2052, opened 2026-09-14
fix(compiler): two wrong-code bugs, from an EMI campaign widened to a loop guard and five levels

gfx:
The EMI campaign now injects its guard as a loop as well as a branch, and calibrates and mutates at all five optimization levels. Its first run found a wrong-code bug at `O1` and another at `O2` — the levels a release build and `wado test` actually run. Both are fixed here, each with a minimized e2e fixture. The campaign then runs clean: 1,729 sources scanned, 1,562 eligible, 34,390 injection sites, 0 findings, both shapes calibrating on every source.

## The `O1` bug

`ValuePool::canonical_local` caches one `Opaque(Local)` per slot. `sroa_param` retypes the slot it scalarizes, so a cached value can answer for a type the slot no longer holds, and the WIR extractor re-emits the read from the type stamped on it — `struct.get $Point 1` against a `String` local, which core Wasm rejects. The cache hit is now conditional on the requested type matching the one the value was minted at.

Fixture: `opt_promote_fields_retyped_local.wado`.

## The `O2` bug

`const_forward` invalidates what a call can mutate through its arguments. A `&mut local` reaches a call from inside a value block just as a bare argument does, so the walk that finds those references descends through blocks; an argument's nested call is its own channel, invalidated by the walk that reaches it.

What an argument hands over is `collect_reference_locals`, the pass's one answer to that question, used by both invalidation walks and by the aliasing scan. It takes a reference-typed read of a local wherever it sits in the argument, and stops at a load: a `StructGet` or `ArrayGet` hands over the field's pointee, not the base, and no fact recorded for a non-`aliased` local is reachable from that pointee. Taking the base instead would cost real optimization — `gale_gen`'s WIR grows by 30 bytes — for no correctness.

WIR output is byte-identical to `main` on `gale_gen`, `json_catalog`, `fts`, `cbor_twitter` and `sieve`.

Fixture: `opt_const_forward_ref_in_block_arg.wado`.

## The campaign

- A guard is written in each `Shape`, `if` or `while`. The loop shape is what reaches the loop passes, and its value block is what puts a reference in a call argument.
- Calibration judges one shape at a time and `corpus.txt` records which shapes each source answers for, so a source that is an oracle for one shape stays a subject under it. The report's `dropped combinations` section names each combination a source lost and why.
- `OPT_LEVELS` is all five; `WADO_EMI_LEVELS` trades levels for corpus under a time budget, and the CI workflow sets it for both stages at job level — a mutation stage reading a corpus calibrated for other levels is not comparing like with like.
- The worker pool is bounded by `MemAvailable` as well as by cores, at two gigabytes a worker. A whole compile lives in a worker and the stdlib's modules peak past a gigabyte each, so a core count alone invites the OOM killer hours into a run.
- CI shards default to 8, and `emi_summary.ts` reports guard shapes alongside the totals and exclusions.


gfx:
Both findings verified against the current code; both were real, and both are fixed. Two commits: `6541270cf42`, `5843637dc2e`.

## Fixed — a container field hands over the local's own object

The inline finding reproduces. Red unit test first: with `b = S { f: 7 }`, `a = Outer { child: b }`, the statement `f(a.child) + b.f` folded `b.f` to the stale `7`.

The framing in the finding is that `collect_reference_locals` should follow a `StructGet` result. That treats it as a rewrite-order problem, and it is not — the same staleness happens with no rewrite at all, because `b`'s object is inside `a` and any path to `a` reaches it. Nor is a fact lookup enough: `a.child = b` may never be recorded (when `a` is itself aliased) while `b` keeps its own facts, and the read is still stale.

The class is that `collect_aliased_in_instr` marked a local aliased only where a reference to it was taken (`RefAsNonNull`). A bare reference-typed `LocalGet` in a `StructNew`, `StructSet`, `ArrayNew`, `ArrayNewFixed` or `ArraySet` hands over the same object. Marking those closes the hole, and it makes `collect_reference_locals`' stop at a load true by enforcement rather than by assertion — the comment now cites where.

The cost is real and stated rather than hidden: `gale_gen`'s WIR loses 38 lines of forwarding, where a freshly built `String`'s `used: 0` no longer survives its own object being stored into an AST node. `json_catalog`, `fts`, `cbor_twitter` and `sieve` are byte-identical. Those forwards were unsound instances of the same class. Recovering them wants a flow-sensitive alias set, which is a design change rather than a sharper syntactic rule.

No e2e fixture: a Wado struct literal holding a reference emits `RefAsNonNull`, which already aliased, so I could not reach the shape from source. The hole is at WIR level and the unit test is the honest level for it.

## Fixed — per-shape site totals

Real. Calibration is exact, since every shape shares one site set, but the mutation stage can leave a source with `(if, write)` and lose `(while, write)`, and the union then belongs to neither shape. `Eligible.shapes` is an `IndexMap<&str, usize>` carrying each shape's own count; `Eligible.sites` stays the union, which is what the corpus line wants.

Not done as asked: a focused test at the `mutate()` level. Constructing "different site sets per shape" there needs a source that loses one pair across a real five-level compile-and-run, which is a campaign and not a test. `the_shape_report_counts_each_shape_s_own_sites` pins `per_shape` instead, which is where the defect lived.

## Validation

Full e2e at all five levels 12,617 passed; `mise run test` 39 suites; `mise run test-wado` 1,373 files, 5,454 tests; `const_forward` unit tests 8; EMI self-tests 29. All zero failures.

gfx:
The `GlobalSet` finding is right, and chasing it undid the cost the previous round paid. `d3bf81c3423`, `fb9490eabbe`.

## Both, at the right altitude

`GlobalSet` is a container like a struct field or an array element, and the global case needs the fix twice over: `invalidate_effects_in_instr`'s `StructSet` arm only recognizes a `LocalGet` base, so a write through `GlobalGet` invalidates nothing on its own.

Fixing it made the shape of the real mistake visible. Marking a local aliased for the whole function because its object goes into a container is the wrong altitude: the store is a point in the flow, and `FieldKnowledge` is already flow-sensitive. The facts a local held *before* its object went in are still good where they were recorded.

So `collect_container_escapes` — `StructNew`, `StructSet`, `ArrayNew`, `ArrayNewFixed`, `ArraySet`, `GlobalSet` — now invalidates at the store, and `collect_aliased_in_instr` is back to the references that outlive a statement.

## The cost is gone

`git diff` of the golden WIR against the state before any of this is **empty**: all 29 fixtures and all 77 lines are back, and `gale_gen`, `json_catalog`, `fts`, `cbor_twitter` and `sieve` are byte-identical to their pre-fix dumps. The pattern that was paying for the old rule — a freshly built `String` losing its own `used: 0` at the site that reads it, because the string is appended to a `List` later in the function — folds again.

## Tests

Three in `const_forward`, ten passing in all:

- `a_field_that_holds_a_local_s_object_is_a_channel_to_it`
- `a_global_that_holds_a_local_s_object_is_a_channel_to_it` — this finding
- `a_fact_holds_until_its_local_s_object_goes_into_a_container` — both halves: the read before the escape folds, the read after it does not. This is the one that catches a regression back to the blunt rule.

Full e2e at five levels 12,617 passed; `mise run test` 39 suites; `mise run test-wado` 1,373 files and 5,454 tests. Zero failures.

gfx:
Two of the three findings were real, and each turned out to have a sibling the review had not reached. `c8f8cbe9a2d`, `bbe1fc11e1d`.

## Fixed — the container enumeration was short by two

`collect_container_escapes` lists by hand every instruction that puts a value somewhere outliving the statement, and `_ => {}` admits whatever the list is missing.

`ArrayFill` is real: `wir_build/calls.rs` lowers `builtin::array_fill<T>`, and `Array<T>` takes reference elements. `TableSet` is the same class and nothing named it, so it went in too — the criterion is what the IR can express, not what today's lowering happens to emit.

`an_array_fill_puts_the_local_s_object_in_the_array` fails with the `ArrayFill` arm removed and passes with it. Nothing can make the enumeration exhaustive, so the doc now states the criterion a new variant is measured against.

## Fixed — the summary dropped whole report sections

`emi_summary.ts` consumed `guard shapes`, `findings` and the counted exclusion buckets, and silently discarded the rest.

`dropped combinations` is real, and worse than it looks: the WEP claims the cost of the unfinished retry-payload item is "on the page", while CI's page never showed it. `corpus` was going the same way, taking the per-root breakdown with it.

Rather than name the two sections, any section without a table of its own is now collected and rendered under its own heading, so a section the report gains reaches the summary without the parser being taught about it. Findings still lead, since they are why a run fails. Verified by running the script over a sample report: both sections appear.

## Skipped — the formatter finding

Not real. `mise run format` leaves no changes, before the review and again after. The guideline is met.

## Distill

Answering a review is one of the times to run it, and it earned its keep here: findings and dropped combinations spelled "this source, this refusal" two different ways, `Finding` was `(String, Excluded)` under another name, and the workers were pre-rendering their dropped reasons into strings the reporter then had no say over. `Eligible.dropped` carries the `Excluded` now and the reporter formats it.

## Validation

Full e2e at five levels 12,617 passed; EMI self-tests 29; `const_forward` unit tests 11. Zero failures. The distill touched only the test harness and a task description, so the e2e above covers the compiler change as it stands.

Issue #1850, opened 2026-08-18
Value-copy ownership analysis: finish the three parts WEP 2026-05-21 specifies but does not implement


gfx:
Hand-off from the branch `claude/value-copy-helper-identity-36oxbx` (work on #1588).

That branch fixed the helper-identity defect #1588 named, and — following the trail it opened — four wrong-code paths in the ownership analysis. In doing so it found that three things [WEP 2026-05-21 — Resource Ownership](../blob/main/docs/wep-2026-05-21-resource-ownership.md) specifies are not implemented. The WEP now states all three, and its implementation status lists them unchecked. This issue is to close them.

None is a regression: each is a place where the implementation approximates the design and the approximation costs precision.

## Background: what the branch established

The design is settled and should not be revisited here. Copy decisions are **caller-side and single-phase** — chosen in `lower`, before NIR exists, with no elision pass (`optimize::escape` / `value_copy_elide` were deleted). `docs/optimizer.md` records that an imprecise copy is the analysis's to fix, not the optimizer's.

The branch's own fixes are in, verified, and are the prerequisites for the work below:

- A reference local owns nothing, so a root reached through one resolves to the place it borrows. This was wrong in three separate predicates (`source_root`, the move analysis's freshness seed, the share analysis's path disjointness), each a wrong-code path on its own: a `let` out of a `&`- or `&mut`-held struct kept seeing the source's later writes.
- A self-recursive function can now prove it returns owned. A least fixpoint that only adds could never lift one, so every `?` on `CborDeserializer::skip_item` deep-copied the error it propagated.
- A place repointed after a binding read it releases that binding (`take` / `drain` / `snapshot`).
- `wrap_value_copy` no longer answers a missing helper by passing the value through — it asserts. That silent path had been dropping deep copies in 236 fixtures.

### 1. Key sharing on liveness

The WEP's rule is *"a read-only binding whose storage is never mutated **while live**"*. `compute_share_eligible` implements only the second half: it is a forward walk with no liveness and no control-flow handling (`walk_block` walks statements in order and ignores branches and loops), so a conflicting write **anywhere** in the body refuses the binding.

Both halves are readings of one backward walk, and that walk already exists in the same file — `last_use::Analyzer` computes liveness for the move analysis. Merging share collection into it makes the rule real: a write conflicts with a binding when the binding is live at that write and the two places may alias.

Measured cost of not doing it, against `origin/main` on the golden corpus:

| copies | where |
|---|---|
[...]

PR #2054, opened 2026-09-14
fix(lower): value-copy hand-over decided by place, and a variant pattern carrying its case index

gfx:
A binding skips its defensive copy only when no other chain can reach the storage it takes, and that question is asked of places rather than root locals. Two components of one destructured aggregate name separate storage, so each moves; two bindings read out of one place reach each other, so neither does. That distinction is worth five copies across the `httpbin_*` fixtures. Last-use liveness follows `break LABEL` to where its labeled block ends, which matters most for a template string, since one lowers to a labeled block.

Alongside it, a variant pattern carries the case index the elaborator resolved, so no later phase looks a case up by name.

## Hand-over

One predicate answers at every site that hands storage on, and it closes the chain on both sides: the storage being taken, and everything live where it is taken. A sibling binding reaches a place from beside the taker rather than above it, so closing one side alone does not see it.

Both sides compare as places. A destructured tuple's components share a root but name different storage, and the selectors are what tell them apart. A chain of alias edges is carried up to `PATH_DEPTH` steps before it widens to its prefix, so a cycle cannot grow one without end.

One copy stands in `parser_synth_id_collision_test`, where two bindings read the whole of one place and one is still live — the shape the rule exists for.

## Sharing

A share licenses the binding and nothing built out of it: `let mut b = a` and `Wrapper { inner: a }` mint a second owner that outlives the binding, so each owes its copy even where `a` shares. Only a move exempts them, and a move is proved of the binding itself.

An arm binding aliases live storage when the scrutinee is live where *that arm* reads it, decided from the live set the arm's own walk produced.

A caller takes a `&mut` handle whole when the callee names a write it cannot re-root there, so a write through a variant payload reaches the caller's summary.

## Liveness through labeled blocks

`break LABEL` resumes where its labeled block ends, at both walks that compute last-use liveness. A template string lowers to a labeled block, so this decides the last use of every local in a function containing one interpolation.

## Variant patterns

`TirPattern::Variant` and `PatKind::Variant` carry the case index resolved at elaboration. `const_eval`, `niri`, `known_case`, `labeled_block_fusion`, `sroa_variant_return` and WIR pattern lowering read that index; the written name stays as a spelling, checked against the index by assertion where both are at hand (WEP 2026-08-12).

## Qualified patterns

A case may be written under a qualifier, and the spec says which ones reach it: a name resolving to the scrutinee's type, `Self` inside an `impl`, any name on the scrutinee's newtype chain, or a namespace prefix the type is reachable through. The name must be one the file can see — only the prelude is in scope without a `use`. A qualifier may restate the scrutinee's type arguments and must then write as many as it carries, so a type declaring none takes none.

Only a bare identifier binds. A qualified path names a case, an associated constant, or an immutable `global` — `ns::TK_FOO` reaches one the way the bare `TK_FOO` does.

## Seeding the copy helpers

The seed reads the types a program declares rather than the expressions it writes, so it covers the temps pattern lowering mints after it runs. The lowered-format goldens grow because a program names more types than its wrap sites do; `dce` removes the helpers nothing calls, and the WIR goldens show the emitted module is unchanged.

Part of #1850. Part 3 stays a known gap in WEP 2026-05-21, which records what closing it would take.

https://claude.ai/code/session_014E2eaqBTN4X5u5Q14Fd6fi

---
_Generated by [Claude Code](https://claude.ai/code/session_014E2eaqBTN4X5u5Q14Fd6fi)_


gfx:
Answered the review in `20fe05bec`.

## Fixed

**The hardcoded `case_index: 0`** (`translate/pattern.rs`). Both sites now pass the index the elaborator resolved. The finding named one site and located it in the wrong function, so I swept every `VariantPayload` construction in the tree instead: eight sites, of which these two hardcoded zero and the other six already carry a real index.

On reachability, though, the finding is wrong. `pattern_is_refutable` answers `true` for every `Variant`, so a variant pattern always routes through `build_pattern_check` / `extract_refutable_sub_patterns`, both of which carry the index. Neither arm is reached by a valid nested pattern — only as recovery past a diagnostic. I wrote a fixture for the claimed miscompile before changing anything; it passed on both sides of the fix, and `variant_unit_payload_match.wado` already pins `Some(Err(code))`, an inner case at index 1 with a payload binding, which is the shape the finding describes. So I deleted my fixture rather than leave a test that reads as coverage while guarding nothing. The fix stands on its own ground: the index selects the payload layout, so the site should read the resolved one.

**The qualifier visibility rule** (`docs/cheatsheet.md`). Correct: `Self` and a same-module type need no `use`. The condensed line dropped the "visible at the pattern" half of the spec's rule and kept only the `use` clause. Now it states visibility, with the prelude exception as the reason an unimported type is an error.

**The place-scrutinee roadmap item** (`docs/wep-2026-05-21`). Correct, and a contradiction with the prose above it: "matched where it lies" holds only for a place nothing can write. Now scoped, with the writable case's hoist stated.

## Not fixed

**The blank line before "What blocks it is a write reached through a variant payload"** is not a code block. The list marker `- [ ] ` gives six-space content indentation, so a six-space continuation after a blank line is a paragraph; a code block would need ten. The same shape appears elsewhere in that list and survives `mise run format`.

**`mise run format`** had already run on these docs, and ran again before this push.

Validated: `cargo test -p wado-compiler --test e2e` — 5088 passed, 0 failed; `mise run format` clean.

---

PR #2062, opened 2026-09-15
feat: warn when a binder shadows a known name, and check every declared world

### LINKED ISSUE #2059: `wado check <file>` refuses to resolve a registry dependency without `wado.lock`, but `wado run .` and `wado test` resolve one fine
## The friction

In a package whose `wado.toml` names registry dependencies, three commands in the same tree disagree about whether `wado.lock` is required.

The tree is `wado-lang/wado-lang.github.io` at `claude/wado-hello-blog-post-cotqxs`: `wado.toml` declares `wado-lang:marl` and `wado-lang:gale-highlight-wado`, `wado fetch` has run, and both components sit in the local cache. No `wado.lock` is present — nothing in the repository generates one, and the site's own build tasks never ask for one.

```
$ wado fetch
Fetched wado-lang:gale-highlight-wado@0.0.28 → /root/wado/ghcr.io/wado-lang/gale-highlight-wado/0.0.28/component.wasm
Fetched wado-lang:marl@0.0.28 → /root/wado/ghcr.io/wado-lang/marl/0.0.28/component.wasm
Fetched 2 component(s)

$ wado run .
sheaf: generated 388 file(s) from 5 post(s) and 188 doc(s)

$ wado check src/main.wado
error: cannot resolve dependency 'wado-lang:marl': no `wado.lock` version for "wado-lang:marl"; run `wado update`
```

`wado test` behaves like `wado run`. So the package builds and its tests pass, while `check` on a file in that same package cannot resolve the dependency the build just used.

What made this cost time is that the failure reads as a property of the file rather than of the command. The obvious conclusion from the message is that a single file carries no package context and `check` is simply not usable here, so the inner loop moves to a full package build and stays there. The actual state — `check` works, it just wants a lockfile the rest of the toolchain never asked for — is not visible from the error, because the neighbouring commands in the same directory contradict it.

After `wado update` writes the lock, `wado check src/main.wado` succeeds.

## Second half of the same session

Checking a non-entry module hits a separate wall in the same shape:

```
$ wado check src/doc.wado
error: function `run` is required as a world entry point but is not defined. Define it with: `export fn run(...)`
```

`--world test` checks it without complaint. That is fine once known, but the two together — a lockfile requirement that nothing else in the toolchain shares, then a world default that rejects every library module — is what turns "check one file" into something that reads as unavailable and gets abandoned.

gfx:
A binder that takes a name already reaching a known symbol is reported under
the new `shadowed_name` lint, so a reader never has to work out which `println`
or which `x` a line means. `wado check` gates what `wado build` produces: with
no file it checks every world `wado.toml` declares, at `O0`, which is 2.5x
faster than the build on package-gale-highlight-wado.

## The `shadowed_name` lint

Every binder counts — a `let`, a parameter, a closure parameter, a type
parameter, a pattern binding, a local item — and so does every symbol it could
take the name from, in any namespace: a function, a global, a type, a trait, a
case, an outer binding.

```wado
fn draw(Point: i32) { }        // warns: `Point` shadows the struct
let println = 1;               // warns: `println` shadows the function
```

The exemption is the derivation `docs/wep-2026-03-25-same-scope-shadowing.md`
sanctions, read off the binder's own source rather than off the `let` keyword. A
binder whose source already mentions the name it binds rebuilds that binding
instead of hiding it, and that holds for `let`, `if let` and `while let` at any
scope. `if let Some(x) = y` warns; a match arm is never exempt, since its
scrutinee stays in scope across every arm.

`#[allow(shadowed_name)]` on the binder waives it and `#![allow(shadowed_name)]`
waives a module, which the parser now accepts on a parameter, a closure
parameter, a type parameter and a `let`. A `#![generated]` module is never
reported: a generator's names are not the author's to rename.

The lint reports nothing across `package-gale`, `package-jade`, `package-marl`,
`example/`, `benchmark/` and the stdlib.

## `wado check`

With no file, `check` takes `build`'s target selection: `[package].lib` and
each `[world]` entry, each against its own contract. Naming a file checks that
file alone, against the world whose `[world]` entry names it and otherwise the
library world, which requires no entry point — so a library module checks as
itself. `--world <name>` opts into a named world's contract in either mode.

It runs every phase a build does, codegen included, so no diagnostic is lost,
and it drops to `O0` because it throws the component away. The optimization
loop reports nothing at any severity, and on a large program it is most of the
run.

| target | check | build |
| --- | --- | --- |
| package-gale-highlight-wado | 5.01s | 12.39s |
| package-gale | 1.16s | 1.43s |
| package-jade | 0.66s | 0.83s |

`check` acquires dependencies like the build it gates: `Acquisition::Build`
pulls what the cache lacks and resolves an unpinned version live. `query` is
the one command held back at `Acquisition::Analysis`, pulling only what a pin
names and degrading a failure to an `unresolved` hint, so an editor request
never waits on a version listing. Offline resolution pins from the cache where
the lock does not, so a project that ran `wado fetch` resolves in every tier
with no lock at all. `wado lsp` installs a prefetcher through
`wado_lsp::host::prefetch`, which warms a pinned-but-cold dependency in the
background.

Closes #2059

## Indexing

A write into an element rather than over it — `c[i].f = v`, and the compound
`c[i].f += v` — dispatches every subscript the target projects through to
`IndexRefMut`, so `o[i][j].f = v` resolves at both levels. That trait alone: a
write needs the `&mut` the element is reached by, so a container offering only
`IndexRef` (a shared alias) or only `IndexValue` (a copy the write would land
on and discard) is refused with a diagnostic.

A library export whose signature names a resource is also refused. Every Wado
`resource` binds one another interface owns, so the library's own instance type
has no export for the handle to point at.


gfx:
## Review round 2 — response

Six findings, verified against the current tree. Five were real and are fixed at the class rather than the site; one is skipped.

### Fixed

**A nested binder taking a name ends the search for a reference to the outer one.** The lint's exemption and `bind.rs`'s same-scope legality rule share one helper, `expr_references_var`, and it was not scope-aware: in `let x = match *opt { Some(x) => x, None => 0 }` the arm's own `x` counted as a reference to the outer `x`, so a `let` deriving from nothing was admitted *and* waived. `pattern_binds_name` / `condition_references_var` / `condition_binds_name` now stop the walk at the binder that takes the name, applied to match arms, `matches` guards, if/while bodies, tuple comprehensions, and for-of bodies. This also brings the implementation in line with the WEP's own table (`let x = |x| x + 1` is not a reference). Fixture: `shadowed_name_arm_binder_is_not_a_reference.wado`.

The root cause sat one level above the lint: `ast.rs`'s default walk visited a binder's pattern before its source, contradicting `bind.rs`. Fixed there, so every visitor sees scope order.

**A projected write requires `IndexRefMut` on every subscript on the path.** `o[i][j] = v` resolved its receiver before the `&mut` marks went up, so the inner subscript was taken as a value read and `IndexRef`/`IndexValue` satisfied it. `projected_subscripts_of_place` now marks the whole path ahead of the receiver. Fixtures: `index_value_field_assign_error.wado`, plus chained and `IndexAssign`-through-`IndexRefMut` cases in `index_ref_mut_field_assign.wado`.

**An exported library signature is followed through the declarations it names.** A resource reached through a struct field, a variant payload, or a newtype base reached codegen and ICE'd (pre-existing, confirmed by `git stash`). `resource_in_lib_sig` now follows declared members with a cycle guard, and diagnoses all three shapes.

**`shadowed_name` is emitted independently of `--no-unused`.** That flag names the unused lints; `shadowed_name` is waived per binder and per module with `allow` instead.

**One prefetch attempt per `(registry, coordinate, version)` per process** in `wado lsp`, so an offline or unauthorized pull is not retried on every keystroke.

**`Wado.g4`** gained `attribute*` on generic params, params, closure params, and let statements — `mise run check-grammar` was the failing step in `Test (Wado)`.

### Skipped

**Finding 1 (WEP sections).** Writing a Roadmap for landed work would be putting a decision of my own into a WEP, which `docs/CLAUDE.md` reserves for the maintainer.

### Verification

e2e 5140 passed / 0 failed; `wado-cli` integration 346 passed / 0 failed. Corpus survey after the fix: 0 warnings, 0 new compile errors.

gfx:
## Re-review response

Two new findings, both real, both fixed at the class. One was a compiler panic.

### Fixed

**`resource_in_lib_sig` matched declarations by spelling (P0).** A `Type::Named` resolves through its `AstId` to a module-specific `DefId`; the bare name cannot see through an aliased import. Repro on the pre-fix code:

```wado
// other.wado
use { Descriptor } from "wasi:filesystem";
pub struct Holder { d: Descriptor }

// lib.wado
use { Holder as Held } from "./other.wado";
export fn take(h: Held) -> i32 { let _ = h; return 1; }
```

```
thread 'main' panicked at wado-compiler/src/component_model.rs:4041:29:
unresolved CM named type reference `Held` while emitting CM instance
```

— the exact ICE this guard exists to prevent. `declared` and `entered` are now keyed by the declaring item's `AstId`, read out of `Resolutions`. Regression test in `cli.rs`.

One correction to the finding's framing: the same-name-in-two-modules case it describes is not reachable. An earlier guard already rejects it (`library type 'Holder' is defined in more than one module; a library's public types must have distinct names`), for a private declaration as well as a public one. The name-versus-identity claim was right regardless; the alias is where it bites.

**The C-style `for` binder.** Round 2's scope-aware guard covered eight constructs in `bind.rs` and left this one. The initializer binds for the condition, the update and the body, so scanning those for an outer reference reports a false one and `bind_let` accepts a same-scope redeclaration that derives from nothing. Fixture `shadowed_name_for_binder_is_not_a_reference.wado` went red (`expected compile error ... but compilation succeeded`) before the fix.

**The fixture nitpick, taken.** `shadowed_name_arm_binder_is_not_a_reference.wado` asserted the lint warning but not the redeclare error, so it could not catch an incorrectly accepted `let`. Added `shadowed_name_arm_binder_redeclare_error.wado` for that path.

### Distilled alongside

`bind.rs` gained `block_references_var`, `if_references_var` and `match_arms_reference_var` — the scope guards had spread byte-identical copies across the expression and statement walks. `parser.rs`'s attribute-aware statement dispatch no longer repeats its `Let | Reactive` arm.

### Verification

e2e 5144 passed / 0 failed; `wado-cli` integration 347 passed / 0 failed. Corpus survey: 0 warnings, 0 errors.

gfx:
## Response to the latest two findings

Both verified against the current tree. Neither survives; here is why.

### Skipped: "Traverse the generic declaration head"

Not reachable. A user-defined generic type never crosses an export boundary — an earlier guard refuses it first:

```
error: export function `take`: generic type `Wrapper<i32>` has no Component Model
value representation at an export boundary
```

I checked this for the direct signature (`export fn take(w: Wrapper<i32>)`) and through a newtype (`pub type Boxed = Wrapper<i32>; export fn take(b: Boxed)`). Both are refused before `resource_in_lib_sig` sees them.

The only generics that reach the `Type::Generic` arm are the built-in containers — `Option`, `Result`, `List`, `TreeMap` — whose heads declare no fields, payloads, or base to follow. Following `g.id` would add a branch no input can take, which passes CI and guards nothing.

### Skipped: "Required WEP sections" on `wep-2026-01-20-indexing-traits.md`

Same reason as the first round, and `docs/AGENTS.md` is explicit about it:

> Deciding is the human's role, to adopt and to refuse alike. Propose, and record what is still open as a gap; never write a decision of your own into a WEP.

> Demoting a roadmap item to a gap, or promoting a gap, is the human's call.

That WEP's content sits under topical headings (`The four index traits`, `Consequences`) rather than the template's names. Re-sectioning a landed WEP, and authoring a Roadmap for work that is already done, is a call for the maintainer to make.

### Distilled

`Expr::WithHandler` still walked its body inline after the earlier hoist, and the C-style `for` arm had grown four levels of nested booleans — both go through the named helpers now, `for_references_var` in the early-return style its neighbours use. The library declaration map reads `lib_type_decl_name` as the predicate the rest of the file reads it as.

Swept by shape for both classes this branch touched: block-statement walks and name-keyed declaration maps. The remaining `stmts.iter().any(` hits are TIR/NIR walkers over other IRs, not copies; no other name-keyed declaration map exists.

e2e 5144 passed / 0 failed; `wado-cli` integration 347 passed / 0 failed.

gfx:
## Response: projected index write on the `IndexAssign` path

Real, and fixed. Confirmed by repro — this compiled cleanly before the fix and wrote through a shared `&Inner`:

```wado
impl IndexRef<i32> for Outer { type Output = Inner; fn index_ref(&self, _i: i32) -> &Self::Output { return &self.inner; } }
impl IndexAssign<i32> for Inner { type Output = i32; fn index_assign(&mut self, _i: i32, v: Self::Output) { self.n = v; } }

let mut o = Outer { inner: Inner { n: 1 } };
o[0][1] = 7;   // accepted, though `o[0] = v` is refused
```

The `IndexAssign` target path returns before the refusal the general assign path already applies, so an inner subscript resolving to `IndexRef` or `IndexValue` was never checked.

I did not apply the proposed patch as written. The general path at the bottom of the same function already holds that exact block, so adding it verbatim would have left two copies of one rule to drift apart. The whole mark / resolve / unmark / refuse sequence is now `resolve_place_receiver`, and both paths call it.

Fixture: `index_assign_through_shared_index_error.wado`.

### Verification

e2e at all five optimization levels (`WADO_FULL_TEST=1`): 12912 passed / 0 failed. `wado-cli` integration: 347 passed / 0 failed.

### Also on this push series

`Test (Wado)` was failing on `check-highlight`, and that one was mine. `Wado.highlights.scm` paints `(genericParam (IDENTIFIER)) @type` outright, and states the invariant that licenses it: *"Only a rule whose whole subtree is of one nature can carry an override."* Admitting `attribute*` into `genericParam` broke it, so `#[allow(shadowed_name)]` inside a type-parameter list had its own names painted as types where the compiler calls them plain names. The attributes attach at `genericParams` now — same accepted token sequences, under a rule that carries no override. `check-highlight` is green and `Test (Wado)` passes.

The earlier `ubuntu-latest` failure was a flake; it passed on retry, and macos and windows passed that same E2E step throughout.

PR #2063, opened 2026-09-15
fix(compiler): a global initializer is a function body, and every pass walks it

gfx:
A global whose initializer is not a Wasm constant gets its own function at reify, so template expansion, effect-dispatch desugaring, CM import binding and monomorphization reach it the same way they reach any body. A body reachable only through `TirGlobal` is a body each of those passes misses, and each miss is its own bug.

Purity is checked on what the position holds rather than on the syntax that reaches it, so an initializer may install a handler and dispatch a user-defined effect's operation, and a namespace-qualified dispatch resolves wherever it is written.

## An initializer is a body

Reify mints `$init$<NAME>`, a parameterless function returning the declared value; the global's slot holds a placeholder and says so — asking a deferred global for its declared value answers "assigned elsewhere". Lowering splices those functions into `$initialize_module` in dependency order and drops them, and asserts that a `Direct` global is still a Wasm constant when it gets there.

`TirGlobal` carries no locals, so template expansion and settled-pack expansion lose their per-global arms, and pre-lower reachability roots a `$init$` function by name rather than walking a slot expression for callees. `TirFunction`'s three fields describing one frame move behind `take_frame` / `set_frame`, which replace them together: a caller swapping a body cannot leave one of them describing the body it replaced.

## Purity is a property of the position

One checker answers for both positions that must be pure — a default-value expression and a global initializer — and its diagnostic names which. A position holds the effects its own `with E => h do` installs, so this compiles:

```wado
global COUNTED: i32 = hold: {
    let mut tally = Tally { value: 10 };
    with Counter => &mut tally do {
        break hold: Counter::next()
    }
    break hold: 0
};
```

A closure literal's body is read where it is written, as it is inside a function: the `fn() with E` a global's annotation gives it grants its body nothing.

Effect-parameter resolution runs before either walk reads a callee's effects, so an `effect E` bound to a concrete effect at the call site is seen by both.

## A dispatch is not an effect the position holds

A user-defined effect's operation is answered by an installed handler and traps where none is — a runtime outcome, not a demand on the position, and the same outcome in an initializer as in a function body. So `global COUNTED: i32 = Counter::next();` compiles and traps at module init.

The alternative rule holds only for a dispatch written directly in the initializer: a signature carries no record that a function dispatches an operation, so the same dispatch one call away is invisible to it. An operation backed by the host, or by a component that reaches the host, still demands a capability the position must already hold, and an initializer holds none.

## A dispatch path names its interface by the segment before the operation

`[ns::]E::op` resolves `E` at `segments.len() - 2`, so `ns::Counter::next()` reaches the same declaration `Counter::next()` does instead of being read as a static method on the namespace. Signature resolution, the effect check, dispatch and WIR all key on the declaration's own name and the bare operation, so neither an import alias nor a namespace qualifier reaches them. `IdentExpr::owner_segment` is that segment, named once for the nine sites across the elaborator and the effect checker that ask for it.

https://claude.ai/code/session_01Led2QMooHdkwZoLNJtsje4

---
_Generated by [Claude Code](https://claude.ai/code/session_01Led2QMooHdkwZoLNJtsje4)_


gfx:
All five findings were real, and each is fixed. Three of them fell out of two classes, and the classes are what got closed.

**A qualified path's owner read as exactly two segments.** `lookup_function_param_defaults` matched `[receiver, method]`, so a namespaced dispatch found no declaration and returned empty defaults while the signature lookup, reading `owner_segment()`, reported the full parameter count — an argument-count error on a call that should compile. Reproduced: `s::Stepper::step()` errored `expected 1 arguments, found 0` where `Stepper::step()` returns `1`. Fixed, with `global_init_namespaced_operation_default.wado` covering it. A sweep for the shape across the tree turned up one more hand-spelled owner index, in `resolve.rs`; both now read `owner_index()` / `owner_segment()`, and no site spells the arithmetic itself.

**Two walkers spelling one question apart.** The tagged-template gap was one instance of `PurityWalker` re-implementing `SemEffectWalker`'s call-shape coverage in a second match. The other instance, which the review did not reach, is an indirect call: `global X: i32 = f()` where `f: fn() -> i32 with Stdout` also compiled. Both walks now read a call through one `call_site_effects`, so a spelling either answers for is a spelling both answer for. `global_init_call_shapes_error.wado` and two tests in `purity_sem.rs` cover them.

**`Impurity::Dispatch`.** Right that it had no producer — but wiring it in as suggested would have misreported: its text said the operation "needs an installed handler", which is the rule this PR removes. An unhandled user-defined dispatch is legal now and traps. What `flag_if_operation` still flags is a host-backed operation, so the variant says that instead, and `global_init_host_dispatch_error.wado` pins the wording.

**The WEP naming two deciding phases.** Real. Reify settles `Direct`; the classifier on the lowered Wasm value settles what reify deferred. Both sections now say so, once.

**The handler-scope fixture.** Right that `global_init_handler.wado` proved less than its comment claimed — `AGAIN` installs its own handler, so it passes on a leaked one too. `global_init_handler_scope_traps.wado` puts a bare `Counter::next()` initializer after a handled one and requires the trap, which a leak would prevent.

Nothing was skipped. `/distill` then ran over the branch: two `operation_requirement` calls that always extended by nothing are gone, `assoc_const_owner_segment` was a re-spelling of `owner_segment` whose two fallbacks hid an invariant now stated as one, and `dispatches_at` is one free function for both walks.

---

PR #2065, opened 2026-09-15
feat(lang): branchless comparison chains, and the bounds checks they unblock

### LINKED ISSUE #1856: BCE / fill-collapse does not see through a chained comparison in an assert
`List`'s index traits assert `index < self.used`. Writing the same check as `0 <= index < self.used` — the form that also covers a negative index — loses the `array.fill` collapse in `array_bounds_elim_version_fill_wir.wado`:

| `IndexAssign::index_assign` asserts | `builtin::array_fill<i32>(` in the O2 WIR |
| --- | --- |
| `index < self.used` | 1 |
| `0 <= index < self.used` | 0 |

The fixture's own comment states the shape it pins: the loop is versioned on the runtime residual `limit < used`, the fast arm drops the assert, and the constant-fill loop collapses to a single `array.fill`. With the chained form the versioning no longer reaches that collapse.

The condition is equivalent for the elimination's purposes — a negative index traps on the Wasm bounds check either way, so the extra comparison narrows nothing the optimizer needs — but the recognizer appears to match the single-comparison shape only.

This is separate from #1855, which is the same source form losing its power-assert operand values. Both together made WEP-2026-06-02 Phase D revert the lower bound on the index traits: it cost `index: 5` in the diagnostic and the fill collapse in the WIR, and bought a message rather than any safety. The lower bound is worth restoring once a chained comparison costs neither.

### LINKED ISSUE #2064: BCE does not fire for a guard on a differently-associated affine expression, and four `array_bounds_elim_*` fixtures pass vacuously
## What

Bounds-check elimination does not see that `(pos + 1) + 2` and `pos + 3` denote the same value, so a guard proving one does not discharge a check on the other. Four fixtures that exist to pin this all pass without the optimization happening, because their probe cannot observe it.

Both halves are pre-existing: reproduced identically on `main` (`953ec307a`) and on a feature branch, built side by side in separate worktrees.

## The optimizer gap

`wado-compiler/tests/fixtures/array_bounds_elim_offset_chain.wado`:

```wado
fn pick(arr: List<i32>, pos: i32) -> i32 {
    if pos + 3 >= arr.len() {
        return 0;
    }
    let p = pos + 1;
    let q = p + 2;
    return arr[q];
}
```

The early exit establishes `pos + 3 < arr.len()`, and `q == pos + 3`, so `arr[q]` needs no check. `wado dump --wir -O2` still emits one:

```
if @unlikely(q_19 >= 5) {
    cold_path;
    "…List<i32>^…IndexValue<i32>::index_value$cold0"(q_19, 5);
}
```

The fixture's own comment states the intended behaviour: "The guard-fact decomposition must sum nested `Add` layers (`vn(q) = ((pos + 1) + 2)`) to match `pos + 3`, so the check on `arr[q]` is eliminated at O2." That summing does not happen — `optimize/condition_implication.rs` compares the guard's expression against the index expression without normalizing either to an affine form.

Related but distinct: #1856 (BCE not seeing through a chained comparison in an `assert`).

## Why the fixtures do not catch it

All four probe with:

```json
"wir_not_expect:O2": ["index out of bounds"]
```

The panic message is not built at the check. It is built inside `index_value$cold0`, from `$const_obj_*` globals holding the text as `array.new_fixed<u8>` byte sequences. The literal string `index out of bounds` therefore never appears in the WIR dump whether or not the check was eliminated, and the assertion holds by construction.

Affected:

- `array_bounds_elim_offset_chain.wado`
- `array_bounds_elim_loop_guard.wado`
- `array_bounds_elim_loop_guard_wir.wado`
- `array_bounds_elim_le_guard_wir.wado`

Each still emits `$cold0` bounds-panic calls at `-O2` (6, 6, 3 and 3 occurrences respectively). The two `*_wir` fixtures also carry `wir_expect` patterns, but those pin the array literal and the `array_get_value` call, not the absence of the check.

## Suggested probe

Assert on what the eliminated check actually leaves behind, scoped to the function under test:

```json
"wir_scope": "array_bounds_elim_offset_chain.wado/pick",
"wir_not_expect:O2": ["$cold0", "@unlikely"]
```

`wir_scope` keeps a stdlib cold path elsewhere in the program from failing the probe. Fixing the probes first will turn all four red, which is the point — they should stay red until the optimizer change lands.

gfx:
Bounds checks that a loop already proves are gone at `-O2`, and `List` index assignment reports a negative index by name instead of trapping bare inside `array.set`. A constant-fill loop collapses to one `array.fill` whether or not another loop encloses it.

A comparison chain is now a branchless range test: `a < b < c` evaluates every operand once, left to right, and joins the comparisons with `&`.

Closes #2064
Closes #1856

## Comparison chains

A chain is a range test, so it evaluates all of its terms. Skipping an operand on one path is what `&&` is for, and the spec, the cheatsheet and the operator-precedence WEP say so.

This also settles an inconsistency the join created: a chain binds its middle terms ahead of the test, so `a < b < c < d` reached `c` on every run and `d` only on some.

Power-assert follows the rule. A chain marks no operand conditional, so a failed `0 <= i < n` reports `n` beside `i`, and the condition needs neither a capture temp nor its `$_seen` flag:

```
condition: 0 <= index < self.used
index: -1
self.used: 4
```

## Bounds-check elimination

`condition_implication` gains three things:

- A bound may be a constant, so a folded `arr.len()` over a literal list matches a guard like any local or field does. This is what #2064 reported: four `array_bounds_elim_*` fixtures passed while eliminating nothing, and their probes now count the outlined cold path instead of a panic string.
- A two-sided check is a conjunction, and dies only when every conjunct holds. The guard answers the upper half; the entry constant of a `+ 1` counting loop floors the lower one. `&&`, a bool `&`, the `if a { b } else { false }` a captured operand turns `&&` into, and the pooled `Select` of the promoted form are all read as the same shape.
- A constant-upper-bound guard refutes a check directly.

Two facts the pass reads are now read once each, with their type: a pooled integer consults `is_unsigned_int` before being taken as `i64`, and a write counts as a fact only where it is unconditional, so `if flag { i = 0; }` supplies no entry constant.

## The fill idiom

`loop_version_bce` replays whatever the loop body left beside the store, once, at the last iterated index, instead of classifying it into two recognized shapes. Those statements must write only locals and not trap; a write a branch guards is allowed only where the slow arm alone names the local and the version `if` runs once.

Casts stop being refused on their kind. `cast_truncates_a_float` is one predicate in `arena_query` now, where `mod_ref`, `expr_node_may_trap` and select lowering each kept their own and disagreed about whether a type table was available.

## Note

`kiln_generator_world::generator_using_core_log_adds_no_wasi_import` fails on this branch and on `main` alike, for a reason unrelated to it.


gfx:
Merged `main`, fixed the red job, and answered the review. Three commits, two classes and one skip.

## The red job: `kiln_generator_world::generator_using_core_log_adds_no_wasi_import`

Not core:log's defect, and not this branch's: the test compared core:log's WASI imports against a noop generator's, which held only while the noop one also carried `wasi:cli/types`. It no longer does — that otherwise-dead type-only import is dropped — so an equality that was never the guarantee failed on a component the Kiln linker instantiates fine.

What the test guards is that the linker is never asked for a WASI *function*, so it now reads the import plan's kinds instead of a baseline: a `SharedTypes` entry asks nothing of instantiation, anything else does. The noop generator must plan no WASI import at all, core:log's must be non-empty and every entry `SharedTypes` — so the check cannot pass by seeing nothing.

## Class: a constant `false` read in only one of its two spellings

A bool literal pools as `ValueKind::Bool(false)`, which no integer parse reads, so all three zero checks in `condition_implication` — an `if`'s else arm, a `Select`'s else arm, and `x == false` — missed every conjunction whose `false` came from source. The reported site was one instance; the other two were the same defect. Measured before and after on

```wado
let ok = if 0 <= i { i < xs.len() } else { false };
if !ok { panic("index out of range"); }
s += xs[i];
```

which kept `select bool(0 <= i, i < used, 0)` and its cold path at O2, and now compiles to the bare `array_get_value` with the guard gone. One `is_const_false` answers all three checks; `array_bounds_elim_bool_false_conjunction` pins it.

The compiler's own capture lowering was never affected — it builds an integer zero, which is why every existing fixture passed while this stood.

## Class: `&&` in prose the code no longer means

The chain expands to `&`. Five places still said `&&`: the desugar's doc comment and its `&&`-joining comment, the `DesugarKind::ComparisonChain` case, the unparser's parenthesis rule, and the pipeline WEP. The precedence WEP also credited Python for behaviour Python does not have — its chains short-circuit — in three places, now worded as syntax borrowed, evaluation not.

## Skipped: a Roadmap section in the power-assert WEP

`docs/AGENTS.md` settles this one: "Demoting a roadmap item to a gap, or promoting a gap, is the human's call." Every unfinished item in that WEP is explicitly unpinned, so writing them into a Roadmap would commit work on @gfx's behalf. The `mise run format` note needed no content change; markdown format is clean.

Verified locally on the merged tree: `integration` 780 passed, `e2e` 5210 passed (O0+O2), clippy and `cargo fmt` clean.

---

gfx:
Round 2. The affine-offset finding was real and it was a compiler crash, not a missed optimization.

## Fixed: the optimizer's own sum overflowed, aborting the compile

`wado-compiler` keeps overflow checks in the dev profile — `[profile.dev.package."*"]` turns them off for dependencies only — which is the profile every test and CI job builds. So the reviewer's "can panic" is what happens:

```
thread 'main' panicked at wado-compiler/src/optimize/condition_implication.rs:352:37:
attempt to add with overflow
   3: wado_compiler::optimize::condition_implication::parse_var_offset_depth
```

from `(i + 9223372036854775807) + 9223372036854775807 < n` on `i64`. Reading an index as `var + const` sums the constants it walks through, and two literals near the maximum leave `i64`. All four sites now sum with `checked_add(c)?` — forgoing the match is what the optimizer's own arithmetic owes, and `classify_conjunct` already did it that way. `array_bounds_elim_offset_overflow` pins it. Pre-existing on `main`, and swept: no unchecked offset arithmetic remains in either pass.

## Fixed: `&` binds tighter than a comparison, so my own wording was wrong

The pipeline WEP read `a < b < c` → `a < b & b < c`, which parses as `a < (b & b) < c` under the table this PR's own precedence WEP sets (level 8 for `&`, 10 for comparison). Both it and the unparser's comment now write `(a < b) & (b < c)`. Good catch.

## Fixed: a chain of `==` is not a range test

`a == b == c` is an accepted chain and no range. The rule now reads "a chain is one test — a range test where the operators are inequalities".

## Skipped: the `spec.md` initializer wording

The finding quotes the first paragraph. The next one already resolves it: "A user-defined effect's operation is answered by an installed handler and traps where none is, in an initializer as in a function body, so an initializer may dispatch one. **It may also install its own handler.**" Nothing is contradictory once both are read, so there is nothing to clarify.

## Skipped: reverting `kiln_generator_world.rs` as out of scope

That change is the fix for the red `Test (Core)` job — the test compared core:log's WASI imports against a noop generator's, and the noop one no longer carries the type-only `wasi:cli/types`. Reverting it puts the required check back to red. The reviewer is right that it is unrelated to #2064 and #1856; it is related to the branch being mergeable, which is why it stays.

## Skipped again: a Roadmap section in the power-assert WEP

`docs/AGENTS.md`: "Demoting a roadmap item to a gap, or promoting a gap, is the human's call." Every unfinished item there is explicitly unpinned. What I could do without deciding for @gfx, I did — that WEP now records two things this branch makes true: a chain's middle operand is reported once because it is evaluated once, and, as a new gap, that a chain reports no per-comparison result, so a failed `0 <= i < n` does not say which half failed where `0 <= i && i < n` renders `0 <= i`. Both measured with `wado dump --assert-plan`.

Verified on the merged tree: 52 bounds-check fixtures green at O0 and O2 including both new ones, `integration` 780 passed, clippy and `cargo fmt` clean. A full e2e pass over all 5210 fixtures was green on the previous commit and is running again on this one.

---

gfx:
Round 3 answered on `d2723016a`. The Major finding is real — it reproduces as wrong code, in two guard shapes rather than one. The Minor one I skipped, with a reason.

## Fixed: a capture block peeled as if it were a fresh read

`capture_block_binding` read any `LabeledBlock` whose tail is a local written once unconditionally as standing for that write's operand. A block holds a **snapshot**, so once the block's own statements write what the operand reads, the two say different things — and a guard parsed through the peel proves nothing about a check below it.

Reproduced as a dropped bounds check: O0 traps, O2 runs on.

```wado
let mut w = Window { limit: builtin::black_box(3) };
let i = builtin::black_box(2);
if i < save: { let saved = w.limit; w.limit = builtin::black_box(1); saved } {
    if i >= w.limit {                  // 2 >= 1 holds; the panic was dropped at O2
        builtin::cold_path();
        panic("index out of bounds");
    }
    println("bound survived its own guard");
}
```

The early-exit guard (`if i >= <block> { return; }`) was unsound the same way, so the gate went where the claim is made — `capture_block_binding` itself — and covers every consumer at once: `parse_bound`, and `loop_version_bce`'s `read_local` and plan scan. Both shapes were checked by hand at O0 and O2 before and after.

Pinned by `array_bounds_elim_capture_snapshot_mutated.wado`: without the fix it fails with `trapped mismatch: expected true, got false` at O2, with it, green at both levels.

Regenerating every WIR golden changed no existing file, so the gate costs no elimination anywhere in the corpus.

## Skipped: `**bold**` for the new Known-gaps entry

`docs/CLAUDE.md` forbids bold **for sub-sections**. The flagged line is a checklist item's label, and every entry in that section is written that way (`- [ ] **Capture a receiver that is not a place.**`, `- [ ] **Render a WithHandler or Resume operand.**`). The sub-sections themselves are `###` headings, as the rule requires, so the new entry follows the convention rather than breaking it.

## Noted, not reproduced: the same shape one function over

`struct_field_init` projects a struct field's *initializer operand* out of its construction site, and `bound_offset_over` then reads it as `gbound + c` against the **current** `gbound`. Were `gbound` reassigned between the literal and the check, that equality would be stale — the same snapshot-versus-fresh-read class. I could not construct it. Four attempts, including the live `List::<i32>::filled(limit + 1, 1)` shape from `array_bounds_elim_le_guard_wir.wado` with the guard local mutated before the loop, all keep the check at O2, and nothing inside this pass is what stops them. Recording it here rather than tightening the projection on a hunch, which would cost the optimization `bound_offset_over` exists for.

## Also in this push

The one witness-to-root dispatch behind `node_modifies` is now `modified_root`, which is what the new gate asks its question through (−45 lines of branching), and `operand_mentions_local` joins the two-worlds readers in `arena_query`. `/distill` over the whole branch additionally put `loop_version_bce`'s five inline `Binary` allocations behind one `alloc_binary` and rewrote the comparison-chain prose in `spec.md` and the precedence WEP to drop a forward reference and two clause-hung dashes.

Verified locally on the merged tree: e2e 5214 passed / 0 failed (O0+O2), integration 780 passed / 0 failed, clippy clean, `mise run format` clean.

---

gfx:
Round 4 (the two 🟡 on `9de55fae5`) — both real, both fixed in `8e5ec0a8f`.

- **"Call all chains comparison tests."** Correct: `a == b == c` is a valid chain and no range test, so rule 5 now reads "A chain is one test: a range test where the operators are inequalities, an all-equal test where they are `==`."
- **"Show once-only chain lowering."** Correct for an entry titled *expansion*: `(a < b) & (b < c)` names `b` twice where a chain reads it once. The pipeline WEP now says the chain **tests** that, each operand evaluated once, so it reads as the equivalence it is rather than as the lowering. `spec.md` already said "Equivalent to" and states the evaluation rule under the examples, so it needed nothing.

The `Out of Scope Changes` pre-merge warning is the same ask as round 2 and the answer has not changed: `kiln_generator_world.rs` is what makes `Test (Core)` pass. Its `generator_using_core_log_adds_no_wasi_import` asserted equality against a noop-generator baseline, and that premise went stale when the noop generator stopped carrying the type-only `wasi:cli/types` import — `left: ["wasi:cli/types@0.3.0"] right: []`. The test now asserts what it was written to assert, that `core:log` adds no *function-bearing* WASI import the Kiln linker cannot provide. Reverting it re-breaks a required check, which is not a trade this PR can make.

Also in flight since your review: the Major finding from round 3 reproduced as a wrong-code miscompile and is fixed in `47f18c278` — see [the round-3 reply](https://github.com/wado-lang/wado/pull/2065#issuecomment-5689356422). Your `9de55` risk note predates that commit.

---

gfx:
`Test (Wado)` was red on `d2723016a`, and it was this PR's: `check-grammar` found a grammar gap my new fixture walked into.

```
error: Wado.g4 rejects 1 file(s) the compiler parses. Every one is a grammar gap:
  wado-compiler/tests/fixtures/array_bounds_elim_capture_snapshot_mutated.wado:9: unexpected input "export"
```

The fixture guards the capture-block peel, so its condition reads a labeled block: `if i < save: { let saved = w.limit; … } { … }`. `Wado.g4` left `labeledBlock` out of `primaryNoStruct`, the rule that keeps a brace-terminated expression out of condition position, so the grammar rejected what the compiler accepts. The compiler takes one in an `if`, a `while`, and a C-style `for` head alike; all three were checked.

Fixed in `6082b7cc3` by admitting `labeledBlock` there. It is not the ambiguity that rule exists for — the `:` before the brace already says the brace opens the block, not the body that follows. Parenthesizing the fixture would have silenced the check while leaving the gap, so the grammar is what changed.

- `mise run check-grammar`: 3091 files, **grammar gaps 0** (was 1)
- `wado test package-gale-highlight-wado`: 62 passed, 0 failed
- the fixture still fails without the optimizer fix (`trapped mismatch: expected true, got false` at O2) and passes with it

---

PR #2067, opened 2026-09-16
feat(core:icu): Unicode character properties, at 104 KB for a program that uses one

gfx:
`core:icu` answers Unicode character properties, as code point ranges or per character, from ICU4X 2.3 bundled as a component asset. A binary property is named by an enum case, so `is_upper` is an integer across the boundary and a borrow of a baked static on the other side — no string copy, no table search. Gale's `\p{...}` and its POSIX classes come from it, which retires the hand-rolled tables and the oracle that pinned them. A program reaching one property carries 104 KB of the 303 KB asset.

## The surface

```wado
use icu from "core:icu";

icu::has(BinaryProperty::Uppercase, 'A');            // bool
icu::set(BinaryProperty::Alphabetic);                // List<[u32, u32]>
icu::ranges("General_Category", "Lu");               // Option<List<[u32, u32]>>
icu::contains("Script", "Greek", 'α');               // Option<bool>
```

A property is named two ways because programs name it two ways. A classification
site knows which property it wants when it is written, and the enum makes that a
constant. A `\p{...}` in a grammar does not, so the string path stays for
whatever the source spells.

## Collecting a component asset

The asset is one component holding one core module. Each core module it holds is
embedded and every other section is copied byte for byte, so nothing is
renumbered and each `alias core export` still resolves. The root set is the
lifted functions the program imports; the rest keep their name and signature as
a stub.

Under that, `wado-wasm-embed` reads `reloc.DATA` — a pointer stored in the data
itself, reaching another data range or a function. A live data range can root a
function, so the data and code edges close over one worklist. ICU has 146 such
pointers; libm has none.

The data-reference map is baked at asset-build time, as libm's is. That is what
keeps the `.wat` form: the map's offsets are segment-relative, so a
print/reparse preserves them where a `reloc` entry's offset into the encoded
code section would not.

## Also here

- `wado doc` follows a `pub use` into a non-Wado source, so a CM type re-exported from a bundled asset renders.
- Gale's Unicode code lives in `package-gale/src/g4/unicode.wado`.
- `scripts/replace-if-changed.sh` is the compare-and-move both bundled-asset tasks end with.


gfx:
Answered all four findings. Three are fixed; the fourth is real and I could not close it, so it needs a decision.

## Fixed

**`DataRange::at` bypassed the checked constructor** (`wado-wasm-embed/src/dataref.rs`) — real, and `new`'s own doc states the invariant it broke: "every constructor goes through here, so `end` is total". `at` now builds through `new` and `resolve_map` turns `None` into `Error::DataRef`. A map spelling `@0:4294967295` is reported rather than panicking. Pinned beside the existing overflow cases in `a_range_that_cannot_end_is_rejected`.

**`strip_action_bodies` had no non-ASCII test** — added. It is the classifier's second consumer and the one whose failure is silent: a lexer rule read as a parser rule scans `~[/[]` as a nesting argument list and swallows every action after it. `ÀBC : ~[/[] ; r : 'a' { act(); } ;` now pins it.

**`Ⅰbc` (U+2160)** — added to the parser test. It is `Nl` with `Other_Uppercase`, so it separates `Uppercase` from `Lu` where the existing `À` case cannot.

**`TODO.md:77`** — the finding asked for better wording; the stated blocker was "Wado's prelude has no `char::is_uppercase`", which is what this branch removes. So `Antlr4Base::emit` now calls `is_rule_name_upper`, the `#[TODO]` on `a non-ASCII uppercase rule name is a TOKEN_REF` is resolved and gone, and the TODO entry says the ANTLR4 port's remaining blocker is the Java twin alone. The TypeScript half of that sentence is unchanged.

## Real, not fixed: unused world-level lifts

`collect_component_asset` does leave an unwired world-level lift as a root. I implemented the fix and it stubbed nothing, for a reason above the collector:

- `wir_build/component_imports.rs:210` pushes an `ImportKind::WorldFunction` entry for every world function the registry knows, not the ones the program calls, so the "used" set is always the full set.
- `used_wasi_functions` is keyed by effect usage (`interface::op`), and a bare world function is not an effect operation, so it has no entry there either.

No phase records which world functions a program reaches. The collection mechanism itself is fine — `embed_component` on a two-lift fixture cuts 274 bytes to 222 — but nothing can tell it which lift to stub. I reverted rather than leave code that cannot fire alongside a test that passes by construction.

Closing it means giving a world-level call a usage record the way an effect operation has one, which changes what the import plan means. That is a phase contract, so it is a decision rather than a fix. The size at stake is zero for every asset in the tree today: ICU exports interfaces only, and both `hlwf` fixtures export one world function.

## Verification

`cargo test -p wado-wasm-embed` 46 passed · `wado test package-gale` 3159 passed, 0 failed, 15 todo (down one, the resolved `#[TODO]`) · `mise run format` clean.

gfx:
Both findings verified against the current tree. Neither survives, so nothing changed.

## Skipped: the bold label under Known gaps

`docs/CLAUDE.md` bans bold **standing in for** a heading: "Do not use `**...**` (bold) for sub-sections. Use markdown sections instead." What is at line 357 is a lead-in inside a list item, not a section.

Its two siblings in the same list are written the same way — "**The collection stops at the export.**" and "**A dropped lift leaves a stub, not a hole.**" — and 49 WEPs use the form. Promoting one of three bullets to `###` would leave a heading interleaved with two bullets in one list, which reads worse and matches nothing else in the file.

## Skipped: the stripper regression's location

The rule is "TDD every g4 change with a unit test in `src/g4/{lexer,parser}_test.wado`", and it governs changes to the `.g4` lexer and parser. This change is to `action_strip.wado`, whose 23 tests all live in `src/g4/action_strip_test.wado` — already under `src/g4/`, beside the code it covers.

Copying the same `strip_action_bodies` call into `parser_test.wado` would assert exactly what `action_strip_test.wado:85` asserts, against the same input, through the same function. A duplicate catches nothing the original does not, and it reads as coverage the suite does not actually gain.

The classification the two files share is covered on both sides already: `parser_test.wado` pins what `parse` does with `ÀBC` and `Ⅰbc`, and `action_strip_test.wado` pins what the stripper does with `ÀBC : ~[/[]`. Those are different failures — the second is the silent one, where a misread rule swallows the actions after it.

PR #2069, opened 2026-09-16
feat(benchmark): a GPT that trains in Wado, and the suite's first object-graph workload

gfx:
`example/microgpt.wado` trains a character-level GPT on a list of names and
samples new ones from it, in pure Wado with no dependencies — a port of
[Andrej Karpathy's microgpt](https://gist.github.com/karpathy/8627fe009c40f57531cb18360106ce95).
`benchmark/microgpt/` runs the same program in Wado, JavaScript and Rust, and
joins the continuous benchmark at `-O1`, `-O2` and `-O3`.

Every other row in the suite's pure-computation section loops over a flat array.
This one walks a graph of small heap objects: a training step builds 31k to 89k
nodes, traverses them depth-first, and accumulates gradients back through them —
the shape where a GC'd language's costs show. Wado runs `train` at 3.40x the
best arm and `infer` at 2.67x.

## Two phases

`train` is forward, backward and Adam over 32 documents. `infer` is the forward
path alone, over the full 16-position attention window, deeper than any training
step reaches on this corpus, where the longest name gives 11 positions.

The two separate two costs. Training spends most of its time in the backward
pass, which sorts the whole graph topologically and then walks every edge again.
Inference never builds that traversal, which is why Wado gives up less there.

Both phases reseed, so all three arms print the same final loss and the same
generated sample. A divergence between them is a bug, not noise.

## Three node representations

Wado's `Graph::value` pushes a node and returns a `&mut Value` into the list,
and a node holds those handles as its children — the shape the Python original
has. Rust cannot say that: a `&mut` into a `Vec` dies at the next push, so its
arm makes a node a `usize` index and every read spells out `self.nodes[i]`.
JavaScript matches Wado's shape. The three files are otherwise the same program,
which is what makes the gap between the rows readable.

`Value.id` is the node's slot in `Graph.nodes`, so `build_topo`'s visited set is
a dense `List<bool>`. Python keys that set on object identity; Wado has no
identity hash, and the arena supplies one.

## Also here

- count-prime leaves the benchmark suite to make room. Its Wado arm lives on as
  `example/count_prime.wado`; integer arithmetic is what `mandelbrot` and
  `sieve` already measure.
- `wado-compiler/src/link.rs` drops an unused `hashmap` import that a release
  build warned on.

---

https://claude.ai/code/session_01F4wnLD1VfHCbaF3msZxe87

---
_Generated by [Claude Code](https://claude.ai/code/session_01F4wnLD1VfHCbaF3msZxe87)_


gfx:
The blank-only corpus is fixed in 49a856f, and the "may still abort" line in the CodeRabbit summary is stale — its own re-review of that commit reports no actionable comments.

Reproduced before the fix, on an `input.txt` of four blank lines:

```
num docs: 0
Caused by:
    wasm trap: integer divide by zero
```

`train` reached `docs[step % docs.len()]` with an empty list, because a blank-only file reads successfully and the `Err` fallback never ran.

Same input against the committed file, exit 0:

```
microgpt: input.txt holds no documents; falling back to the built-in names
num docs: 256
step    1 /    2 | loss 3.3139
```

The fallback keys on the parsed document count rather than on the read succeeding, and the line parsing is shared so the file text and the embedded names go through one path.

---

PR #2077, opened 2026-09-17
perf(value-copy): key a callee's writes by the handle each is reached through, 467x on json_catalog_v2

gfx:
`value_copy` keys a callee's writes by which of its parameters each is reached
through, so a caller learns what the handle it passed can observe rather than
what some value of that type somewhere was written. A parser holding an
`Array<u8>` input buffer keeps reading fields off it across calls that allocate
`Array<u8>`s of their own, and `json_catalog_v2`'s `parse_area` measures
413 MB/s on this machine.

The rate a bare `TypeId` key bought was 1.2 MB/s: a formatter's scratch buffer
under `panic` and the parser's input are both `Array<u8>`, so every call that
could reach integer formatting reported a write to the parser's buffer, no
write was re-rootable at the handle, and `let raw = self.repr` deep-copied the
whole input on each call.

## Origin tagging

A write is `(Origin, TypeId, field)`, where `Origin` is the parameter position
the write is reached through, or `Unknown` for a handle the walk cannot trace
to one. `Writes::re_rootable_at(position, owner)` and
`Writes::fields_of(position, owner)` both ask `Origin::could_be(position)`, so
a gate and the extractor behind it cannot disagree over `Unknown`: a write the
gate lets through is one the caller records.

## Dropping what stays in the frame

At a call, each parameter position is classified by what the caller put there.
A position filled with storage the caller allocated is `FrameLocal`, and the
callee's writes through it are dropped on absorption — no caller of this frame
holds a handle that reaches them. A position filled with a handle the caller
was itself lent carries the callee's writes out, re-tagged to that handle.

Classification is positive: only a named place this walk resolved is dropped.
An argument the walk cannot classify, and a parameter position it did not
reach, both read as `Unknown` and keep their writes. A `Value` naming no place
still aliases its operand where `builtin::select` returned it, and
`lends_storage` admits a `&T` position that `could_write_through` does not, so
the `handed` map is built over every argument.

## Fixtures

`value_copy_callee_frame_local_write` asserts a parser whose callee allocates
its own `List<i64>` emits no `$array_clone_dst`.
`value_copy_callee_writes_handle_field` is its control: the callee writes the
very field the caller read before the call, and the copy has to stay.


gfx:
Fixed in d63312576ad.

The finding is real. `Writes::writes_anything()` was a whole-callee predicate, so a projection recorded for one argument activated on a write the callee made through a different one: `callee(&mut outer.a, &mut scratch)` charged `outer.a` for what the callee wrote through `scratch`, the caller reported a write it does not make, and its own caller copied defensively against it. Output was never wrong — the cost was precision.

The class is the one this PR is about: asking about a single handle and answering with a fact about all of them. `writes_anything()` was the last such question left, so it is gone rather than patched. `Writes::writes_through(position)` replaces it and routes through the same `Origin::could_be` as `re_rootable_at` and `fields_of`, and `PendingProjection` carries the callee position its argument fills.

One correction to the suggested patch: the position to compare is the callee's parameter index, not the caller-side `Origin` already stored in `PendingProjection::fields`. Those are different frames — the stored `Origin` says which of *this* function's handles the projected path is rooted at, while the gate needs which of the *callee's* parameters the argument fills. Comparing the stored one would have gated on the wrong frame's numbering.

`value_copy_callee_writes_other_argument` covers it: 4 `$array_clone_dst` before the fix, 0 after, identical program output either way.

PR #2078, opened 2026-09-18
perf(value-copy): a value argument names storage of its own, restoring json_catalog_v2 to 44 clones

gfx:
`value_copy` reads an argument that names a value — a literal, a
construction, a call returning what it owns — as storage of the calling
frame's own, so a callee writing through that position writes nothing a
caller of that frame reaches. json_catalog_v2's `parse_area` keeps its
input buffer across every call instead of deep-copying it, and the module
emits 44 array clones rather than 59.

Treating such an argument as a position the walk could not classify tagged
the callee's writes with a handle that could be any of them. Functions
writing nothing at all then reported a write through every handle their
callers held: `StrSlice::sub` carried one, and it reached each of the
eighty-odd stdlib entries a formatter sits on the path of.

## Classifying every position

`handed_at` is total, so `re_tag` has no position left to guess at. A value
is storage of this frame's own; a name the walk cannot resolve is the one
handle that could be any of them; a type carrying no storage hands the
callee nothing to write through; and a fresh value aliases nothing this
frame was handed, which `is_fresh_value` already answers.

## Where the aliasing case belongs

A core builtin declaring no `#[result(part_of)]` may hand back an operand
rather than a copy — `builtin::select` does — so `names` reports a
storage-carrying result of one as `Names::Unknown`, not as a value of its
own. Every reader of `names` was being told otherwise, and the distinction
between "a value" and "something that may alias" is that function's to
draw rather than each caller's.

## Fixture

`value_copy_callee_value_argument_names_no_handle`: `sub` writes nothing,
yet `take` came away reported as writing its receiver's whole type, and
`scan`'s field read copied the input buffer. One `$array_clone_dst`
without the change, none with it, same output either way.


gfx:
Fixed in 472c4d50707.

The finding is real, and sharper than it reads: `OwnedCalls::is_owned` returns true for *every* `part_of`-less core builtin, so `is_fresh_value` called a storage-carrying `builtin::select` fresh and `handed_at` answered `Unobservable` before `names` was ever asked. The `Names::Unknown` this PR added in `place.rs` was unreachable from that path — which is why that hunk produced no WIR diff on json_catalog_v2, a signal I had noted and not chased.

I removed the freshness early return rather than reordering it. Measuring the three conservatisms separately showed freshness contributed nothing: `carries_storage` alone and `carries_storage` + freshness both left `parse_area` at 15 clones, and the whole recovery to 0 came from reading `Names::Value` as storage of the frame's own. Every argument freshness would have classified is one `names` already calls a value or a place this frame owns, so it was a second answer to a question with one authority — and the wrong answer on exactly the aliasing case. The oracle threading it needed goes with it: the diff is three files now instead of five.

`parse_area` still emits 0 clones and json_catalog_v2 44; e2e is 5350 passed, 0 failed; clippy clean.

PR #2080, opened 2026-09-18
perf(optimizer): devirtualize iterator-adaptor closures — CBOR serialize +46%, microgpt inference +10%

gfx:
An `IndirectCall` whose callee traces back to a single `ClosureToCanonical` is dispatched straight to the functor's `$call`, so the inliner can splice the closure body into its caller. Against the merge base, CBOR serialize gains 46% on twitter and 39% on catalog, and microgpt gains 10% on inference and 8% on training.

## `closure_devirt`

`lower`'s fn-param specializer takes the closure whose value never leaves its declaring local. Every iterator adaptor parks its closure in a struct field instead, which escapes by that rule, and the read of that field only appears once `inline` copies the adaptor's `next` into the caller — so `xs.map(f).collect()` pays a `ref.cast` and an indirect call through a wrapper per element.

The rule runs in the post-inline `peephole` session, where that field read exists. It follows the callee through bindings, borrows, blocks and struct-literal fields on a fixed budget, binds the functor to a local of its own, and has the direct call read that local.

Two conditions bound it. The callee operand must neither write nor trap, because naming the local is what drops it and with it whatever the trace saw through. The binding must run before the call in a block holding both, because the devirtualized call is what makes the functor's original value dead.

## `let_block_flatten` through a reference

A reference around a block-tailed binding applies to the tail's value either way, so `let x = &mut { …; Struct { … } }` — what an inlined by-reference receiver leaves — hoists its leading statements like the unwrapped form. This is what puts the adaptor's functor binding in the block holding the call, which `closure_devirt` requires.

## LICM aliasing

A GC struct held by value is a handle on its object just as `&T` is, so a write through any other handle on it is visible through this one, and a field hoist past such a write is unsound. `written_field_types` and `clobbered_pointee_types` key on the resolved type, since one struct is interned under more than one `TypeId`.

## Names

`name.rs` mints the closure functor struct name, its `FqTypeName`, the mangled `$call`, its `LocalMethodName` and `MethodName`, and the predicate recognising one — per "Only `src/name.rs` knows a name format". `lower`, `nir`, `dce` and `wir_build` name it rather than spelling it.

`nir_visitor::exprs_under` walks one region, and `reachable_exprs` is its body-root case.

## Measurements

`benchmark/README.md` and `wasm-size/README.md` are re-measured whole: best of three `benchmark-all` passes picked per row, `http-routing` at both worker shapes, and `report-wasm-size`.


gfx:
Both findings were real. Fixed, with a note on the class each belongs to and on what neither of them reaches.

## The reload path — fixed, and widened

Real, and P0: a stale hoisted value is wrong code. The claim is precise — `mut_ref_pointee` compared the argument's pointee against `clobber_types` by `TypeId`, so a `&mut T` whose `T` carried another of that struct's ids never registered as a clobber and no reload was inserted.

This is the same defect as the one the branch already fixes for the direct-write case, and the reviewer reached an instance I had not. Patching `mut_ref_pointee` alone would have left the next one standing, so the comparison now has one home: `PointeeSet` (and `PointeeFieldSet` for a type's individual field) stores the resolved type and owns both the insert and the query. Nothing in the pass compares `TypeId`s any more — `clobber_types`, the `hit` set that picks which specs to reload, `clobbered_pointee_types` and `written_field_types` all go through it.

No fixture. The condition is that one struct reaches the pass under two ids, which is a property of interning rather than anything source can ask for; a test asserting it would pass on construction and guard nothing. `opt_licm_closure_captured_write` covers the direct-write half.

## Grammar size — fixed

Real. `gale_gen.wado` sets `input_size` to `RUST_LEXER.len() + RUST_PARSER.len()`, and the two files are 9,685 + 32,185 = 41,870 bytes, which is what the printed throughputs divide by. Corrected.

Swept the rest of the same class while there: `twitter.json` 631,514, `canada.json` 2,251,051, `citm_catalog.json` 1,727,204 and `queries.sql` 13,321 all match their files. The grammar figure was the only wrong one.

## What neither finding reaches

The type table interns one struct under several `TypeId`s — `Box<i32>` appears as both `TypeId(1522)` and `TypeId(2385)` with identical `ResolvedType`. `PointeeSet` makes `licm` immune, but every other type-keyed analysis in the compiler has the same blind spot, and `licm` is only where it happened to surface twice. Canonicalizing interning is a type-table change with tree-wide reach, so it belongs in its own issue rather than in a perf PR.

PR #2081, opened 2026-09-18
refactor(tir): make type identity a checked question, not a TypeId comparison

gfx:
A `TypeId` addresses a slot in the type table, and both newtype erasure and the boxing rewrite leave many slots resolving to one type. `TypeTable::type_key` answers the identity an id cannot, handing back a `TypeKey` whose inner id is private — so a pass keying a collection by type identity is type-checked into doing it right instead of remembering to. Three passes ask through it here, and `TypeTable` loses five APIs.

Every benchmark and every `wasm-size` program compiles byte-identically, `sqlite_highlight` included. Generated code moves in one place: the widened variant-return SROA reaches `serde_newtype_of_struct`, 52 bytes smaller.

## One mechanism for a redefined slot

Two late phases redefine a `TypeId`, and both register a redirect. Boxing goes through `redefine_to`, which points the borrow's slot at its `Box<T>` and retires that borrow's `intern` key, so a later `intern(&T)` mints a live borrow — `sroa_param::FieldForm::of` mints one post-boxing and depends on it. `get_unerased` still reads the borrow off the slot, which is where `is_mut_box` and `fq_type_name_unboxed` read its sharedness.

`retain` rebuilds the intern map from the surviving slots and skips a redirected borrow, so a DCE prune leaves the retirement standing. `redefine_to` asserts that only a borrow is ever redefined, since that shape is what `retain` recognises a retirement by.

`resolved_id` is the single redirect hop `get`, `try_get`, `get_pruned` and `type_key` take, next to the assert that it is only ever one.

Retired: `replace_type`, `push_fresh` (no callers), `register_boxed_ref`, `TypeMap::replace`, and the `boxed_ref_shared` side table whose one bit the slot now carries itself.

## The passes that key by identity

- `licm` keys its pointee sets by `TypeKey` — a hash lookup per query, and no `ResolvedType` clone per insert.
- `value_copy_demote` keys `helper_by_type` by `TypeKey`. Reached by an id the table also spells another way, that map drops a call-graph edge and under-approximates the `leaky` fixpoint.
- `sroa_variant_return::Rebind` compares payload keys and carries the table they were taken from, so the two sides of a comparison cannot come from different ones.

`wado-compiler/AGENTS.md` states the rule: compare and key by `TypeKey`, never by the id.

## Also

`sroa_variant_return` and `wir_build::calls` call `TypeTable::peel_refs` rather than spelling its loop out.

The `wado-performance` byte-compare sweep passes its world as one `--world=…` token. An unquoted two-word variable does not split under zsh, which the `http_routing` row answers with `invalid option`.


gfx:
## Review response

**`tir.rs:4531`, `impl_receiver_key` — real as mechanics, unreachable at that site, and pointing at a larger class. Fixed at the class.**

The mechanics were right: after `redefine_to`, `get_unerased` reported `Ref`/`MutRef` while `get` followed the redirect to `Box<T>`, so `RefKind::from_resolved(self.get(id))` answered `None` and the receiver key degenerated to `builtin("")`. But all fourteen callers of `impl_receiver_key` sit in `elaborator/`, `synthesis/`, `monomorphize/` and `cm_binding/`, every one of which runs before `lower::plan::boxing`. No redefined slot reaches it, so the "incorrect compiler behavior" verdict does not hold up.

The class does. `get_unerased` means *the type before newtype / flags erasure*, and a boxing redefinition had quietly given it a second meaning. Three readers branch on `Ref`/`MutRef` through it:

| site | reachable after boxing |
| --- | --- |
| `impl_receiver_key` | no — this finding |
| `newtype_link_owning` | no |
| `fq_base_type_name` | **yes**, via `lower::translate::pattern` |

The third is the one on a live path, and the review never reached it.

**The fix is to the contract, not the line.** A boxing redefinition is not erasure, so `get_unerased` resolves it: a redefined borrow reads as its `Box<T>` there, as in every other view. `spelled_borrow` is now the sole reader of the borrow left in the slot, which is all `is_mut_box` and `fq_type_name_unboxed` ever wanted from it, and `box_redefinition` is the predicate telling a boxing redirect from an erasure one. All three readers answer as they did before a slot could be redefined, so the suggested patch to `impl_receiver_key` is not needed.

What that site did get is the invariant stated rather than papered over: the borrow kind now comes from a view that is guaranteed to agree with the erased one, so an `expect` fires instead of a receiver silently keying under the empty name. 13,427 e2e tests exercise it without tripping.

**Two tests**, both guarding what the class fix turns on: `a_redefined_borrow_still_says_whether_it_was_mut` fails if `spelled_borrow` ever reads through `get_unerased` again — `&T` and `&mut T` share one `Box<T>`, so the slot is the last thing that distinguishes them — and `a_redefinition_reaches_every_view_of_the_type` replaces one that asserted the contract now removed.

**Nothing skipped.** This was the only finding.

One correction to the PR description while here: the widened variant-return SROA does move generated code, on `serde_newtype_of_struct`, 52 bytes smaller. The benchmark and `wasm-size` corpora are byte-identical, which is what that claim was actually measured against.

PR #2083, opened 2026-09-19
feat(cm): let any module bind a Component Model import, in any namespace

gfx:
A Wado module can declare its own Component Model bindings. `#[cm("acme:demo/gfx@0.1.0#get-thing")]` on an interface operation names the import, `#[cm(...)]` on a `resource`, `record`, `variant`, `enum` or `flags` declares the type it carries, and the call lowers to that import. The namespace is whatever the interface says — `wasi:`, `acme:`, anything — and a name the standard library also uses is this module's own. That is the whole of what `wasi:webgpu` needs from the compiler, ahead of the `web:*` modules being bundled.

Two rules make it safe to hand out. A CM interface has exactly one declaring module, so a module declaring into `wasi:http/types` is refused by name, with the declaration it would have collided with. And a CM name that reads as a resource operation — `[constructor]T`, `[method]T.m`, `[static]T.m` — must name a resource the interface declares; on a resource declared `linearity = "unrestricted"`, which crosses as a plain handle and declares no CM resource, the binding is rejected at analysis time instead of reaching the component validator as an internal compiler error.

## Resolving a declaration

A CM declaration is reached through the interface that declares it, never through its name alone. The registry is keyed by `(source_interface, wado_name)` throughout, every emitter passes the interface it is describing, and one walker answers which resources a signature references, returning each with its declaring interface so no caller resolves it twice.

`error-code` goes further and keys by `DefId`. `wasi:sockets/types` and `wasi:sockets/ip-name-lookup` each declare one, and they are unrelated variants with different cases; a key built from the package names one of them. `ComponentModelContext` holds those aliases under the declaration's identity, and an alias is emitted only where the instance type it reads from exported one.

## At the boundary

An interface that defines resources may also declare plain functions of its own, and the imported instance type carries both. Two Wado names may bind one CM operation: the instance type exports the CM name once, and the core module takes one alias per used Wado name. An `Option` around a record or a resource reaches the same type walk as the bare form, so a lowering that needs the `memory` canonical option gets it at any depth.

## Tests

Seventeen e2e fixtures cover the boundary shapes, three of them refusals. `e2e.rs` gains `wat_lines`, which checks the emitted component's WAT — a CM import's shape lives there and nowhere a run reaches — and `compile_only`, for a program whose imports the runner hosts nothing for. Two cases stay integration tests: one enters through `dump`, the other spans several files.

## Documentation

The declaration-identity WEP records the one-declaring-module rule as following from its adopted identity rule, and two known gaps: component codegen still keys resource own-handles and per-package composites by name, and a transmission future still carries a package rather than a declaration. The resource-inheritance WEP states that an unrestricted resource declares no CM resource and so carries no resource-qualified operation.

https://claude.ai/code/session_01Y2so9tASYHn7tmMaKNeLhL

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y2so9tASYHn7tmMaKNeLhL)_


gfx:
Fixed in c261eb0f9.

**The class.** The finding named one instance — `compile_only` returning ahead of the WIR block — but the class is a spec key the runner accepts and silently does not honour. Two shapes admit it: an assertion placed after the execution dispatch (any future one, not just this branch's), and WIR that was never retained, which the old `if let Some(wir_text)` skipped without a word.

**What changed.** The WIR patterns are checked by `assert_wir_expectations`, called before the `compile_only` return alongside `assert_wat_lines`, so everything asserted about the compiler's own output now runs ahead of the dispatch and only execution sits behind it. The "no WIR retained" arm is now an assertion rather than a quiet return: a fixture stating expectations whose WIR was dropped fails instead of passing.

**What guards it.** `cm_interface_plain_func.wado` now sets `compile_only` and `wir_expect:O2` / `wir_not_expect:O2` together, so the combination is exercised. Confirmed red first — a deliberately absent pattern failed at `assert_wir_expectations`, which it could not have done before — then green. No fixture set both keys before this, so nothing was passing on dropped checks; the defect was latent.

e2e 5384 passed / 0 failed, clippy `-D warnings` clean.

---

gfx:
Second round, both fixed.

**`only_opt` excluding `Os`** (`e2e.rs` 1129) — real. The guard read `skip_os` directly, so a fixture keeping `-Os` out through `only_opt` was still required to set it and failed at the levels it does run.

Fixed a level up rather than at the site: `TestSpec::runs_at_os()` is now the one answer to "does `-Os` run this fixture?", read by both the runner's `-Os` skip and the WAT-symbol guard. The suggested patch would have inlined the `only_opt` rule into the guard, leaving the runner and the guard each holding their own copy of it — the drift that produced this finding. One predicate cannot drift from itself. While there, the runner's inline `OptLevel`→name match turned out to be a second copy of `common::opt_level_name`; that is gone.

**The `-Os` CI failure**, for the record, since it was mine and not reported here: five fixtures matched `wat_lines` against `$`-prefixed instance names, which `-Os` strips. They now carry `skip_os`, and the guard above refuses a `$` needle at every level so the next such fixture fails where it is written rather than in the one job that runs `-Os`.

420 fixtures across all five optimization levels, clippy `-D warnings` clean.

---

gfx:
Third round: one finding, skipped as not real.

**`CmBindingKind` needs `PartialEq`** (`component_model.rs` 932-935, 🟠 Major) — it already has it. `#[derive(Clone, Copy, PartialEq, Eq)]` sits on line 931, immediately above the cited range; the outside-diff scan read the enum body without its attribute line. The stated consequence, "the crate will not compile", is also contradicted by this commit: clippy `-D warnings` is clean and 13457 e2e, 29 emi and 776 integration tests pass on it.

No change made.

---

PR #2084, opened 2026-09-19
perf: beat JavaScript on both MicroGPT phases, and halve gale-gen

gfx:
MicroGPT runs ahead of JavaScript on training and on inference, and the Gale generator takes half the time it takes to generate a Rust parser. Both come from the same place: a constant a callee stores into the heap is now shared across every call that builds it, so it joins the live set once instead of once per call, and the live set is what a WasmGC collection walks.

| MicroGPT | Rust | **Wado** | JavaScript |
| -------- | ---: | -------: | ---------: |
| Train    | 78.023 ms | **152.109 ms** | 171.599 ms |
| Infer    | 85.901 ms | **95.655 ms** | 107.769 ms |

`gale_gen` goes to 74.663 ms, and json-catalog `core:json` deserialize passes the hand-written parser written to be the mark it had to reach. No `.wado` under `benchmark/` is touched; every row is the optimizer and the runtime knobs.

## Sharing a stashed constant

`const_object_globalization` hoists a constant whose uses only read it. A callee that stores the argument away answers a second question, in `optimize/shared_escape.rs`: sharing is safe when nothing in the whole program writes through the object. The analysis taints the object itself, follows it into whichever slot it lands in — a field name, a callee parameter, a callee result — and refuses on the first write and on every shape it does not model. "No write reaches this" is a safety property, so a query re-entering itself reads `true` and a verdict resting on that assumption is not cached.

A body-less callee has no body to follow, so its parameter is read from what the declaration states: a `&mut` position, or a `#[retain(p)]` clause, is what refuses. Only `core:builtin` answers that way, being the one module where those clauses are complete.

## Pricing the hoist, and declaring what may not move

A hoist costs a global that lives as long as the module, so what it buys depends on what becomes of the constant. One a callee stashes would join the live set once per call, so it pays whatever its shape. One that dies with the call saves only the work of building it, and must own heap storage to be worth a global.

That leaves legality to be stated rather than guessed. `#[immediate(p)]` marks a parameter that codegen lowers to a Wasm immediate and reads the literal of — `v128_const`'s bit pattern and the SIMD lane indices, which are every position in the backend that reads one. Nothing hoists an argument there, or a `let` that delivers one, because a global read at such a position is a broken lowering rather than a slow one.

## The GC heap is a knob

`--gc-heap-initial` sets what a guest starts with, and `wado run` defaults to 256 MiB. wasmtime collects before it grows, so a heap starting at zero doubles its way up to the working set and pays a full trace at each rung. MicroGPT and gale_gen hold enough live to want more still, and take 512 MiB from `gc_heap_flags` in `benchmark/wado.sh`; every other row measures flat or slower there and takes the default.

## Reading a `&mut` where it is still spelled

`link` snapshots what each body-less declaration says about its parameters, and takes the `&mut` positions from the parameter type. That is the only thing that says `&mut` at link time — `TirParam::is_mut_ref` is filled by `lower::plan`, which runs later, and boxing has not yet collapsed `&mut T` onto `&T`. `shared_escape_stashed_mut_builtin_write.wado` pins it: the write it makes goes through `builtin::array_set`, whose `arr` is `&mut` and is declared nowhere as an attribute.

## CLI knobs

`knobs.rs` carries two families: `CompileKnobs` for what the compiler is told, `RuntimeKnobs` for what the wasmtime engine is built with. A guest-hosting subcommand — `run`, `serve`, `test` — holds one field per family and declares no spec of its own.


gfx:
## レビュー応答

### 直したもの

**P0: `-O1` の miscompile — `shared_escape::reads_arg` が「結果」を脱出経路として見ていなかった。**

body-less な callee に「この引数をどう扱うか」を尋ねるとき、宣言から読み取れるのは `&mut` 位置（書く）と `#[retain(p)]`（保持する）だけで、戻り値はどちらでもない。`black_box<T>(value: T) -> T` は引数をそのまま返すので、定数が呼び出しを生き延びてフィールドに届き、そこ経由で書かれる。`opt_hfs_field_ref_arg.wado` の `"z"` が全部ひとつの共有 `String` になり、callee の `*v = …` がそれを in-place で潰していた。`-O1` で trap する。

参照を保持しうる戻り値型は、`#[result(owned)]`（返すオブジェクトは渡されたものではない、と述べる唯一の節）がない限り拒否するようにした。スカラー戻り値は影響を受けないので `array_get_value` は従来どおり答え、microgpt の wasm はバイト単位で同一。

**`#[immediate]` を trait / interface の requirement に書けてしまう穴。** 上を検証する途中で見つけた。`#[retain]` / `#[result]` と同じクラス — 三つとも body-less な宣言を記述するもので、requirement は impl に届く。`reject_declaration_attrs_on_requirement` に統合した。

### 見送ったもの

**CodeRabbit の「body-less メソッドへの `#[immediate]` で panic する」。** 再現しない。trait requirement 経由でも `#[unavailable]` メソッド経由でもコンパイルが通る。ただし隣接する本物のバグ（上の requirement の穴）は出てきたので、主張自体には価値があった。

### 反証したもの

**Performance Alert。** `origin/main` のコンパイラを worktree でビルドして全ベンチマークをコンパイルし直して比較した。`fts`・`json_twitter`・`json_canada` の wasm は main と**バイト単位で同一**で、アラートはこの3行に 1.5〜1.7x の回帰を主張している。同一バイト列は動かない。残り4本は +7〜+9 バイトの差で、いずれも 1.5x を説明しうる差ではない。フレークという読みで正しい。

### 検証

`CI=1 cargo test --locked -p wado-compiler --test e2e` を全5最適化レベルで: 13552 passed / 0 failed。`mise run format` クリーン。

PR #2087, opened 2026-09-19
feat: core:prng, reproducible randomness at 1.65 G u64/s, on a multi-value parameter ABI

gfx:
`core:prng` gives a Wado program reproducible pseudo-randomness. `Seed` is a value, `Rng` serves one u64 per call and `VectorRng` sixteen, and four engines implement them: xoshiro256++, its SIMD form, SHISHUA, and the keyed Squares64. That is 403 M u64/s from RomuTrio, 187 M from xoshiro256++, and 1.65 G from SHISHUA's batch — the batch figure being what the engine gives with no trait between it and the caller. `core:secure_random::seed()` is the single crossing from host entropy into a reproducible stream.

Carrying that batch across a trait boundary in registers is what the compiler half buys. A parameter whose every use is a field read takes a multi-value ABI: one Wasm parameter per field, no heap aggregate. Paired with the multi-value return ABI, a call whose result flows straight into such a parameter moves N values to N values and allocates nothing between them.

## `core:prng`

- `Seed` holds 256 bits, built `from_u64` / `from_bytes` / `from_str`. `split(index)` derives a child seed, so worker _n_ derives its own with no coordination.
- `Rng` requires `next_u64` alone; `random_range`, `random_f64`, `random_bool`, `shuffle` and `choose` are defaulted over it. `random_range` reaches eight integer widths and both range types through `SampleUniform` and `SampleRange` rather than overloads, and both traits are public so a caller can bound on them.
- `VectorRng::next_batch` returns sixteen u64 as eight `u64x2`, a tuple the call site destructures.
- `Xoshiro256pp::jump()` / `long_jump()` separate streams. `Squares64::at(counter)` is keyed and indexable, and its key predicate is an assert at construction.

`docs/wep-2026-09-18-core-prng.md` carries the design and the open questions. `example/prng_bench.wado` measures the library that ships rather than a copy of it, so the table above keeps being checked.

## The multi-value parameter ABI

`optimize::multi_value_param` classifies a parameter as multi-value when the function is reached only by direct calls, the parameter is neither `mut` nor `&mut`, its type has 2..=8 eligible fields, and every use in the body reads one of them. `wir_build` splits the argument at each call site: a matching multi-value call result binds field by field, a struct literal is spilled per initialiser and never built, and anything else is read field by field off a temporary.

Two invariants hold it together.

- Each split field lives in a WIR local minted by `name::multi_value_split_local`, which opens with and separates on the internal `$` prefix. No source local can spell one, and no two `(base, field)` pairs collide.
- A split argument is evaluated into the call's prelude, so every argument to its left is spilled. Source evaluation order survives whatever position the split argument sits in.

`NirFunction::only_reached_by_direct_call` is the one gate the return side, the parameter side, and `sroa_variant_return` all ask.

## A projection compared unequal to itself

`Rng` could not be implemented from any module that did not also import `SampleRange`, so `example/prng_bench.wado` failed to compile with `expected 'S::Value', found 'S::Value'`.

Materializing a trait's defaulted method for an impl in another module re-resolves the signature there. `bound_declaring_assoc_type` resolves the bound's trait at its own reference site, which fails where that trait is not in scope, so the projection interns with no owning trait and compares unequal to the qualified one. `check_projections` now compares two projections the way `TypeTable::resolve_assoc_type_qualified` already resolves one: by parameter and name, with an absent owning trait matching any. A qualified type name renders its owning trait, so the mismatch says which trait each side means instead of printing the same string twice.

The deeper repair — giving a materialized default body the trait's module for written names while `Self` stays the impl's — is a decision about the phase's contract and is not settled here.

## Generated

`docs/stdlib-core-prng.md` and `docs/stdlib-core-secure_random.md` are regenerated from the module sources.

https://claude.ai/code/session_01LU9MUjLutgKQRJ7bddfrag


gfx:
Pushed as `6e83a404b`. All four findings were real; each is fixed as the class it belongs to, and the CI failures on the previous head are fixed too.

## The inline finding: two algorithms for one name

Confirmed, and it had a twin the review did not reach. `register_single_function` derived the Wasm parameter names with a scheme of its own — `_{local_index}` on a duplicate — while the body derived them with `resolve_local_names`, which extends the suffix until the name is free. For `e`, `e_2`, `e` at indices 0, 1, 2 the signature reads `e_0, e_2, e_2` and the body `e_0, e_2, e_2_2`: a duplicate in the signature, and a base the split locals hang off that the body does not use. The plain (non-multi-value) name at the same site had the same divergence.

`resolve_param_names` is now the one answer, read by the signature, the body, and the dispatch stub. A unit test covers the `e` / `e_2` / `e` case. The unreachable panic this made of `split_locals_for_params`'s name lookup is gone with it.

## `drain_batch_fill` on an empty buffer

Confirmed: `0 % BATCH_WORDS == 0` passes and the loop then advances by zero. It now rejects an empty buffer as `drain_scalar_fill` does. Those two are the file's only loops striding by a runtime length.

## The generated stdlib docs

Confirmed. `docs/stdlib-core-prng.md` and `docs/stdlib-core-secure_random.md` are regenerated with `wado doc`, so they carry the doc comments the sources now have. `docs/stdlib-core-secure_random.md` had drifted the same way.

## The WEP's Roadmap section

`docs/AGENTS.md` lists Roadmap among a WEP's sections, so it is back. It says nothing is committed rather than listing what landed, which keeps the "no landed checklist" rule in the same file.

## Why CI was red, and what came out

Every failing job — Tidy, Benchmark, Test (Core), Test (Wado), Test (Gale O2) and all four E2E+stdlib levels — failed on one line: `core:prelude/traits.wado:1591: unknown function 'Instant::from_str_slice'`.

An elaborator change on this branch resolved trait default-method bodies in the trait's own module, so a default could name that module's private items. But `Self::x` is rewritten to a bare type name before lookup, so `FromStr::from_str` calling `Self::from_str_slice` stopped finding `Instant` and every other `core:temporal` type. The 5400-fixture e2e suite missed it because no fixture imports `core:temporal`; a six-line file reproduces it.

That change is reverted, with its two fixtures. `core:prng` compiles and its 28 tests pass without it, so it was never load-bearing here. Making it work needs a decision rather than a patch: a materialized default body wants the trait's module for the names it wrote and the impl's for `Self`, and `decl_key_in` answers from one frame by design. That belongs in its own change.

---

gfx:
One of the two fixed in `8d6d8ee02`, one skipped.

**`VectorRng`'s cursor claim** — right, and fixed at the source rather than in the generated file. "A consumer keeps no cursor" only holds for one reading whole batches; a per-draw consumer carries one, which is what `Rng` and `VectorRng` being separate traits is for. The doc comment in `wado-compiler/lib/core/prng.wado` now says what the fixed width buys — a whole-batch consumer knows it without asking the engine — and `docs/stdlib-core-prng.md` is regenerated from it. The WEP's own uses of "cursor" are about the engines, which genuinely keep none, so they stand.

**The Roadmap section** — leaving it as it is. `docs/AGENTS.md` says a roadmap item "will be done, so it is ordered and each entry says what finishing it means", that a WEP carries "no update log, superseded alternatives, or landed-checklist", and that "demoting a roadmap item to a gap, or promoting a gap, is the human's call". Everything the Decision settles is in the library, and nothing else is committed, so an ordered list here could only be a checklist of what landed — which the second rule forbids — or work I decided to commit to, which the third reserves for a human. Saying the section is empty and why is the reading that keeps all three. It is also what the earlier round of this review proposed as the fallback ("If no committed work remains, state that explicitly").

---

PR #2089, opened 2026-09-20
fix(gale): reject the escapes and labelled blocks ANTLR4 rejects, with a span on each

gfx:
Gale no longer accepts two `.g4` forms the published jar refuses. A grammar that compiles here now compiles upstream, so a fixture can reach the oracle scripts in the shape they run. Both diagnostics carry a span and name the offending text, unlike the whole-grammar fixpoint checks beside them.

## An escape the lexicon does not define — ANTLR4 error 156

A string literal takes `\n \r \t \b \f \\ \' \uXXXX \u{…}`. A char set takes the same set with `\- \] \p{…} \P{…}` in place of `\'`. `\uXXXX` takes exactly four hex digits. Anything else is an error rather than the bare character.

`~[\[\r\n]` is the form that motivated this: `[` needs no escape inside a set, so `~[[\r\n]` is the portable spelling and the two mean the same thing. Reading `\[` as `[` compiles a grammar the jar refuses and says nothing about it. Being a superset is allowed where the meaning is unique; being a *silent* superset is what costs a comparison.

The lexer gives the malformed escape its own token kind (`G4Token::InvalidEscape`) spanning just the escape, while still scanning the literal to its closing quote so the next token starts where it would have. In a char set, `take_char_code` reports it as a `ParseError` placed at the escape.

A `\u{…}` whose digits run to a `}` is read to that `}` before anything is decided, so the report names `\u{110000}` rather than a prefix of it. The scan consumes only hex digits and a `}` directly following them, so it can never cross the enclosing literal or character set.

## A label on a block that is not a set — ANTLR4 error 130

A set is two or more alternatives, each a single terminal. So `x += (A|B)` takes a label, while `x = (ID)`, `x += (A*)`, `n = (ID ',')+` and `x += (.|A)` do not. `(e)` is not the free wrapper it looks like: a label belongs on an atom or a set.

Four fixtures spelled a label on a block ANTLR4 would reject; each now carries the label on the element that binds, which is what the rule was pinning either way. `ll_not_alt.g4` reads `x=.`, and `wado_label_repeat_shape.g4` reads `( xs += A B )*?`.

With the rule in place, `unwrap_labeled_transparent_groups` is unreachable — a strippable bare wrapper has one alternative and a labelable set has two or more — so it is gone, along with `wado_label_group.g4` and the driver test that existed to pin the superset. No generated parser changes as a result.

## Scanning

The char-range-to-`String` copy stood in six files under four names (`collect_range`, `push_range`, `slice`, `copy_span`). `src/text.wado` holds the pair every scanner carrying a `List<char>` imports. Both assert their range runs forward rather than returning an empty string for a backwards one.

`found_desc` falls through to the raw token text, which says nothing for a kind the lexer could not form. Those kinds are the enum's trailing block, so one range assert catches the next one added without a branch there.

## Corpus

Every one of the 618 tracked `.g4` files was checked against the new lexicon and the label rule, the ANTLR4 descriptor corpus included: none carries a now-rejected form. The generated corpus under `tests/generated/` was regenerated; only the Kiln cache keys move, since the generator's own sources changed.

https://claude.ai/code/session_01XzgWeeEYaB341gp6KpfQo4


gfx:
Not taking the `lexer.wado:368-369` finding, because the oracle refutes its premise. Measured against `antlr-4.13.2-complete.jar`:

| grammar | ANTLR4 | Gale |
| --- | --- | --- |
| `X : [\u{41x}] ;` | `error(156): invalid escape sequence \u{41x` | `\u{41` |
| `X : [\u{4gg}] ;` | `error(156): invalid escape sequence \u{4gg` | `\u{4g` |
| `X : [\u{110000}] ;` | `error(156): invalid escape sequence \u{110` | `\u{110000}` |
| `X : [\u{}] ;` | `error(156): invalid escape sequence \u{}` | `\u{}` |
| `X : [\u{41] ;` | `error(156): invalid escape sequence \u{41` | `\u{41` |
| `X : [\uZZZZ] ;` | `error(156): invalid escape sequence \uZZZZ` | `\u` |

ANTLR4 never extends the report to the closing `}`. `\u{110000}` printing `\u{110` gives the actual rule away: the text is `\u` plus at most four following characters, stopping early at `}` or the set's `]`. That is a formatting artifact of its lexer, not a boundary that means anything, so advancing through the `}` would move Gale away from the jar rather than toward it.

All six reject, at `error(156)`, on the same column. Compatibility is a capability contract, not byte-for-byte output (`package-gale/CLAUDE.md`), so the residual difference is one character of message text on inputs the 618-file corpus does not contain.

The bounding concern behind the suggestion is sound, and it is why the scan stays where it is: `scan_hex_digits` consumes only hex digits, and the `}` only when it directly follows them. Neither can be a `'` or a `]`, so the scan cannot cross the enclosing literal or character set — a property that a `}`-seeking recovery would have to re-establish with a delimiter set threaded through both call sites.

What the finding did catch is an overstatement in my own PR description, which claimed a `\u{…}` is consumed whole. It is consumed to its `}` when the digits run to one. I have corrected that sentence.

---

PR #2092, opened 2026-09-20
refactor(stdlib): name the numbers the stdlib was spelling out, and document a struct's constants

gfx:
A numeric bound in the stdlib is written as the constant that names it. Where the bound is a type's range, it reads `i8::MAX`; where it is a rule of its own, it has a name of its own — `PORT_MAX`, `STORED_MAX`, `PAGE_SIZE`, `NEG_ZERO_BITS`. `i128` and `u128` gain the `MAX` and `MIN` every other integer type offers, and `wado doc` publishes a struct's associated constants, so those four and the ones `Formatter` and `Instant` already declared reach the reference.

## Which constant a bound gets

Two different questions share a value, and the distinction decides the name.

A check asking "does this fit the type I am about to cast to" takes the type's constant. That covers the `parse_{i,u}64_radix` bound arguments behind `from_str` and `from_str_radix`, the narrowing checks in `core:serde`, the i64 second range in `core:temporal`, and the u64 and i128 bignum boundaries in `core:cbor`. A signed type's negative magnitude is `T::MAX as u64 + 1`. `core:json_nsd` emits a number as a string outside the range JSON consumers hold exactly, which is `f64::MAX_SAFE_INTEGER`.

A constraint that belongs to a format or a protocol gets its own name, because it is independent of whatever type first bounded it. A URL port cap is `PORT_MAX`, a DEFLATE stored block is `STORED_MAX`, Wasm's linear memory unit is `PAGE_SIZE`, and the f64 bit pattern of -0.0 is `NEG_ZERO_BITS` — an i64 whose meaning has nothing to do with i64's range.

A bit mask, an IEEE-754 exponent bias, an array index, and a test's expected value stay as they are. A name there costs the reader more than it saves.

## 128-bit constants

`i128::MAX`, `i128::MIN`, `u128::MAX` and `u128::MIN` are declared alongside the other integer types' and used for the `i128` parse boundaries.

## Associated constants in the reference

`wado doc` collected associated constants only for impls on primitive types, so a `pub const` on a struct never reached the output. `collect_impl_constants_for_type` sits beside the existing `collect_impl_methods_for_type` and serves the struct path too; both renderers emit what they already emitted for a primitive. This is what makes `i128`/`u128`'s constants visible, since those are structs, and it also publishes `Formatter::NO_WIDTH` and its siblings and `Instant::EPOCH`. Only structs carry associated constants today, so the enum, variant, flags and newtype paths are untouched.

Both impl collectors read the name an impl targets through `ast::type_head_name`.

`docs/stdlib-core-prelude.md`, `docs/stdlib-core-temporal.md` and `docs/stdlib-core-icu.md` are regenerated.

## `scripts/changed-sources.sh`

The script used `grep -P`, which GNU grep provides and BSD grep does not, so it failed on macOS with `invalid option -- P`. It now matches the tab with bash ANSI-C quoting and runs on both.


gfx:
## Review response

**Fixed — `internal struct` under `--all` (`wado-cli/src/doc.rs`)**

The claim holds. `Visibility::Internal` renders as `internal `, so `strip_prefix("struct ")` returned `None` and the `expect` fired. Reproduced with a test before fixing:

```
panicked at wado-cli/src/doc.rs:1258:18: a struct signature names its type
```

Fixed a level up rather than at the site. The class is the renderer recovering a declaration's bare name by re-parsing its rendered signature: six of the `Doc*` types with signatures — `DocVariant`, `DocEffect`, `DocResource`, `DocFlags`, `DocType`, `DocGlobal` — carry a `name`, and `DocStruct` was the one that did not, which is why it was also the one site doing the string surgery.

So `DocStruct` carries the name `build_doc_struct` already has, and the renderer writes it:

```rust
writeln!(out, "\nimpl {} {{", s.name).unwrap();
```

Adding `internal ` to the stripping would leave the next `Visibility` variant to reintroduce this. There is now no prefix list to keep in step, and the fix is 7 lines shorter than the code it replaces.

Covered by `an_internal_struct_heads_its_impl_block_with_its_own_name`, which renders the `--all` simple view of an `internal struct` with an associated constant.

Regenerating the stdlib reference produces no diff, so published output is unchanged.

PR #2094, opened 2026-09-20
feat(core:fs): directory removal, path operations, metadata, a walk, and a write that replaces rather than truncates

### LINKED ISSUE #2061: `core:fs` cannot remove a directory, and the WEP's stated reason for leaving it out no longer holds
## The friction

Sheaf regenerates `_site/` on every run. A file that a previous run wrote and the current one no longer produces stays there, and the deploy publishes whatever is in the directory — so clearing the output before writing it is the thing you reach for.

`core:fs` offers `remove_file` and nothing else in that direction. There is no `remove_dir` and no recursive removal, and `remove_file` on a directory is documented as failing. So clearing an output tree means walking it with `read_dir`, recursing into every `Directory` entry, `remove_file`-ing the leaves, and then having no call available for the directories themselves — the walk cannot finish. The only way through is `root()` and raw `wasi:filesystem`, which is the plumbing `core:fs` exists to avoid, for the one operation in the set that is missing.

The outcome was that the build does not clear its output at all. Not because that is the right behaviour, but because the alternative was worse than the problem.

## Why this is worth filing rather than just noting

`docs/wep-2026-09-12-core-fs.md` already names this:

> Deliberately absent: append, rename, symlink, `remove_dir`, metadata / `stat`, file times, permissions, random access, and reading a file that does not fit in memory. Each is `wasi:filesystem` through `root()`. They are out because no caller in the repository needs them, not because they would not fit.

The exclusion rests on a factual premise — no caller needs it — and that premise is what changed. Sheaf is a caller that wants it. It is not in this repository, but it is a `core:fs` consumer written against the module's actual surface, and it is the first real one outside the tools the WEP's roadmap lists.

So this is not "the WEP forgot something". The design note is explicit and the reasoning was sound when written. The friction is that the condition it made the decision under has since stopped being true, and a caller that hits the gap has no way to signal that from outside the repository except by filing it.

The asymmetry is also what makes it surprising in use: `create_dir` and `create_dir_all` are both there, so a module that can build a directory tree cannot take one down. You find that out after writing the code that assumed otherwise.

gfx:
A tool written against `core:fs` can now take a tree down, read a path apart, ask what a path names, walk a directory, and write a file without risking the one it replaces. Sixteen I/O functions and six path functions, and the repository's own callers go through them: marl walks with `walk_dir` and asks `metadata` what a CLI argument names, `tree` and the gale scripts join through `fs::join`, and the gale facade and the descriptor extractor take their parent and stem from here. Six local path helpers are gone.

Closes #2061.

## A write replaces the file

`write` streams into a file beside the target and renames it over, so a reader sees the old file or the new one and a writer that dies partway leaves the original. That matters for what this repository writes: a `.md` the user is editing, a committed grammar, a checked-in baseline.

The temporary file is a sibling, not a file in a temp directory — rename is intra-filesystem, the host refuses one between grants whose permissions differ, and WASI offers no temp directory. `Create | Exclusive` makes the host settle uniqueness, so no `Random` enters the signature of every function that writes a file.

`write` refuses a target that is not a regular file. A rename replaces a symlink instead of following it, and no read here follows one either.

Atomicity is not durability: nothing calls `sync_data`, and the doc comment says so. `write_in_place` keeps the truncating write for a file too large to exist twice.

## Time of check, time of use

A path is a name, not a handle, so every call resolves it again and what a call reports is true of the moment it ran. `write` and `create_dir_all` each read a path before acting on it and so carry a window; every other call is one the host settles by itself, and `read` asks the descriptor it opened rather than resolving the path twice. `remove_dir_all` and `walk_dir` act on a listing. Each doc comment says which of those it is, whether or not it has a window, so a caller never has to infer it from the implementation.

No window reaches outside the preopen. Every path opens with `PathFlags::none()` and a symlink entry is `Other` rather than `Directory`, so the swap that turns `rm -rf` into a way out of a sandbox gets `Io(Loop)` or unlinks the link. A test covers exactly that.

## Path operations, in this module

`join`, `parent`, `file_name`, `file_stem`, `extension` and `normalize` live in `core:fs` rather than a `core:path`. A WASI path is UTF-8, `/`-separated and preopen-relative, so there is no path syntax to abstract over and no owned path type to own. The rules these functions apply are this module's: a leading `/` resolves against no preopen, `..` past the preopen is the sandbox boundary, and `""` names the preopen itself.

They declare no effect, so calling one adds no `wasi:filesystem` import to the component — the boundary a separate module would draw, effects already draw.

`core:url` also joins and normalizes paths and shares none of this code. Its rules come from the WHATWG URL standard. Two standards that look alike are the reason to keep two implementations.

## Removing a directory

`remove_dir` hands the host's answer back, so a non-empty directory is `Io(NotEmpty)`. `remove_dir_all` walks `read_dir` itself; a path that names nothing is not a failure, since what the call promises is that the tree is gone.

A path that *resolves* to the preopen is `Io(NotPermitted)`. The check runs `normalize` rather than comparing text: `"./"` and `"x/.."` name the preopen as surely as `"."` does, and the walk starts before the host ever sees the path.

## A walk is a list

`walk_dir` returns the whole tree. An iterator is not available — every standard library trait declares `with ()`, `Iterator` included, so a `next` that reads a directory is a design error. The pruning an iterator would have bought is a `descend` predicate, asked about each directory before the walk enters it.

A `WalkEntry` carries the path that reaches it, not its name alone, so a caller reads or removes what it finds without rebuilding the path.

## Asking whether a path is there

`exists` answers `bool`, folding everything that stopped the look into `false`. `try_exists` answers `Result<bool, FsError>` and reads `false` only as "nothing is there", so a directory with no search permission does not read as absence. Rust splits the pair the same way and gives the short name to the folding one.

## The effect comes from this module

`core:fs` re-exports the `wasi:filesystem` names its own signatures carry — `Preopens`, `Descriptor` and `ErrorCode` — and `Instant` from `wasi:clocks`. A caller writes `Preopens` in every `with` clause, so importing `wasi:filesystem` for that one name would be importing the module this one exists to replace. Twenty-two files take the effect from here; the files that call `wasi:filesystem` themselves still import it.

## Errors

`FsError` carries the path, and `Display` renders `build/out.json: no such file or directory`. Every error out of `write` names the path the caller passed, never the temporary file beside it.

Design and the known gaps are in [WEP: Whole-File I/O (`core:fs`)](https://github.com/wado-lang/wado/blob/claude/wado-rm-rf-impl-k9qgyb/docs/wep-2026-09-12-core-fs.md). `docs/stdlib-core-fs.md` is regenerated.

https://claude.ai/code/session_01GA1KmDMSDp4jyH4vmScgr1

---
_Generated by [Claude Code](https://claude.ai/code/session_01GA1KmDMSDp4jyH4vmScgr1)_


gfx:
Both CodeRabbit findings reproduced and fixed in `053786ea8`.

**`walk_dir` pruning entered `.git`.** Confirmed: walking `"."`, `walk_into` builds the entry path as `"./.git"`, so `e.path != ".git"` stays true and the walk descends. Rather than comparing against `"./.git"`, the predicate takes the component, which holds whatever root the walk started from:

```wado
fs::walk_dir(".", |e| fs::file_name(&e.path).unwrap() != ".git")?;
```

The pruning test used the same loose shape (`!e.path.ends_with("/skip")`) and now goes through `file_name` too, so the test and the documented idiom agree. `walk_dir`'s doc comment and the WEP now say `descend` receives the whole path.

**`NameTooLong` on a legal path.** Confirmed: a 254-byte final component writes fine through `write_in_place`, but `write` failed, because `<name>.wado-tmp` is 263. `NameTooLong` on the temporary file now drops the target's name and keeps its directory — the next candidate is `<dir>/.wado-tmp`, nine bytes whatever the target is called, tried once per write with `Exclusive` settling the rest. A test writes a 250-byte name and reads it back.

**Out of Scope Changes — not acting on this one.** The check asks to narrow the PR to `remove_dir`/`remove_dir_all` and split the write, metadata, walk and path helpers into separate PRs. These are one piece of work rather than several: the replacing `write` needs `rename`, `remove_dir_all` and `create_dir_all` and `walk_dir` all need `join`, and `write` and `ensure_dir` both need `metadata`. Splitting would produce a stack of dependent PRs, not independent ones. Recorded for the repository owner to decide.

---

gfx:
Scope decision from the repository owner: this PR stays as one. The call-site migrations (`example/tree.wado`, `package-gale`, `package-marl`) are not a separable change — they are what deletes the six hand-rolled path helpers the new `core:fs` path operations replace, so splitting them out would leave the duplicates standing in the tree between the two PRs.

No further action on the "Out of Scope Changes" note.

---

gfx:
Both findings from the latest review verified against the current code. `cd69f934c`.

## The class behind finding 1

`replace_file` stats the destination and then renames; `ensure_dir` asks `create_dir`, reads `Exist`, and then stats. Both are check-then-act, and both were written up as if the refusal were absolute. `wasi:filesystem` offers no rename that validates its destination and no create that would serve as the check, so the code is already as strong as the primitives reach — what overclaimed was the wording.

Fixed by saying it once rather than at the site the review named:

- The WEP gains **What holds against another writer**: `Create | Exclusive` and `rename` are atomic because the host makes them so; every refusal that reads the path first is best-effort, `create_dir_all`'s included.
- The doc comments on `write` and `replace_file` say where the look sits relative to the rename.
- Known gaps records the residue and what closing it would take.

## Finding 2, in two parts

The `Loop` half does not hold: `stat_at` passes `PathFlags::none()`, so `metadata` succeeds on a symlink and reports it as `Other`. `exists` is `true` for one, which the WEP already recorded.

The `Access` half is real — a path the preopen grants no search permission on reads as absence. Making `exists` fallible or adding `try_exists` is a public API change, so it is recorded as a known gap and put to the repository owner rather than adopted here. `exists` matches Rust's `Path::exists` as it stands, and `try_exists` is Rust's answer to the same split.

---

gfx:
The `microgpt/train (-O2)` alert is not this PR's, and there is nothing to port.

`benchmark/microgpt/microgpt.wado` imports `core:cli`, `core:collections`, `core:benchmark`, `core:prng` and `wasi:clocks`. This branch touches no Rust and no stdlib module on that list — its source diff is `wado-compiler/lib/core/fs.wado`, its test file, the callers in `example/` and the two packages, and documentation. Nothing in it reaches the code microgpt compiles to.

`benchmark/AGENTS.md` asks for best of three because a single cloud-VM run does not settle a ratio; the alert is one run against one.

No change pushed for it.

---

gfx:
Both findings verified and fixed in `f5d549cd0`.

**"Every call resolves its path" overstated the contract.** Correct: `join`, `parent`, `file_name`, `file_stem`, `extension` and `normalize` read a path as text and open nothing, and the module doc contradicted itself — the paragraph above the time-of-check section already said exactly that about them. The statement now names the filesystem calls it describes, in the module doc, the cheatsheet and the WEP, and the API reference is regenerated.

**`remove_dir_all` refusing the preopen was missing from the API reference.** Correct, and it was a class of two: `remove_dir` calls `deny_preopen` as well, and neither doc comment mentioned it. Both say it now, which is what `wado doc` renders.

The scope note is unchanged and settled — see [this comment](https://github.com/wado-lang/wado/pull/2094#issuecomment-5748365495).

---

gfx:
Review 5260090457 answered in 850380e32.

## Fixed

The `remove_dir_all` doc claimed `Io(NotDirectory)` for everything that is not a directory, and said "a symlink is unlinked" without saying that means an entry under the root. A probe against wasmtime settles both halves:

```
remove_dir_all(link) = Result::Err(FsError { path: "…/link", kind: FsErrorKind::Io(ErrorCode::Loop) })
still there: true
link there: true
```

`read_dir` opens the root with `PathFlags::none()`, so a symlink root is neither followed nor unlinked — the call fails and both the link and its target survive. The doc now separates the two cases, and the sentence about unlinking is scoped to entries under the root. `docs/stdlib-core-fs.md` is generated by `wado doc`, so the source of the change is the doc comment in `wado-compiler/lib/core/fs.wado`; the Markdown was regenerated.

The exact error code is left out of the sentence on purpose. `path-flags` in the WASI WIT documents what `symlink-follow` does and not what its absence returns, so `Loop` is what wasmtime gives rather than what the interface promises. The doc states the property (not followed, not unlinked), and the test asserts that property rather than the code.

## Added

`remove_dir_all refuses a symlink root rather than emptying it` — the existing test covered a symlink *entry* inside the tree; the root was the other half of the one property here that a wrong answer would make dangerous. 61/61 green in `wado-compiler/lib/core/fs_test.wado`.

## Not changed

The merge-risk note about `write`'s symlink refusal being racy is the check-then-act window already documented as a known gap in [WEP: core:fs](../blob/claude/wado-rm-rf-impl-k9qgyb/docs/wep-2026-09-12-core-fs.md). `wasi:filesystem` has no rename that validates its destination and none of its creates reports what it found, so the window is not this module's to close; what it cannot become is an escape from the preopen, which is what the symlink tests pin.

The "Out of Scope Changes" warning stays as settled earlier in this PR (#2094 comment 5748365495): the owner asked for no split.

---

PR #2095, opened 2026-09-20
fix(elaborator): a projection names its declaring trait, and a type parameter its dense slot

gfx:
Two identities the elaborator was carrying loosely are now exact. An associated-type projection knows the trait that declared it, so two `T::Output`s reached in one frame — the one a signature writes and the one `a * b` synthesizes — intern to the same type and a generic body typechecks against its own return type. A type parameter knows the slot it fills, so a declaration carrying an `effect` parameter numbers the rest densely instead of naming a slot no argument reaches.

The second one closes two compiler bugs. `struct Holder<effect E, T>` panicked WIR build with an unregistered generic instance, and `impl<effect E, A, B> … for Two<A, B>` shifted every slot past `E`, so `type First = A` resolved to `B` and reached codegen as an invalid module while `type Last = B` resolved to nothing at all. Both are pinned by fixtures.

## Projection identity

`make_frame_projection_of_trait` is the one place a projection is built from a trait. `ResolvedType::AssocTypeProjection` carries the `DefId` of the declaring trait, and the `bounds` it stores are part of what it interns by, so they are read through `TraitEnv::assoc_type_decl` — the declaration `owning_trait` names, never the associated type's bare name.

`TraitEnv` holds the walk that answers which trait declares a name: `assoc_type_decl`, `declares_assoc_type`, `trait_declaring_assoc_type` (the trait or the supertrait it inherits from), and `bound_declaring_assoc_type` (which of a binder's bounds). Each bound resolves through its own reference site, supplied by the caller, so a trait's defaulted method materialized for an impl in another module reads a bound that module reaches only through a namespace import.

`TraitEnv::build` runs ahead of the type-collection passes, which is what lets a declaration-level resolver name the declaring trait as well: `ParamSlot` carries each type parameter's name and its bounds, and `resolve_type_static_with_params` takes those slots.

## Slot identity

`RealTypeParams` is the type a declaration's type parameters have once the dense type-argument space holds them, and its only constructor filters. A `StructFieldInfo`, `VariantInfo` or `GenericNewtypeInfo` takes no other list, so its arity and its slot `TypeId`s cannot disagree.

Two rules meet here, and they differ. A `struct`, `variant` or newtype numbers by `is_real_type_param`: an effect or `fn`-bound parameter fills no argument position. An `impl` numbers by `!is_effect`, because a `fn`-bound parameter can still be a target argument — `Holder<F>` puts `F` at position 0, which is where `bind_declared_target_params` numbers it.

## Checking

`contains_fillable_slot` separates a type that mentions a slot from one a value assigned to it could still fill. `I::Item` is the first: inference cannot invert a projection, so a field declared that way is checked once the type arguments are known. Every field the first pass defers is revisited, coercion or not.

Two projections agree exactly when they intern to the same type; `typecheck` has no rule of its own for them.

## Elsewhere

Reify's one remaining resolution reads a global's declared type in that global's own module scope — the module level, where no function's type parameters are in scope.

`docs/wep-2026-08-12-declaration-identity.md` records that a projection's bounds are declaration data read through a `DefId`, and `docs/wep-2026-05-26-elaborator-rearchitecture.md` the facts that remain AST-shaped on `TraitEnv`. The latter also carries a known gap: a `struct`, `variant`, newtype or `impl` may declare an `effect` parameter that names nothing a use site can fill, and whether to reject it at its own span is a language-rule decision left open.

The WIR golden fixtures were regenerated.

https://claude.ai/code/session_01LU9MUjLutgKQRJ7bddfrag


gfx:
## Review response

Four findings across two CodeRabbit passes. Three were real and are fixed; one is not constructible in the language. Both real code findings turned out to be instances of one class — a type-parameter list counted as written where the dense type-argument space was meant — and chasing the class found **two compiler bugs**, both P0 under `AGENTS.md`.

### Fixed

**`types.rs` — `ParamSlot::list` took the whole declaration list.** Real, and the class was wider than the one site: eleven places built a declaration's `type_params` from the raw list, so the arity disagreed with the slot `TypeId`s beside it. `struct Holder<effect E, T>` minted a slot no argument could ever fill and panicked WIR build:

```
[WIR] generic instance Holder<i32,T> (as neither a struct nor a variant) is not registered
```

There is now one type for it, `RealTypeParams`, whose only constructor filters. The field can no longer be assigned a raw `Vec`, so a twelfth site cannot get it wrong. Pinned by `data_decl_non_real_type_param_slot.wado`.

**`orchestration.rs:3578` — `param_at` indexed the raw impl param list.** Real, fixed with `ParamSlot::list`. Fixing it alone changed nothing, though — the probe still produced invalid Wasm, which led to the second bug below.

**`register_impl_block_params` (`elaborator.rs:187`) — the actual culprit.** It numbered the slots an impl's methods resolve against straight off the written list. With `impl<effect E, A, B> … for Two<A, B>`, `A` got slot 1 and `B` slot 2 while the target supplies arguments at 0 and 1:

- `type First = A` resolved to `B` — `take_first` returned `String` where `i32` was declared, reaching codegen as an invalid module (`expected i32, found (ref $type)`).
- `type Last = B` resolved to nothing — `unsubstituted AssocTypeProjection 'First' reached codegen`.

A control without the `effect` parameter compiled correctly in both directions, so the shift was the cause. Pinned by `impl_assoc_type_non_real_param_slot.wado`, which covers both slots.

The rule at that site is `!is_effect`, not `is_real_type_param`: an effect parameter can never be a target argument, but a `fn`-bound one can (`Holder<F>` puts `F` at position 0) and keeps its written position, which is what `bind_declared_target_params` numbers by. Filtering it here would have shifted slots the other way.

**WEP `type_lookup()` count.** Real — it is 14, not 13. Corrected.

### Skipped

**`expr.rs:4068` — positional `substitute_type_params` in the deferred struct-field pass.** Not real. The mechanism needs a struct's *declared* field type to mention a parameter from an enclosing scope, and the language does not admit that:

```
local.wado:2:31: error: unknown type 'T'     // struct Box2<U> { a: U, b: T } inside fn wrap<T>
```

No fixture can reach it, so adding one would assert something true by construction. (The identity-keyed `subst` at `expr.rs:4042` is a different case — it substitutes the literal's *value* types, which really can carry an enclosing item's parameters. That one is correct as written.)

### New known gap

A `struct`, `variant`, newtype or `impl` may declare an `effect` parameter that names nothing a use site can fill and nothing a body can read. Both panics are gone, but the declaration is still admitted and silently ignored, and `Reflect` derivation stops short of such a type with a message naming neither the parameter nor the declaration. Whether to reject it at its own span is a language-rule decision, so it is recorded as a Known gap in `docs/wep-2026-05-26-elaborator-rearchitecture.md` rather than decided here.

### Also distilled

`call.rs` carried a second spelling of the same predicate and three copies of the blanket-static fallback, two of them the arms of an `if`/`else` that did the same thing.

### Verification

Full e2e green on the final tree: **5452 passed, 0 failed** (+2 from the new fixture; the earlier run was 5448). Regenerating the WIR goldens produced exactly one new file and moved no existing one, so the slot renumbering perturbs nothing else in the corpus.

---

gfx:
## Second review round

Both findings were real. One was mine, introduced by taking the previous round's suggestion without checking it against the rule on the other side.

### Fixed: two slot rules for one `impl`

`register_impl_block_params` numbers by `!is_effect`; `ParamSlot::list` filters by `is_real_type_param`, which also drops `fn`-bound parameters. Adopting `ParamSlot::list` at the associated-type registration last round put those two paths on different rules, so with `F: fn(...)` before `I: Trait`, `I` would be bound at one index and resolved at another.

An `impl` cannot use the data-declaration rule. A `fn`-bound parameter can be a target argument — `Holder<F>` puts `F` at position 0 — and `bind_declared_target_params` numbers target arguments by their written position, so filtering it out would shift every parameter after it. `ast::GenericParam::fills_impl_slot` now names the impl rule, and both paths read it.

### Fixed: a variant's slot ids and its arity could disagree

`VariantInfo.type_param_type_ids` was built by reading back every name `register_generic_params` left in scope. A `fn`-bound parameter is registered there — as its function type, consuming no index — so that list could be longer than the arity stored beside it. It now goes through `slot_type_ids` like the struct branch, from the same `ParamSlot::list` the arity comes from.

### No fixture for either

Neither is reachable from a compilable program, so a test would assert something true by construction. A `fn`-bound impl parameter fails both ways:

```wado
impl<F: fn(i32) -> i32, B> Pair for Two<F, B> { ... }
// error: no method 'first' found on type 'Two<fn(i32) -> i32, String>'

impl<F: fn(i32) -> i32, T> Holder for Box2<T> { ... }
// error: the type parameter `F` is not constrained by the impl target or the trait reference
```

Named in the target, the impl never matches; left out of it, `F` is rejected as unconstrained. The skew is real in the code and worth closing — a later change making `fn`-bound impl parameters work would otherwise inherit it silently — but nothing today can observe it.

That also narrows the merge-risk note above: the `fn`-bound half of it could not affect a program that compiles.

### Verification

Full e2e green on the final tree: **5452 passed, 0 failed**.

---

PR #2096, opened 2026-09-20
fix(cm): four kinds of value that lowered to the wrong bytes at the component boundary

gfx:
Four kinds of value crossing the Component Model boundary lower to the bytes the
Canonical ABI defines for them: a payload-less `variant`, a component-imported
`enum` or `flags`, a newtype, and a resource reference the CM binding rebuilds
from a `TypeId`. Each takes its width from the same registry walk that decides
the offsets around it, so a record places every field after such a value at the
offset the ABI gives it, a params-buffer store lands inside its own slot and
inside the buffer, and a `future<option<own<error>>>` over `wasi:tls`'s `Error`
lowers as the `Option` it is. `package-cm-catalog` covers each shape against
wasmtime's own lift and lower.

## A reference carries its declaring interface

A nominal CM reference resolves through the interface its declaring module
registers, a resource reference rebuilt from a `TypeId` included, so resolving
one happens in a scope rather than by searching for its name.

Where no interface is available the bare-name search remains, and it is one
list. `CmTypeKind::ALL` holds the registry's six keyspaces in one order, and
both searches — `find_binding_source` for a bundled namespace,
`find_kiln_source` for `core:kiln/*` — walk it, so a kind reaches both at once
and which kind answers first is a single tiebreak.

[WEP: Declaration Identity](docs/wep-2026-08-12-declaration-identity.md) records
the three positions a reference still arrives in without a declaring interface,
and what two kinds answering with different interfaces yields, which depends on
which of the two searches asked.

## One layout, walked once

`cm_abi` has one layout core per behaviour, parameterized by what reads a
field's size and alignment. `cm_layout_with_registry` is the single answer for
size, alignment and offsets, and every call site reads it once for all three.
`CmLayout::payload_offset()` names the variant payload slot, and `cm_layout_i32`
is the one place a layout becomes the i32 pair a `realloc` argument or a buffer
stride takes.

`cm_discriminant_byte_size` is the one discriminant width, taken from the case
count, so a variant past 256 cases carries two bytes everywhere one is stored,
freed or lifted.

## A known gap states what it admits

`docs/AGENTS.md` says a known gap states what is missing and what it admits, and
not how to close it: a written approach anchors whoever comes to the gap on what
its writer saw before the problem was understood. The WEP corpus follows, each
gap stating the property of the code it rests on. The elaborator WEP's entries
are measurements with owners and an order, so they sit under its roadmap.

## Corpus

`package-cm-catalog` gains `Status` and four identity exports covering it bare
and inside `option`, `list` and `tuple`, plus `narrow-scalars-in-params-buffer`,
an async export whose `Option` parameter forces the one-buffer path. The
committed catalog component under `wado-compiler/tests/fixtures/sub/` is
regenerated from it.

https://claude.ai/code/session_019vGS1FrXwzMQdrbs2SaZF2

---
_Generated by [Claude Code](https://claude.ai/code/session_019vGS1FrXwzMQdrbs2SaZF2)_


gfx:
Answered in 59367880a. Two fixed, one skipped, one left open as a language decision.

## Fixed

**`package-cm-catalog/src/lib.wado` — the params-buffer oracle could be satisfied by wrong bytes.** `tag * 1000 + o * 10 + p * 100 + seen` overlapped: `o` is a `u8`, so `o * 10` spans 0..2550 and runs through both the `p` band and the `tag` band. `Red` with `Some(107)` and `Green` with `Some(47)` plus `Read` both total 2572. This fixture exists to catch a store that writes over the slot beside it, so a total a wrong store can reach reads as coverage without being it. Each field now sits in a band wider than its own range, and the component is regenerated from the source.

**`docs/wep-2026-02-28-serde.md` — the type-argument claim was false.** It said a trait bound takes associated-type bindings but no type arguments, so `Self: From<Self::Base>` is unspellable. A bound does take them: `fn take<T: Pick<i32>>(…)` and `trait C: From<String>` both compile and dispatch. The sentence now states what the gap actually rests on.

## Skipped

**`docs/wep-2026-08-12-declaration-identity.md:750-753`, the declaring-interface guarantee.** The "when available" scoping is already there, in the next clause: "Where a reference arrives without one, the registry searches for the name instead." The section then lists all three positions a reference arrives in without a declaring interface, CM-instance emission among them. Line 750 read apart from 752.

## Open, and not this PR's to decide

Correcting the serde sentence turned up a real defect. `trait C: Make<Self::Base>` with `impl C for UserName { type Base = String; }` and a matching `impl Make<String> for UserName` is rejected with *"type 'UserName' implements trait 'Constrained' but not its supertrait 'Make'"*. `trait C: Make<String>` passes, so it is the projection and not the grammar, and a hand-written `From` fails identically, which rules out #1932. `enforce_assoc_type_bounds` projects `Self::Assoc` through a `SelfBinding` on the call path; `enforce_impl_supertraits` reads the written spelling instead.

Not fixed here. `docs/wep-2026-07-27-super-traits.md` records no gap either way, so making it work settles a language rule no WEP states, and that is the human's call under `docs/AGENTS.md`. It is recorded as a known gap in the serde WEP meanwhile.

Integration 780 passed, e2e 5454 passed, 0 failed.

---

gfx:
`windows-latest` failed on 59367880a, and the failure is not this PR's.

It died downloading the VS Code editor archive for `wado-vscode`'s extension tests:

```
- Downloading (319.94 MB)
✖ Error downloading, retrying (attempt 1 of 3): aborted
✖ Error downloading, retrying (attempt 3 of 3): aborted
Error: Failed to download and unzip VS Code 1.131.0
```

Three attempts aborted inside 0.6 s total, so nothing was transferred. This is the editor download in the job's setup, before any extension test body runs — the 51 tokenization tests in the step above it all passed. The diff touches no file under `wado-vscode/`, and `windows-latest` was green on the immediately preceding head (2a888eb62), whose `wado-vscode` tree is identical to this one.

No fix exists to port: the failing call is `@vscode/test-electron` fetching from `update.code.visualstudio.com`, not repository code. Re-running the job is blocked while the rest of the workflow is still in progress; I will spend the one re-run as soon as the run completes, and will treat a second failure as real.

---

gfx:
Resolved without my re-run: `windows-latest` retried on its own and passed on the same commit (job 106072300302, green at 11:50:35Z), confirming the aborted VS Code download was transient. No re-run spent.

---

gfx:
Fixed in e091df7a4. The finding is right and cites the rule: `docs/CLAUDE.md` says not to document implementation details, because a reader this far from the code has no way to notice when one goes stale. Naming the supertrait check and its missing projection was exactly the part a fix would change, so the gap now states what a reader hits instead — every impl of `Constrained` rejected as not implementing `From`, the ones that spell it included.

---

PR #2098, opened 2026-09-20
feat(kiln): key a build on the header a generator read, not the whole input

gfx:
A Kiln generator may declare how many leading bytes of each input determine its output, and the build keys that invocation on exactly that range. A checkpoint that is gigabytes of tensor payload behind a few megabytes of schema is rehashed over the schema alone on every incremental build; a payload-only edit is a cache hit. Every input also crosses the generator boundary as `stream<u8>`, so a schema is not required to be text and a generator reads only as far as it needs.

Two compiler bugs found on the way are fixed here as well: a nested variant's tag was stored at the wrong width, and a tuple returned at the wrong pack shape reached codegen.

## The probe

A generator may export `probe` beside `generate`:

```wado
export fn probe(path: String, content: Stream<u8>, options: Options) -> Result<u64, Error>
```

It is called once per input file and answers an extent, the leading byte count the output depends on. Two invariants make that sound, and a generator owes both: the output depends only on the first `extent` bytes, and the extent is determined by bytes inside it. The driver enforces the first rather than trusting it, cutting each input to its extent before `generate` sees it, so a generator reading further gets EOF. The second is what makes a cache hit cheap: the extent is recorded with the invocation, and a later build hashes that range without instantiating anything.

An extent reaching the end of the file records as the whole file. Such a probe was stopped by the file's length rather than by its contents, so recording the length would let a longer file sharing the prefix hash equal and hit the cache.

`options` is present only when the generator declares an `Options` struct, matching `generate`. It is there because an option can move the extent — Geas's `inline_weights` makes the output depend on the payload rather than the header.

A generator exporting no probe keeps the whole-file meaning.

## Inputs are bytes

`input-file.content` is a `stream<u8>`. `core:kiln` supplies `read_all` and `read_text` for a generator that wants the whole file, and `String::push_utf8` lets `read_text` decode as it goes rather than holding a second copy: it appends the longest complete prefix and answers how much it took, leaving a sequence cut short by the chunk boundary for the caller to carry forward. `utf8_sequence_len` distinguishes malformed from merely short, since only the caller knows whether more bytes can arrive.

This is a world version bump, so every recorded invocation is invalidated.

## Two compiler fixes

A variant's discriminant is an i32, but its flat slot may be wider: the Canonical ABI joins that slot across an enclosing variant's cases, so `result<u64, error>` gives the error's tag an i64 slot. The lowering coerces the tag to the joined width.

A tuple over one pack is assignable only when the fixed elements on either side of the pack line up. `[..R, L]` is not `[..R]` and not `[L, ..R]`, both decidable before the pack expands. Without that, `swap_ends([7, true])` read `7` as a bool and `true` as an i32 with no diagnostic, and the arity mismatch reached codegen as an ICE. Two fixtures pin each.

Three WIR type lookups answered `WirType::I32` for a local, field, or element that was not registered. Each now states what the caller owes.

## Geas

`docs/wep-2026-09-20-geas.md` is the design the probe was prerequisite work for: tensor compilation where an axis is a type and a checkpoint's header is its schema. `example/tensor_axes.wado` demonstrates the axis half, which needs no compiler change — a shape error is a compile error naming the axis that disagreed. `docs/research-ml-inference-webgpu.md` records the survey behind it.

The variadic WEP records that one pack per parameter list is the implemented rule, and what a second would allow.

Gale's committed corpus was regenerated; its parsers are byte-identical and only the invocation records moved.


gfx:
Answered the review. Two findings were real, two are skipped with reasons below.

## Fixed

**Tuple elements compare invariantly** (`typecheck.rs`) — real, and it was a compiler bug rather than a style point. Rule 8 compares a generic instance's type arguments at `Position::Invariant`, and a tuple is a generic instance, so the pack pre-check reading the surrounding position accepted what the expanded tuple rejects. Red/green:

```
fn widen<..Rest>(t: [..Rest, Node]) -> [..Rest, EventTarget] { return t; }
```

`Node` satisfied `EventTarget` through resource subtyping, the unexpanded pack left the check `Deferred`, nothing asked again after expansion, and the module reached codegen:

```
Internal compiler error: WIR pipeline generated an invalid core Wasm module
Offending function: widen<[i32]>
Validation error: type mismatch: expected (ref $type), found (ref $type)
```

`pack_shape_conflict` now compares at `Position::Invariant` and no longer takes a position at all. `tuple_pack_element_subtype_error.wado` pins it.

**A local's WIR type** (`translate.rs`) — real. `translate_block` has a live branch for a function that arrives without the lower phase's local allocation, and instrumenting it showed the branch is common, so a read of a local declared only by a `let` was reachable. The fix differs from the suggestion: `declare_locals_from_stmts` records the type it already reads off each `let`, and the read resolves from there. Restoring the `WirType::I32` fallback would type a GC-ref local as i32, which is a miscompile no test catches; `AGENTS.md` bans a fallback whose validity cannot be argued. A read that neither source can answer still fails loudly.

## Skipped

**`assert_eq!` on the probe result length** (`kiln_driver.rs`) — not reachable. `run_probe` walks `request.files()` and pushes one entry per file, so the length is structural and no generator input can move it. Only a broken `CompilerHost` implementation reaches it, which is what `AGENTS.md` says to write an `assert!` for. Converting it to an error would let a broken host produce a wrong answer instead of a loud one.

**Axis identity in the Geas WEP** — the claim is correct: a safetensors header carries tensor names and positional shapes, so `Batch` and `Hidden` cannot be derived from it and two tensors cannot be given a shared axis identity from it alone. Recorded as a known gap naming both ways to close it — a second input declaring the architecture's axis schema, which would join the invocation's file set and its probe, or restricting header-only generation to positional checks. Which one is a design decision, so it is not made here.

Branch distilled again afterwards.

PR #2100, opened 2026-09-20
refactor(compiler): the host supplies the stdlib and saves the ICE artifact, so wado-compiler opens no file

gfx:
`wado-compiler` performs no filesystem I/O. Both remaining reads and writes go through `CompilerHost`, the seam the crate already owns for host I/O, so the compiler's behaviour no longer depends on what the process it is linked into can open.

An internal compiler error now saves its invalid binary where the host says. The CLI creates it in `std::env::temp_dir()`, which exists on every platform the CLI runs on and needs no write permission beside the sources; a host with no filesystem keeps nothing and the panic says so.

## The stdlib

A dev build serves `wado-compiler/lib/` from disk so editing a module takes effect with no rebuild. The host reads those files and hands them over:

- `stdlib::install_dev_stdlib(sources)` pins the table for the process. The first install wins — a run that compiled two versions of a module describes neither — and only that one walks `sources`, so a lazy iterator costs every later caller nothing.
- `stdlib::dev_stdlib_files()` names every file an install has to cover, relative to `stdlib::DEV_STDLIB_ROOT`.
- `stdlib::installed_dev_stdlib()` answers what was pinned, for a caller that has to say which files this process compiled against.

`wado_lsp::host::install_dev_stdlib` is the one reader. Every host installs as it is built (`FilesystemCompilerHost`, the dev-tools host, `MapHost`, the playground's `ProgressHost`), as does `Engine::new` and each command that reaches the stdlib without a host — `wado doc <stdlib module>` and `workspace/textDocumentContent`. A host is built per LSP request, so that lazy reader matters: after the first install nothing is read. A dev build that reaches the table with nothing installed is named at the door rather than serving an empty stdlib.

A release build and any `wasm32` build embed the sources, and every one of these entry points compiles to a no-op there.

## Watching what a run compiled against

`RunCache` observes the pinned stdlib as an input like any other, so `wado test` reports a mid-run edit to `core:cli` the same way it reports one to a fixture.

## The ICE artifact

`codegen::emit_wasm` returns `Result<Vec<u8>, Box<InvalidArtifact>>` carrying the failing bytes, a file stem, and the report. `compile_after_load` hands the bytes to `CompilerHost::save_internal_artifact` and panics with where they landed. The host names the file, and the CLI's creates it through `tempfile` under an opaque name, so nothing planted at a predictable path in the shared temp directory is written through.

https://claude.ai/code/session_01GA1KmDMSDp4jyH4vmScgr1


gfx:
Both findings were real and are fixed in `50c0645ed`. Nothing skipped. Each turned out to be an instance of a class wider than the site named.

**The installer did all its work on every call.** `install_dev_stdlib` is called by every host constructor, `Engine::new`, `doc::run` and `RunCache` — by design, many times per process — and `OnceLock::set` discarded everything after the first. The leak is one instance; the other is that `host_for_uri` builds a fresh host per LSP message (`server/dispatch.rs`, `publish_diagnostics`), so each request also re-read 5.75 MB from disk before throwing it away. Consuming `sources` inside `get_or_init` closes both: every call site passes a lazy `.map()` reader, so after the first install nothing is read and nothing is leaked.

**A write to a path the process did not create.** `save_internal_artifact` builds the file with `tempfile::Builder`, which creates it exclusively under an opaque name, so a symlink planted ahead of it is not written through. The host owns the filename now and the compiler supplies only a stem (`invalid-core`, `invalid-component`). Sweeping the rest of the tree for the same shape: every other `std::env::temp_dir()` site is `#[cfg(test)]`, and all but one already carry the pid — this was the only production instance.

On the benchmark alert above: it is not this PR. `benchmark/fts/fts.wado` compiled at `-O2` is byte-identical between this branch and its merge base `e72a0a50d` (sha256 `73c1b30d2fcac5a6ee54ba49828d8ceba1a5d208cb87224f2b56933e4457764e` both sides), and every line the branch adds is behind `#[cfg(all(debug_assertions, not(target_arch = "wasm32")))]`, which a release build compiles out. All 13 benchmarks moving together by 1.55–1.84x is the runner, not the diff.

---

PR #2104, opened 2026-09-21
feat(stdlib): token generators in core:secure_random, and a UUID 67 bytes smaller

gfx:
`core:secure_random` gives a program the four things it reaches for when it needs a secret: `bytes(n)` for a key, a salt or a nonce, `token_hex` and `token_base64url` for a session token, and `token_from(alphabet, len)` for a one-time code someone has to read aloud. `with_buffered` draws in blocks, so the small draws inside its body can share a host call. The `Random` effect is re-exported, so installing a handler needs no `wasi:random` import.

`core:base64`'s decode tables are Wasm constants, so a program that never decodes carries neither. That lets `core:uuid` draw through `core:secure_random` instead of keeping its own copy of the loop: `example/uuid_v7.wado` at `-Os` is 8516 bytes, 67 below a `core:uuid` that duplicates it, and a base64 encode-and-decode program is 6405, 174 below the loop that built the tables at startup.

## What the tokens are

`token_hex` and `token_base64url` default to 32 bytes. `token_from` draws uniformly: `256 / alphabet.len() * alphabet.len()` bounds the accepted byte values, and a draw above that bound is discarded rather than folded back onto the front of the alphabet.

`bytes(n)` answers with exactly `n` bytes whatever the installed `Random` does — it loops over a short read and truncates a long one. Every token generator draws through it, so none of them can be lengthened or shortened by a handler.

Compare a token against an expected one with the prelude's `eq_constant_time`.

## Effects come from the module that demands them

A caller has to name an effect to install a handler for it or to declare it onward, and `wado-compiler/AGENTS.md` now records where the name comes from: the module whose public signatures carry it. `core:uuid` hands out `Random` and `SystemClock`, `core:temporal` the `SystemClock` its `now()`s read, and `core:benchmark` the `Stdout` and `MonotonicClock` that `Benchmark::run` takes. `core:cli`, `core:fs` and `core:router` already did. A handler for one of those clocks has to name `Instant` too, so `core:uuid` and `core:temporal` hand that out alongside. An effect no caller ever names stays private, which is why `core:log` re-exports nothing: it reads the clock under `#[ambient]`.

Two more conventions join it. A module carries its `#[synopsis]` test in the module itself, since `wado doc` renders it as the module's `## Synopsis`; `core:fs` gains the one it lacked. Tests for what a module exports live in `<module>_test.wado` beside it, which is what `core:cli` gains. `core:prelude` owes neither: every program already imports it, and the e2e fixtures hold its behaviour.

## DCE reads a call's function summary

`remove_dead_global_sets_block` decides whether a dead global's initializer can go with the global by asking `deletable_value`, the same predicate its sibling passes use, which reads a call's whole-function `FnEffect` rather than refusing every call on sight.

`run_dce` computes that summary once per round and hands it to both global passes. A second walk runs only where `unhoist_unobserved_globals` reports that it rewrote a body: those bodies only lose work, so their real summaries shrink, and a table still reporting the larger ones would refuse deletions with no reason left to refuse them.

`core:base64`'s decode tables reach the constant path through the existing dead-global pass; a test rebuilds both from the rules their entries follow and asserts they match.

https://claude.ai/code/session_01FzjxBukQzmN4yT7tKoygNT

---
_Generated by [Claude Code](https://claude.ai/code/session_01FzjxBukQzmN4yT7tKoygNT)_


gfx:
Review round, answered.

**Fixed**

- `with_buffered` promised one host call for every draw inside its body. A short read from `wasi:random` makes a draw reach the host again even under the block size, so the module doc, the synopsis comment and the function's own doc now say small draws *can* share one, and the function names both escapes: a draw larger than the block, or larger than what a short block held. `BufferedRandom::new` states the positive size it asserts, which the rendered doc could not otherwise show.
- The `core:fs` synopsis wrote `if let Err(e) = read_to_string(...)`, so a read that unexpectedly succeeded skipped the only assertion. It binds `unwrap_err()` now.
- The merge conflict against `main`: all 55 conflicted paths were golden WIR fixtures, regenerated with `mise run update-golden-fixtures` rather than hand-merged. `git merge-tree` against `origin/main` is clean, and e2e on the merged tree is 5488 passed / 0 failed.

**Skipped**

- *Reset the synopsis directory before setup* — `missing.txt` occurs exactly once in the tree: the read that expects it to be absent. No code writes it, and `target/core-fs-synopsis` has no other user, so `remove_dir_all` there would guard a case nothing can reach. `AGENTS.md` bans a defence whose validity cannot be argued. The guard that does hold is the `unwrap_err()` above: it fails loudly if the read ever succeeds, rather than passing in silence.

---

PR #2105, opened 2026-09-21
feat(geas): check an ONNX graph's axes at build time, as a Kiln generator

gfx:
An ONNX graph's axes are checked when the module compiles. Geas is a Kiln
generator: it reads a `.onnxtext` file, names each tensor's axes from the
`layout` patterns at the `use` site, discharges every operator's extent
conditions, and emits one marker type per axis so two tensors that share an
axis share the type that names it. A graph whose extents do not agree stops
the build with the axis and both lengths:

```
kiln: generator error: invalid schema: geas: 'example/mlp.onnxtext':
T = MatMul: axis Hidden is 64 on the left and 32 on the right
```

`package-geas/example/mlp.wado` is that end to end — a two-layer MLP whose
tensor types come from the graph, where passing the output where the input
belongs is a type error.

## The generator

`generate` reads the graph, infers shapes, emits the module. Its `Options`
carry the `layout` patterns, so the axis schema is typed at the use site and
lives nowhere else. There is no tier below this one, so an extent the algebra
cannot discharge is a build error rather than a runtime check.

## The shape algebra

`Shape` carries axes of a name and an `Extent`, static or symbolic. `MatMul`
and multidirectional broadcasting are ONNX's, and every failure names the node
and both extents. The operator set is three families: `MatMul`, five binary
elementwise operators, and nine unary ones; every other operator stops the
build naming itself.

## The ONNX reader

`package-geas/grammar/Onnx.g4` is the textual format's grammar, verified
against `onnx/defs/parser.cc` rather than written from memory, and the reader
walks the CST Gale builds from it. The parser is committed under
`package-geas/generated/`, since a Kiln generator cannot itself run a
generator; `onnx_grammar_test.wado` holds the invocation that keeps the
grammar and the committed module equal.

## A Kiln `Options` field type from another module

A generator's `Options` may name a type the generator shares with its own
library. The generator's CM surface carries its submodules' public type
declarations, and a CM named type the interface FQ does not place resolves
through its own recorded module, so the record's field reaches the named-type
registry with a declaration behind the name.

## Gale's `RK_` spelling

A rule's node-kind constant is its snake-case name uppercased, so `valueInfo`
reaches `RK_VALUE_INFO`. `gen_util::rule_kind_const` is the one place that
spells it, which is what keeps the declaration and every reference one name.
The generated corpus is regenerated for it.

## The WEPs

The Geas WEP states what stage 0 does and what it does not: the ONNX graph as
the notation, the axis schema as an invocation option, and `f16`/`bf16` as
storage-only primitives the prelude does not carry yet, which is roadmap step
2. The Kiln WEP adopts nested generators and records the cycle rule they need.


gfx:
All eight findings verified against the code; all eight were real. They fall
into five classes, and the fixes are at the class rather than the site.

## A compiler panic, not a wrong binding (`types.rs` 145-146)

The claim understated it. `register_unique` is an `assert!`, so two submodules
declaring one public `struct` name **panic the compiler**:

```
CmInterfaceRegistry: duplicate struct registration for `Shared` in interface
`kiln:generator/generator@0.1.0`. Each (interface, name) pair must be
registered exactly once.
```

`--lib` never reaches that assert because `first_duplicate` bails with a
diagnostic first — a check gated on `lib_world.is_some()`. Every published type
reaches the registry under one interface FQ, so the gate is now
`synth_world_fq.is_some()`: whoever synthesizes a world owes the check.
`duplicate_public_type_across_generator_modules_is_diagnosed` covers it.

I swept the other three `lib_world.is_some()` gates. The duplicate-`export fn`
check is `--lib`-only by construction (a generator's exports come from one
module). The resource-handle refusal guards another emitter panic, but no
resource type is reachable from `core:kiln`, so a generator cannot construct
the input — left alone rather than widened speculatively.

## The graph's declared boundary was never discharged (`infer.wado`)

Both halves reproduced. `infer` now walks `graph.outputs`:

- an output no node produces is `Z: no graph input, initializer or node defines this`
- a declared extent that disagrees is `Y: axis B is declared 7 and produced 10`
- a declared rank that disagrees is `Y: the graph declares rank 1 and the nodes produce rank 2`
- a declared `?` claims nothing, so it discharges against whatever was produced

`emit_boundary` reads the shape it is owed instead of returning `Ok(())` — that
branch was a defence with no contract behind it, and its doc comment named a
caller that did not exist.

## The grammar accepted a subset without saying which

Both productions are real gaps against `docs/Syntax.md`, and I swept the BNF
production by production rather than adding the two named. Both landed, with
parser and reader tests.

The trailing attributes are a second **optional**, not a second alternative:

```
node : nodeLabel? outputNames? EQ qualifiedId attrs? LPAREN inputNames? RPAREN attrs? ;
```

The suggested two-alternative form makes `node` a multi-alternative rule whose
alternatives share an opaque rule-ref prefix, which is an ATN-class decision on
the hottest rule in the grammar. `gale dump` confirms the optional keeps `node`
`Simple` with no new ambiguous decision (3 before, 3 after).

What stays out is local functions, and with them the `@name` attribute
reference only a function body can carry. Both are stated in the grammar header
and in the WEP's known gaps.

## An adopted decision that is not yet in the code has to say so

Three findings, one class — and the WEP convention is that a Decision section
states the design while the Roadmap states what will be done, so the fix is a
sentence at each decision naming the step that closes it, not a rewrite:

- nested generators: the host cannot run an inner generator yet; `package-geas`
  commits its parser and imports it
- the checkpoint in `inputs`: read by roadmap step 3
- `architecture`: roadmap step 6 writes the tables

One correction: the Geas example does **not** use the nested path. It imports
the committed generated parser precisely because that path does not work.

## Prose

`every operator example/mlp.onnxtext uses passes...` is a garden path; fixed as
suggested.

## Also found on the way

The opset a model declares is read by nothing. `opset_import` says which
operator versions the graph is written against and the shape rules assume one
without consulting it. New known gap.

Nothing was skipped.

PR #2107, opened 2026-09-21
fix(kiln): diagnose a duplicate public type name instead of asserting

gfx:
A Kiln generator whose submodules declare one public type name twice gets a
diagnostic naming the type. That name reaches the CM interface registry, which
registers each `(interface, name)` pair exactly once and asserts when asked
twice, so the check that refuses the duplicate belongs to whoever synthesizes a
world — `--lib` and a generator alike.

Two more things a graph declares are now checked, and the ONNX grammar takes
the productions it was short of.

## The graph's declared boundary

`infer` discharges the output declarations the file itself carries:

- an output no node produces is `Z: no graph input, initializer or node defines this`
- a declared extent that disagrees is `Y: axis B is declared 7 and produced 10`
- a declared rank that disagrees is `Y: the graph declares rank 1 and the nodes produce rank 2`

A declared `?` claims nothing, so it discharges against whatever the nodes
produced. `emit_boundary` reads the shape every boundary tensor is then owed.

## Two ONNX productions

A node's attributes may follow the input list, and an initializer's data may
live in a file the text names (`= ["location": "weights.bin"]`). The trailing
attributes are a second optional rather than a second alternative, which keeps
`node` token-led instead of making it an ATN decision on the grammar's hottest
rule. The committed parser under `package-geas/generated/` is regenerated.

What stays out is local functions, and with them the `@name` attribute
reference only a function body can carry.

## The WEPs

Each decision that is adopted but not yet in the code says so where it is
stated: nested generators, the checkpoint in `inputs`, and the architecture
table each name the roadmap step that closes them. The opset a model declares
is read by nothing, which is a known gap.


gfx:
Review answered. Four findings, verified against the current tree; three fixed, one skipped.

## Fixed

**The ONNX node rule accepted attributes on both sides of the input list.** Verified concretely first: `Y = Transpose<perm = [1, 0]>(X) <other = [0, 1]>` parsed, and `attrs.len() == 1` — the second list silently replaced the first. The rule is now two alternatives starting on different tokens (`<` and `(`), so the choice stays token-led rather than becoming a decision Gale has to scan for, and a node with two lists is a parse error. Pinned by `test "a node carrying attributes on both sides does not parse"`.

I did not take the suggested patch: `(attrs? LPAREN … RPAREN | LPAREN … RPAREN attrs)` shares `LPAREN` as a first token across its alternatives, which is exactly the ambiguity that costs a `gale dump` decision class.

**`read_value_info` dropped an external initializer.** A `["location": "weights.bin"]` initializer was parsed and then discarded, leaving a tensor whose weights silently came from nowhere. It now names the initializer and refuses. Pinned by `test "an initializer pointing at another file names itself rather than being dropped"`.

**The duplicate-name diagnostic kept its `--lib` wording.** The gate was widened to everything that synthesizes a world (a generator surface included) but the text still said "library type" / "a library's public types". It now reads "public type `X` is defined in more than one module; the types a component publishes must have distinct names", and the assertion in `cli.rs` follows it.

Also tightened the Geas WEP: `architecture` was described in the present tense ("selects a table Geas ships") for something that is not read at all. It now says so, and names roadmap step 6 as what writes those tables.

## Skipped

**`emit.wado:56` — `shapes.get(vi.name).unwrap()`.** `infer` gives every boundary tensor a shape and refuses a graph where an output has no producer, so the `None` arm is unreachable. AGENTS.md "Contracts, not defences" is explicit that inventing a fallback for a case you have not shown to be reachable turns a broken call into a wrong answer no test catches. The invariant is stated at the call site and enforced in `infer`, which is where it belongs.

`wado test package-geas`: 79 passed, 0 failed. `cargo test -p wado-cli --test integration`: the duplicate-name tests pass.

PR #2108, opened 2026-09-21
refactor(stdlib): one import path per effect, and a block size on `with_buffered`

gfx:
An effect has one name and one place to get it from. `core:cli` hands out `Stdout` and `core:fs` hands out `Preopens`, because those effects are theirs; everything else that merely performs an ambient effect imports it privately, so `use { MonotonicClock } from "wasi:clocks"` is the only way to write it and the import line says which clock is meant.

`with_buffered` takes the block size, so `core:secure_random` is usable end to end without the caller ever naming `Random`.

## The rule

`pub use` preserves identity: re-exporting an effect abstracts nothing, it adds a second path to one type. And an effect is part of the public signature — the caller has to know which one it is to install a handler — so the second path buys no encapsulation either, only the question of which `Random` or which clock a given import means.

`wado-compiler/AGENTS.md` states this under **Standard Libraries**. `core:benchmark`, `core:uuid`, `core:temporal` and `core:secure_random` import their effects privately under it; `core:cli`, `core:fs` and `core:router` are unchanged.

## `with_buffered(body, pool_bytes)`

```wado
with_buffered(|| { ... });        // the default 4096-byte block
with_buffered(|| { ... }, 32);    // a block of 32
```

`BufferedRandom` stays public, so a span that is not a closure can still take the handler by hand — that caller names `Random` from `wasi:random` like any other.

Generated stdlib docs were regenerated.

https://claude.ai/code/session_01FzjxBukQzmN4yT7tKoygNT

---
_Generated by [Claude Code](https://claude.ai/code/session_01FzjxBukQzmN4yT7tKoygNT)_


gfx:
Both findings land on `docs/stdlib-core-secure_random.md`, which is `linguist-generated` — `wado doc` writes it from the doc comment in `wado-compiler/lib/core/secure_random.wado`, so nothing is edited there directly.

**`pool_bytes` precondition** — staying as it is. `AGENTS.md` > General Rules: "Invariants: state them as assertions, not comments. An assert is checked; a comment goes stale." `assert pool_bytes > 0;` in `with_buffered` is that statement, and a doc sentence repeating it is the duplicate the rule exists to prevent.

**`mise run format`** — already ran, before the commit and again just now; the tree comes back clean. The `Tidy` CI job checks the same thing.

---

PR #2112, opened 2026-09-21
feat(traits): a supertrait clause may write `Self::Assoc`, and a projection keeps its identity everywhere it travels

gfx:
`trait Constrained: From<Self::Base>` works. `Self` in a supertrait clause is the implementing type, so an impl binding `Base = String` owes `From<String>`, and a generic body under `T: Constrained` reaches that impl: bound check, static call and instance call all land on it. This is the shape the serde constrained-type design asks for, and a struct wrapper now gets it end to end.

Underneath, a projection is a type with an identity rather than a string. `T::Base` travels as a base, a member name, and the trait that declared the member, from the elaborator through to the monomorphizer. Two traits declaring `Item` on one type are two answers, and a call on `X::Item` reaches the one its own trait binds.

## Reading a clause in the space it was written in

A supertrait clause writes its arguments in the *declaring* trait's parameter space, which is not the space of whatever frame reads it. The clause therefore travels with the chain of clauses that reaches it, each step written in the space of the step before. A reader carries its own arguments down that chain, resolving each step at what the step before answered, and reads the clause in the space it arrives at.

Walking is what answers a projection. `trait Uses<X: Elem>: Needs<X::E>` read at `X = P::Inner` settles `X` before anything reads `X::E`, so the base is a type by the time the member is asked for. No spelling denotes a projection off a type, which is why nothing re-spells the clause.

`ParamSpace` is what a trait's declared parameters stand for at a frame; `Elaborator::in_space` runs a body inside one. Every reader of an inherited bound goes through it — bound checking, method lookup, associated-type projection, and the impl-side supertrait check alike.

## A projection's identity, end to end

`TypeHead::Projection { base, assoc, owning_trait }` carries a projection through `FqTypeName`, so substitution reaches it structurally: `FqTypeName::rewrite` replaces every position a type stands in, and a projection nested inside an argument is answered as one standing alone is. `Make<List<T::Base>>` reaches `impl Make<List<String>>`.

At monomorphization, `type_at_instance` resolves a projection whose base is itself a projection, recursing until a base the frame binds turns up. `C::Iter::Item` answers once `C` does.

`Resolution::Projection(AstId)` records a `T::Assoc` type site as reading a member off a binder, rather than leaving it unresolved. A site that must know which declaration it names asks that, and a bare member name no longer reaches whatever else the module happens to call it.

## Scope

`docs/wep-2026-07-27-super-traits.md` and `docs/wep-2026-09-01-trait-resolution.md` state the rules above. The serde WEP's constrained-type gap narrows to newtypes, whose auto-generated `From` the supertrait check does not see.

The AST-level type-parameter substitution helpers are gone: deciding semantics on written spellings is what allowed the class of defect this closes. The elaborator-rearchitecture WEP tracked them as a known gap, which this closes.

Fifteen e2e fixtures cover the clause shapes and the dispatch paths through them, including two rejections.

https://claude.ai/code/session_019vGS1FrXwzMQdrbs2SaZF2

---
_Generated by [Claude Code](https://claude.ai/code/session_019vGS1FrXwzMQdrbs2SaZF2)_


gfx:
## Review response

Four of the six findings were real. All are fixed in `00bce85c8` and `0b1f3e0b4`.

### Fixed

**`reads_a_projection` skipped function types.** Real. The match answered `false` for every shape it did not list, so `Uses<fn(P::Item) -> i32>` hid the projection its argument carries.

**The inherited-bound skip compared spellings.** Real. `enforce_type_arg_bounds` matched a clause's written name against the direct bounds, so two traits a module calls by one name read as one. It compares declarations now, falling back to the spelling where a bound names none, as `merge_bound` already does.

**`push_unique_inherited` collapsed two chains.** Real. Two clauses writing one supertrait alike in their own parameter spaces reach different arguments, and the chain a reader walks is what tells them apart, so the chain is part of the key.

**Each `via` step's identity.** Real, and the same defect one level down: `inherited_space` resolves each step to a **declaration** and reads that declaration's defaults for the positions the step leaves out, while the key recorded only the written arguments. The suggested patch keys on the clause's `AstId`, which over-discriminates — every syntactic site becomes distinct, and the bare-vs-constrained merge in that same function would never fire again, dropping an associated-type constraint. The key is the declaration instead: `via` now carries `ViaClause { bound, decl }`, `inherited_space` reads `step.decl` rather than resolving the step's name a second time, and the key is `(decl, written arguments)` per step.

**A WEP contradiction**, separately: the super-traits Known gaps said `T: Sink` leaves `T::Item` unresolved, which Elaboration in the same document contradicts. It resolves; the gap is gone.

### Skipped

**Omitted defaults before supertrait substitution.** Not reproducible. `trait Derived<T = String>: Base<T>` under `X: Derived` resolves `Base<String>` and runs.

### The class behind them

Three of the four are one defect: a semantic decision keyed on a written spelling instead of the declaration it names. The fourth is a second class — a predicate over `ast::Type` that spells its own recursion answers `false` for the arm it forgets. `Type::any` is the one structural walk now, and each predicate passes only what it asks at a node. Three siblings carried the same hole and are fixed with it: `mentions_empty_tuple` let `ns::Gen<[]>` and `fn([]) -> T` past the CM boundary check, `lib_sig_uses_named_type` read a namespaced type as naming no user type, and `return_type_has_future` missed a `Future<T>` under either.

### One known gap, recorded

Closing the function-type hole surfaced a deeper one, now written down in `docs/wep-2026-09-01-trait-resolution.md`: a function type stands in a trait's name as one opaque spelling, holding no position a substitution reaches, so `U: Uses<fn(P::Item) -> i32>` passes `wado check` and fails where the call is monomorphized. `Uses<P::Item>` and `Uses<List<P::Item>>` both dispatch. Left as a gap by the author's call.

Validated at each step: `mise run format`, `cargo clippy -- -D warnings`, and e2e 5512 passed / 0 failed.

---

gfx:
## Review response

Verified each finding against the current code before deciding anything.

### `scope.rs:461` — merging drops a chain's associated-type constraints — skipped, not real

Both chains' constraints are enforced today. With `Left: Base<Item = i32>` and `Right: Base<Item = String>` on one type, and `T: Left + Right` at a call:

- an impl binding `Item = i32` is rejected on the `String` constraint;
- an impl binding `Item = String` is rejected on the `i32` constraint.

Neither is dropped, so nothing is lost to the merge.

### `orchestration.rs:111` — the diagnostic calls a projection a type parameter — fixed

Reworded to cover both forms. The other four arms pairing `Binder` with `Projection` return `None` and word nothing, so this was the only instance.

### `type_resolution.rs` — `Self` unbound while the bound closure resolves — a real asymmetry, and it led somewhere worse

The claim holds. `Self` in a bound on `T` names `T`, and that is true in the associated-type constraint position (`T: Sink<Cb = Self::Item>` compiles and runs) but not in the trait-argument position (`T: Uses<Self::Item>` reports `unknown type 'Self::Item'`). One `Self`, one bound, two answers by where it was written.

Probing that position with the projection spelled off the parameter itself found a compiler crash:

```wado
fn drive<T: Src + Uses<T::Item>>(t: &T, a: T::Item) -> i32
```

overflowed the stack. Reading that argument asks for `T`'s bound closure, which is what reads it, so the walk re-entered itself without end. The existing re-entrancy guard covers a different path.

**Fixed**: a closure cannot answer itself. The re-entrant ask gets nothing, and `T::Item` stays the abstract projection its instantiation settles. `bound_trait_argument_projects_off_its_own_parameter.wado` covers it and passes.

The `Self::Item` spelling of the same position still reports `unknown type`, kept as `bound_self_projection_in_trait_argument.wado` under `#[TODO]` so the suite reports it as pending and hard-fails the day it starts working.

### Verification

`cargo clippy -p wado-compiler --all-targets -- -D warnings` clean; e2e 5594 passed, 0 failed.

---

PR #2115, opened 2026-09-21
fix(optimizer): two wrong-code defects the nightly EMI campaign found

gfx:
Hot-field scalarization keeps a field in a scalar only where nothing else holds a handle on the object, so a closure that writes a captured local through its environment struct is seen. Parameter SROA retypes a whole-param read to the field the local now holds, so the multi-value split projecting off that read names a field the type carries. Both are reachable from ordinary source, not only from an EMI mutant: the closure case prints the wrong sum at every optimization level, and the SROA case is a `[WIR emit]` panic.

## The handle a pass cannot see is still a handle

A GC local stored bare into an object being built gives that object a second handle on it, and a write through the handle bypasses any scalar the field is held in. The alias scan reaches every allocating expression kind — struct, tuple, list and variant construction — so each publishes its operands.

A closure's by-reference capture is one of these. The captured local's `Box` cell rides the environment struct, and the call that writes it names no `&mut` for a call-site scan to find. `tests/fixtures/opt_hfs_closure_captured_write.wado` is that shape.

## A read of an SROA'd param carries the field's type

`rewrite_param_reads` keeps a read of the whole parameter where it forwards to another scalarized position, and retypes it to the scalar the local holds. A later round otherwise reads the wrapper type as licence to project the field a second time, and codegen panics on a field the field's own type does not carry.

## Every field_scalarize walker is the visitor's

Four scans in `field_scalarize` walked the arena by hand, each re-deriving descent and each missing a different shape — the alias scan's missing struct-literal arm is what let the first defect through. They are `NirRefVisitor` implementations now (`ParamUsageScan`, `AliasScan`, `FieldAccessScan`, `CallTouchedScan`), so a shape with no arm is still walked and the arms name only what the analysis cares about. Three copies of "peel a `&`/`&mut` and read the `Local` underneath" collapse into one `local_place`.

The WIR golden fixtures are regenerated.

## Findings are re-runnable whichever stage finds them

The EMI harness recorded a finding only from its mutation stage. A calibration finding — an empty guard that crashes the compiler, which is what caught the SROA defect — left a campaign report naming the file and no source to re-run, so the CI artifact could not reproduce it. Both stages now narrow to the guards that carry the finding and write the reduced program out.

Each stage keeps its own definition of wrong code: a guard that moves the output is a finding at mutation and a disqualification at calibration, and what a stage does not call a finding is neither reduced nor written.

A `[WIR emit]` panic names the function being emitted, which is what turned an 1100-guard mutant into a three-guard reproducer here.

https://claude.ai/code/session_014Tjos7wbVTVpuQA7vLYu3H

---
_Generated by [Claude Code](https://claude.ai/code/session_014Tjos7wbVTVpuQA7vLYu3H)_


gfx:
## Benchmark alert: not this PR's

The `Benchmark` check itself concluded **success** — 1.53–1.79× is over the 150% alert threshold but under the 200% fail threshold, so the step warned only. The comparison is against the last recorded `main` datapoint, measured on a different runner at a different time, and the runtime suite carries no control rows, so a uniform slowdown across every workload at once is what a slower runner looks like.

Measured locally instead, with two `wado` binaries differing **only** in this PR's compiler sources, runs interleaved (not in blocks) to beat host drift. Verdict is range overlap, not delta:

| row | base best | head best | base range | head range | verdict |
| --- | --- | --- | --- | --- | --- |
| microgpt train | 418.3 | 428.7 | 418.3–498.8 | 428.7–474.0 | overlap |
| microgpt infer | 294.5 | 297.8 | 294.5–344.3 | 297.8–321.4 | overlap |
| zlib compress | 9.056 | 8.873 | 9.056–10.718 | 8.873–10.071 | overlap |
| zlib decompress | 2.398 | 2.421 | 2.398–2.526 | 2.421–2.582 | overlap |
| json ser | 1.536 | 1.483 | 1.536–1.736 | 1.483–1.791 | overlap |
| json de | 7.421 | 7.513 | 7.421–8.297 | 7.513–8.896 | overlap |

(ms/iter, 6 reps each on microgpt, 4 on zlib/json.) Every row overlaps — no regression. An earlier sequential-block run did read one row as disjoint; re-running it interleaved showed identical binaries spreading ~9% on this host, which is what that reading was.

Also worth recording: with `field_scalarize.rs` reverted to `main` the emitted Wasm for all three benchmarks is **bit-identical to this PR's head**, so the visitor refactor and the HFS conservatism cost nothing in codegen here. The whole `.wasm` delta is the `sroa_param` retype.

## `Test (Core)` — fixed

`remarks::scalarized_struct_copy_is_not_remarked` was red, and it was this PR's. The `sroa_param` retype folds `$sroa_b_x` away, which turned the test's `` `${a.x} ${b.x}` `` into a fully constant region; the const-region remark then correctly reported that `i32::fmt_decimal` still runs there.

That missed fold is pre-existing, not new — on `main`, `println(`${1} ${9}`)` already emits the same remark. The test's program had simply never reached that shape. Pushed a fix that puts one field behind `builtin::black_box`, so the test keeps asserting what it is about (the scalarized copy is not remarked) without being coupled to how far constant folding reaches.

Full `integration` target: 782 passed, 0 failed.

---

PR #2116, opened 2026-09-21
fix(compiler): bind a `&mut` destructure to the scrutinee, and check every attribute

gfx:
A `&mut` destructure binds the scrutinee's own storage at any nesting depth, so `let { inner: { items } } = &mut o; items.push(2)` writes into `o`. Every `#[…]` the compiler recognizes is declared in one schema — its name, the declarations it may sit on, the arguments it takes — and a name, position or argument list outside it is a compile error, so `#[expect_trab]` is reported rather than dropped without a word. Diagnostic codes name one fault each: a receiver mismatch, an unsatisfied trait bound and a missing return are three codes, and a fixture states which one a rejection carries. Importing a Component Model component whose export takes no arguments works.

## `&mut` destructuring

`lower_let_pattern` peeled the scrutinee's reference into a `Deref` and bound the result to a temp the value-copy fold treated as owned, which asked for a deep copy of the referent. The temp keeps the reference now, the projections read through it, and it is owned only when it holds none.

The same projection loop existed once in `lower_let_pattern` and again in `lower_pattern_to_lets`, so a nested sub-pattern bound its temp by value however the outer one was bound. Both paths go through one tuple and one struct lowering, and a value that only reads a place is re-read rather than bound. `lower_variant_pattern` joins them: it binds the payload temp either way and guards the payload extraction alone, so a scrutinee whose payload no pattern takes still evaluates.

## The attribute schema

`wado-compiler/src/attribute.rs` holds every attribute's name, the targets it may sit on and the arguments it takes. `Analyzer::check_attributes` walks every attribute a module writes and reports `UNKNOWN_ATTR` or `ATTR_MISUSE` against it. A misplaced attribute names where it does belong, and one written `#[x]` where it takes `#[x]` is told the spelling rather than only the place.

A shape no central rule captures is declared `AttrArgs::Read` and checked where the attribute is read: reify for `#[param]`, `#[result]`, `#[retain]`, `#[immediate]` and `#[wire]`, the parser for `#[cm]` and `#[canonical]`. Every reader names its attribute through a constant in that module, so a name is spelled once.

`impl`, `use`, and an impl's associated types and consts carry their attributes through the parser, the schema walk and the formatter. `#[allow(…)]` reaches every declaration, expressed as an empty target list so a new `AttrTarget` belongs to it without being added.

## Diagnostic codes

`Code` splits into 65 variants with no dead entry and no two sharing a name. The type errors that shared `TYPE_MISMATCH` separate into `ARITY_MISMATCH`, `NOT_CALLABLE`, `NEEDS_TYPE_ANNOTATION`, `TRAIT_BOUND_NOT_SATISFIED`, `AMBIGUOUS_CANDIDATE`, `TRAIT_NOT_IMPORTED`, `TRAIT_DECL_INVALID`, `RECEIVER_MISMATCH`, `STRUCT_FIELD_MISMATCH`, `MISSING_RETURN`, `CLOSURE_INVALID`, `CM_BOUNDARY_TYPE` and `EFFECT_HANDLER_INVALID`. Name resolution gains `METHOD_NOT_FOUND` and `IMPORT_NOT_FOUND`. The lexer and parser stay together under `INVALID_SYNTAX`, since a reader reaches for the position and the message there, not the code.

`#[allow(...)]` reads a separate namespace (`ast::lint`), so no user source names a `Code`.

## Where a language test goes

A fixture states a rejection two ways: `compile_error` matches the whole report, position included, and the new `compile_error_codes` names the `Code` it was raised under. The compile-error integration tests move to fixtures on that basis, leaving `tests/integration/compile_errors.rs` with what no fixture can state — the `CompileError` the CLI prints, and the source file a diagnostic is attributed to. `AGENTS.md` says which of the two a new test wants.

---
_Generated by [Claude Code](https://claude.ai/code/session_015Z4ahQ4ppYkzUVFxkCTZg7)_


gfx:
Fixed in 61cf39fd7.

**The class:** an item's attributes were reachable through two accessors, and `ast::Item::attrs()` grouped five attribute-carrying variants into an arm returning `&[]` — `BuiltinTypeDecl`, `Use`, `Impl`, `World` and `Test`. It names each of them now, leaving only `Item::Error` empty. `unparse::item_attrs` asked the same question a second time and is gone; `get_item_first_line` reads `Item::attrs()`.

**One correction to the finding.** The symptom it names is not reachable. `Item::attrs()` has a single caller, the block-local-item loop in `resolve.rs:712`, and a local item is only a `struct` or a `type` — both of which already answered correctly. `Item::Impl` is excluded a line earlier regardless, and a `use`, `world` or `test` cannot be a block-local item. So `#[allow(shadowed_name)]` was not being ignored anywhere. What made it worth fixing is a public accessor documented as "the attributes written before the item" that answered for none of five, which is the trap the next caller walks into.

**No test added.** `test_format_keeps_an_attribute_on_impl_use_and_assoc_members` already covers the accessor: its second item's blank-line anchor is that item's attribute, and I confirmed the test fails when `Item::attrs()` is made to return `&[]` for `Item::Impl`. A test I first wrote for this — a doc comment above the attribute — passed under the same sabotage, because `unparse.rs:500` skips `get_item_first_line` when the last leading comment is a doc comment. It guarded nothing, so it is not in the commit.

Clippy is clean and the 169 formatter tests pass.

---

PR #2117, opened 2026-09-21
fix(compiler): a tuple zip over type packs can be bound to a local

### LINKED ISSUE #2111: ICE: binding a tuple `.zip()` result to a local emits invalid core Wasm (exact struct ref where a nullable one is expected)
## Summary

A tuple `.zip()` whose result is bound to a `let` and then iterated generates an invalid core Wasm module. Iterating the same expression inline compiles and runs.

## Reproduction

```wado
fn count<..A>(a: [..A], b: [..A]) -> i32 {
    let mut n = 0;
    let t = [a, b].zip();
    for let p of t {
        n += 1;
    }
    return n;
}

test "zip bound to a local" {
    assert count([1, "x"], [2, "y"]) == 2;
}
```

```
thread 'tokio-rt-worker' panicked at wado-compiler/src/codegen.rs:60:5:
Internal compiler error: WIR pipeline generated an invalid core Wasm module
Validation error: type mismatch: expected (ref null $type), found (ref (exact $type)) (at offset 0x5ef)
Offending function: func #13 count<[i32,core:prelude/string.wado/String]>
Failure at body offset +0x0061 (`>>` marks the failing operator):
       +0x0059  RefAsNonNull
       +0x005a  StructNew { struct_type_index: 8 }
       +0x005d  RefAsNonNull
       +0x005e  StructNew { struct_type_index: 13 }
    >> +0x0061  LocalSet { local_index: 3 }
       +0x0063  Block { blockty: Empty }
       +0x0065  LocalGet { local_index: 3 }
```

The `StructNew` result is an exact reference, and the local the transposed tuple is stored into is declared with the nullable type, so the `local.set` does not validate.

## What works

The same transpose iterated inline compiles and the test passes:

```wado
for let p of [a, b].zip() {
    n += 1;
}
```

So the defect is in giving the `zip` result a local, not in the transpose itself.

## Scope

Pre-existing: reproduces on `origin/main`, independent of the multiple-pack work. The single-pack `zip` above is the smallest case; a `zip` over two distinct packs is rejected at elaboration and never reaches codegen.

## Fix notes

An e2e fixture under `wado-compiler/tests/fixtures/` reproducing the `let`-bound form, then the local's declared type (or the `StructNew` result type) reconciled so the store validates.

gfx:
`[[..A], [..A]].zip()` has a type a `let` can hold: a tuple of one pack whose element `k` is `[A_k, A_k]`. Binding the transpose to a local and iterating it from there compiles and runs, as iterating it where it is written does. Closes #2111.

## The transposed type

A tuple transposes slot by slot, and a pack slot yields a pack. `[[i32, ..A], [String, ..A]].zip()` puts the column `[i32, String]` ahead of a pack whose element `k` is `[A_k, A_k]`. Naming the pack rather than the positions it expands to is what makes the result writable while the pack is still open: expanding it there would fix an arity the pack has not been given.

## One transpose

`TypeTable::transposed_tuple` answers the whole shape from the rows' `tuple_layout`, so a fixed slot and a pack one are read the same way. `transpose_tuple_expr` builds the tuple literal and takes its column types from that answer, and both the elaborator's inline expansion and the monomorphizer's deferred one call it — the expression and the type it carries have one source.

## Diagnostics

Rows of unequal length are no `zip` at all, and the message says so. The "every row must be the same length" diagnostic speaks only where a row carries a pack, which is the case where the lengths cannot be compared, so mismatched rows of a concrete tuple are reported once by the message that names the method.

https://claude.ai/code/session_01BkCgjzcKLMfWwgNeWTkKxv

---
_Generated by [Claude Code](https://claude.ai/code/session_01BkCgjzcKLMfWwgNeWTkKxv)_


gfx:
Two findings: one fixed, one skipped.

## Fixed — a `&` receiver reached a helper that admits only a tuple

The finding named the monomorphizer's call site. The class behind it: reify builds the `TupleZip` node over the receiver as written, so its type keeps any `&`/`&mut` the source put there, while the shape the zip was admitted on is the peeled one. The two callers of `transpose_tuple_expr` disagreed about which of the two they hand it.

It reproduces from source, and on `main` as well, one phase later:

```wado
fn f<..A>(t: &[[..A], [..A]]) -> i32 {
    let z = t.zip();
    ...
}
```

`transpose_tuple_expr` now peels the receiver's type itself and takes no shape argument, so there is nothing left for a caller to get wrong. `variadic_zip_ref_receiver.wado` covers the bound and the inline form.

## Skipped — pack indices cannot disagree where the assert stands

Naming one pack twice in a scope takes a method pack shadowing its impl's, and there the impl's pack resolves to a scalar `TypeParam` rather than a second `TypePack`:

```
fn "Box2::mix"<A>(self: &Box2<A>, other: [..A]) -> i32 {
    let x: [A] = self.items;      // Fixed slot
    let y: [..A] = other;         // Pack slot
```

The layouts differ, so those rows are rejected by the existing unequal-rows diagnostic before the assert is reached. The assert states an invariant that holds, which this repo asks for in place of a silent `None`.


---

gfx:
## Review response

### Fixed — the receiver ran once per cell (`2782bcc`)

Real, and wrong code rather than a style point. `rows().zip()` over a 2×2 evaluated `rows()` four times:

```wado
global mut calls: i32 = 0;
fn rows() -> [[i32, i32], [i32, i32]] { calls += 1; return [[1, 2], [3, 4]]; }

test { let z = rows().zip(); assert calls == 1; }   // calls: 4
```

Both expansions had it — the inline transpose in reify and the deferred `TupleZip` the monomorphizer expands — so it was the class, not the one site the comment names.

The class is "a builder that reads one source expression once per cell must bind it first", and this repo already closes it twice: the `..base` functional update uses `hoist_once`, and the variadic for-of unroll emits `let $tuple_N` unless the iterable is already a place. The transpose was the instance that did not. It is now bound with that same `hoist_once`, in the `zip` arm ahead of the choice between the two expansions, so neither phase can duplicate it and the monomorphizer needed no change of its own. `transpose_tuple_expr` asserts that what reaches it is a local read.

A sweep by shape for other builders cloning a source into each generated cell turns up no further instance.

`variadic_zip_receiver_evaluated_once.wado` covers both paths and asserts the transposed values, so it fails if the binding regresses or the transpose is wrong. E2E: 5568 passed, 0 failed.

### Skipped — the pack-index assertion in `transposed_tuple`

Carried over from the previous round, where it was raised as the claim that `tuple_layout` compares pack names only, so two same-named packs with different indices could reach the `assert!`. Checked by construction: a method pack shadowing its impl's gives one row a `Fixed` slot and the other a `Pack` slot, the layouts differ, and the unequal-rows path answers before the assertion is reached. The suggested `return None` would trade a checked invariant for a silent wrong answer.

### Not a finding — the cancelled CI runs

The two "Test" failures on `6a8c1dd` and `f231f8b` were cancellations, not test failures: every job reports `cancelled`, and the aggregate gate fails on that. `f231f8b` was superseded by the tidy bot's own `4864298`, and `6a8c1dd` by my next push.

---

gfx:
@coderabbitai review

Reviews were auto-paused during the receiver fix and the distill passes. The branch has since merged `main` (conflicts were golden WIR only, regenerated), so this asks for one review of the final state.

---

PR #2120, opened 2026-09-22
perf(kiln): 64% off a cold Kiln build — 92s to 33s

gfx:
A cold `package-geas` build takes 33s where it took 92s, and the Wasm it emits is byte-identical. The win is in the generator chain: `mlp.wado` pulls the geas generator, which pulls the gale generator, and compiling that generator was 88% of the wall clock. Two things paid for it — a Kiln policy that bought the wrong thing, and post-loop optimizer fixpoints that re-scanned the whole module every round.

Measured with the `--log-level debug` span stream on `--no-cache`, which is also where the numbers below can be re-read.

## Generator opt level

A generator is a build-time tool. Its opt level never reaches the artifact — the Wado it emits is byte-identical at O0, O1 and O2 — so all it buys is generation speed. On the heaviest grammar in the tree (`Wado.g4`) generation takes 0.19s at O1 against 0.16s at O2, while the generator's own compile takes 57s at O1 against 93s. Three hundredths of a second of generation for thirty-six seconds of compile.

O2 is worse still at the step before generation. Its inlining grows the component from 2.04MB to 2.70MB, and `Component::from_binary` — Cranelift AOT, which dominates `kiln/run` — scales with that size. `kiln_provider.rs` compiles generators at O1.

`kiln/aot` and `kiln/generate` (and `kiln/probe`) are now spans of their own, so the next reader sees native compilation apart from what the generator then does, on both the run and the probe path.

## Gated post-loop fixpoints

`run_bounded_fixpoint` drives the cleanup fixpoints that run after the main optimization loop. Each now owns a `FunctionGate`: round 0 processes every function, each later round only what the round before rewrote and its 1-hop neighbours. The fixpoints iterate over the whole module otherwise, so their cost is the module's size times however many rounds the slowest function needs — on the gale generator, `store_load_forward_post_scalarize` spent 2.4s a round across four rounds with no decay at all.

Four passes take a gate and a column of their own: `store_load_forward`, `const_folding`'s uncached entry, `const_branch_prune` (both entries share one column, since each fixpoint gets a fresh gate) and `condition_implication::eliminate_post_promote`. Each round also emits a `<name>/round <i>` span, so the decay is visible rather than inferred.

`CallGraph::build` deduplicates a function's callees through an `IndexSet`. It is built once per gate, and a post-inlining body can call hundreds of distinct functions.

## Still on the table

The main loop's `param_spec` (9.7s) and `inline` (8.4s) rebuild whole-module tables — `collect_call_constants`, `reachable_function_positions`, `written_locals`, `body_has_loop`, `find_recursive_functions` — every round, before the gate gets a say. Incrementalizing them needs "revision unchanged implies body unchanged", which the gate does not promise: it is built so that skipping costs optimization quality and never correctness, and a cache keyed on it could hand the inliner stale facts. Asserting that invariant in debug builds is the step before the optimization.

https://claude.ai/code/session_0164e5YXFWQtu9KT5krsFw8q

---
_Generated by [Claude Code](https://claude.ai/code/session_0164e5YXFWQtu9KT5krsFw8q)_


gfx:
Fixed in 922b518b2.

The finding is right: `FunctionGate::mark_changed` bumps the rewritten function's revision and, conservatively, its direct callers' and callees' as well, so a later round can process a neighbour whose body did not change. `gate.rs`'s own module doc already says this; the two descriptions of the post-loop fixpoints had dropped it.

It was one class with two instances, not one comment. The second was the `run_bounded_fixpoint` doc comment in `wado-compiler/src/optimize.rs`, which made the same claim in the same words. Both now read as the gate's contract does.

Nothing skipped.

---

gfx:
Both fixed in `fa8565da3`.

**`any_pending` before whole-program setup** — real, and grounded in a rule this repo already states. `docs/optimizer.md` says "a drained column skips the round before the pass builds any whole-program state", and the loop keeps it: the `gated!` macro is `gate.any_pending($id, len) && $pass(p, &mut gate)`, so a drained pass is never entered. The four post-loop entries built their maps first and read the gate second.

It is one class with four instances, so all four now read the column first:

| entry | setup a drained column no longer pays for |
| --- | --- |
| `forward_stores_to_loads` | `first_param_types`, `CallImmutability::new`, `build_ctfe_builtin_map` |
| `fold_constants_uncached` | `build_fold_maps`, `build_global_view` |
| `const_branch_prune::run_rule` | `pure_builtin_callee_ids` (one guard serves both public entries — they share the column) |
| `eliminate_post_promote` | `resolve_panic_ids`, `CallImmutability::new` |

This is the cheap half of the follow-up the description names. The other half — `param_spec` and `inline` — is untouched: those need "revision unchanged implies body unchanged", which the gate does not promise, and keying a cache on it could hand the inliner stale facts.

**Wording** — taken, in both places, since this push already changes those files.

e2e: 5596 passed, 0 failed, unchanged from before the guards.

---

PR #2121, opened 2026-09-22
perf(optimize): keep the post-loop const fold whole-module

gfx:
`fold_constants_uncached` walks every function on every call, and the optimizer's dirty-set gate no longer sits in front of it. The pass folds reads against a whole-program view of the constant stores to immutable globals, and a store reaches its readers through the global rather than through a call, so the gate's 1-hop caller/callee propagation cannot re-open the functions that need folding. It still reports what it rewrote, so the passes downstream of it stay gated as before.

The gate is not buying anything here either: on the `package-geas` cold start (`--no-cache`, the heaviest path through the optimizer) the gated and ungated forms measure 43.1s and 42.9s median over three runs, with byte-identical output.

## Mechanism

`GatedPass` loses its `ConstFoldUncached` column, since nothing consults it. The pass keeps the `&mut FunctionGate` parameter and calls `mark_changed` for each function it rewrites, which is what re-opens the later passes in the same fixpoint round.

`docs/optimizer.md` records why this one entry is exempt, next to the paragraph describing how the post-loop fixpoints are gated.

https://claude.ai/code/session_0164e5YXFWQtu9KT5krsFw8q

---
_Generated by [Claude Code](https://claude.ai/code/session_0164e5YXFWQtu9KT5krsFw8q)_


gfx:
Fixed in 5280914e1.

`docs/optimizer.md` — the paragraph stated the rule and its exception in one breath, so the opening sentence read as a universal claim that the rest of the paragraph then withdrew. It is now two paragraphs: what every post-loop cleanup fixpoint does with its gate, then the one step that declines to read it and why. The split also lets the exception say that `fold_constants_uncached` marks what it rewrites, which is what keeps the steps beside it gated — the fact a reader needs to tell "ungated" from "unhooked from the gate".

Nothing skipped; this was the only finding.

---

PR #2122, opened 2026-09-22
refactor(stdlib)!: one parser per type, and a readable reason on every error

gfx:
Every error type in `core:*` implements `Display` and `Error`, so `` `${e}` `` renders why something failed, and a caller holding one only through a trait bound can still report it. Every type that parses text does so through `FromStr` under one name, taking any `AsStrSlice`. Between them, an error reaching a user says what went wrong, and a generic function bounded by `FromStr` reaches every type that can be parsed.

## `Error`, and `?` across error boundaries

`Error: Display` is what every standard library error type is; a bound on it says only that the failure has a readable reason. `FromStr::Err` and `LenientFromStr::Err` carry that bound, so a parse failure reached through the trait can always be reported. `wado-compiler/tests/fixtures/from_str_err_needs_error.wado` pins the rejection.

A wider error that carries a narrower one interpolates it rather than wording the failure again, and declares `impl From<Narrower> for Wider` so `?` converts at the call site. Four such impls replace about thirty hand-written `match e { Ok(v) => …, Err(e) => Result::Err(e) }` arms.

## One parser per type

`FromStr` has a single method, `from_str<S: AsStrSlice>(s: S)`. `AsStrSlice` already admits an owned `String`, a reference to one, and a view, so the split into `from_str_slice(&StrSlice)` plus a `from_str` default carried no distinction. Parsing a field out of a larger buffer still allocates no substring:

```wado
i32::from_str("xyz42abc".as_str_slice().sub(3, 5))
```

`core:temporal`'s `Plain*::parse` and `core:url`'s `Url::parse` are gone the same way. Each `impl FromStr` documents the spelling it accepts, and each module's synopsis shows the call. The `Url` impl also says where a relative reference goes (`Url::resolve`), which nothing stated before.

`bool` and `char` gained `FromStr` impls — `ParseBoolError`, and `ParseCharError` with `CharErrorKind` — mirroring Rust's types and messages. `char`'s `LenientFromStr` delegates to it, since one Unicode scalar has one spelling; `bool` keeps its own lenient body for `TRUE` / `1` / `0`.

Those two impls are what let `core:router` stop hand-rolling its `bool` and `char` path params. Both arms are now one call to the shared `parse_scalar`, and the `char` arm no longer re-derives the UTF-8 lead-byte length table `core:prelude` already has. Their errors gain the parser's reason, as the integer arms already had:

```
path param 'on' is not a valid bool: 'TRUE' (provided string was not `true` or `false`)
```

## Positions that are true

`Utf8Error` carries where decoding stopped, and `rebased(base)` shifts it for a decoder working in pieces — `core:kiln`'s streaming reader owes it that, or it reports an offset into a buffer its own caller never saw. `FsErrorKind::NotUtf8` carries the `Utf8Error` instead of discarding it.

`DeserializeError`'s positionless constructors default `offset` to `-1`, and `Display` prints `at offset N` only when there is one. A caller with no position omits the argument rather than passing `0`, which the reader would otherwise trust.

A fields value that will not serialize now reaches `core:log`'s installed sink under `log.fields_error`, rather than being dropped to `Value::Null` around it.

## Compiler fix

`shadowed_name` fired on a bare case pattern whose type the pattern's module does not import, suggesting a rename that would change which case is matched. The resolver answered from the module's own case tier, but a bare case pattern reads against the scrutinee's type, which only the elaborator knows. `Scopes` now holds a program-wide set of case names for the lint to consult. It feeds the lint alone, so it can only make the warning quieter, never change what code means.

## Breaking

`FromStr::from_str_slice`, `PlainDate::parse` and its four siblings, and `Url::parse` are removed. Each has one replacement: `from_str` on the same type.

Generated stdlib docs and golden WIR fixtures were regenerated.

---
_Generated by [Claude Code](https://claude.ai/code/session_013VnMjzEzoGEK3xHZqxbdWp)_


gfx:
All four findings verified against the current code; all four were real. They fall into three classes, and each is fixed at the class rather than at the site.

## A stream handle abandoned on an error path (`read_text`)

Confirmed, and the reason is one level up from `read_text`: `synthesis/resource_cleanup.rs` excludes `GenericResource` — `Stream<T>` and `Future<T>` — from drop elaboration, so those handles are the author's to drop on every path, and nothing diagnoses a path that misses one. The emitted TIR for a `Stream<u8>` parameter carries no drop at all:

```
fn take(s: Stream<u8>, fail: bool) -> Result<i32, i32> {
    let chunk: StreamChunk<u8> = core::rt::cm_stream_read_u8(&s as i32, 16);
    if fail {
        return Result<i32, i32>::Err(1);     // no drop
    }
```

`read_text` now carries the decode failure out of the loop through a labeled block, so `content.drop()` is reached before every return — verified in the TIR, not just by reading.

Swept the tree for the shape. Every other reader — `core:fs` (`read`, `read_dir`, `stream_out`, `finish`), `package-gale`'s `read_all` and `write_output`, and the two in `extract_antlr4_descriptors.wado` — already drains, drops, and only then decides. `read_text` was the one that decided while still holding the handle. The gap itself is now written down in WEP 2026-05-21, which documented the exclusion nowhere.

No test: `Stream::new()` written and read from one task deadlocks (`event loop cannot make further progress`), so the malformed-input path cannot be reached from a test block. Said here rather than adding an assertion that holds by construction.

## Empty port rejected for an IPv6 host

Confirmed, and it was an inconsistency inside one function rather than one host form: `host:` was accepted as naming no port while `[::1]:` failed. WHATWG's port state reads the buffer only when it is not empty, so both forms now go through a single `apply_port_suffix`. The regression test asserts both, and was red on exactly the `[::1]:/` line with the reg-name half already passing.

## Documentation pointing at a removed parser

Confirmed. `PlainDate::new`'s doc comment still named `parse`; `docs/stdlib-core-temporal.md` is generated from it, so the two findings are one fix plus a regeneration. The earlier sweep looked for call syntax and prose without it survived — so this time the sweep was for the bare names across every file type, which turned up one more the reviews did not reach: `opt_loop_convergence.rs` called `T::from_str_slice` inside a Rust string literal. That one broke the build; it is fixed in the same branch.

Nothing was skipped as not-real.

Verification on the merged tree: e2e 5676 passed / 0 failed; integration 764 / 0; stdlib 1992 / 0; `test-wado` 5698 passed / 0 failed / 15 todo across 1399 files.

---

PR #2123, opened 2026-09-22
feat(compiler): f16 and bf16 as storage-only primitives

gfx:
A machine learning checkpoint is mostly half precision, and most of the ones published today are bf16. Wado can now name those element types: `f16` and `bf16` are primitives whose representation the compiler knows, so a `List<f16>` is a packed 16-bit array rather than a million GC objects, and a Loam graph declaring `float16` or `bfloat16` compiles. They carry bits, not arithmetic — Wasm has no half precision instruction to lower an operator to, and neither caller computes in half precision. `core:cbor` gets the type its canonical float ladder was missing.

The design is [WEP: Half-Precision Primitives](https://github.com/wado-lang/wado/blob/claude/f16-bf16-compiler-design-c1vc74/docs/wep-2026-09-22-half-precision-primitives.md).

## The API

Bits go through `to_bits` / `from_bits`, values through `From` / `TryFrom`. `From` widens and is exact; `TryFrom` narrows and answers `Ok` only where the value survives the round trip, which is the question a canonical encoder asks. `from_f32` narrows unconditionally, rounding to nearest even and saturating to an infinity.

Neither type takes part in `as` at all. `f32 as u32` converts a value and `f16 as u16` would reinterpret bits; one syntax cannot carry both meanings, so every `as` naming a half is a compile error whose message names what to write instead.

`Display` widens to `f32`, so every format specifier works unchanged. `Inspect` uses exponent notation. `Serialize` widens to `f32` and `Deserialize` rounds back, so JSON and CBOR carry a half with no new `Serializer` method.

## The boundary

The Component Model's `defvaltype` stops at `f32`, so an `export fn` naming a half is refused. Every `export fn` is checked, not only the one the world names — each lands on the component's surface. That check was missing entirely: an extra export reached WIT emit, which only warned and dropped the section, and `v128` had the same hole.

## The bit accessors are unsigned

`f32::to_bits` answers `u32` and `f64::to_bits` answers `u64`, with `from_bits` taking the same. The signed forms mirrored the Wasm opcode names, and almost every caller in the corpus undid the signedness on the next token. The `builtin::` layer keeps the opcode names, and the conversion lives in the one prelude wrapper above it.

## One registry for the primitives

`PrimitiveType` lives in `wado_compiler::primitive`, and the name lists that used to be written out by hand — in `method_lookup`, `name::is_builtin_shape_name`, the elaborator's static type resolution, `doc`, and a regex in `wado-cli`'s TextMate generator — all derive from it. `ModuleSource::of_primitive` answers which module holds a primitive's impls, so a primitive whose methods are not in `prelude/primitive.wado` resolves like any other. Every `match` over `PrimitiveType` is exhaustive, so a new primitive has to choose an answer rather than inherit a wildcard's.

The generated TextMate grammar and the stdlib reference docs were regenerated.

---

https://claude.ai/code/session_01Tq2MnKqZp51Tf5XyfppKK3

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tq2MnKqZp51Tf5XyfppKK3)_


gfx:
The `bf16::from_f32` finding is correct, and fixed in `6c7f4fd8f`.

The claim that nothing overflows and nothing becomes subnormal was false in both halves. A mantissa carry out of `0x7F7F` walks into the exponent and reaches `0x7F80`, an infinity; an `f32` subnormal keeps its zero exponent and lands on a `bf16` subnormal. `round_to_even` covers both already, so only the comment was wrong.

The replacement says what the code cannot: bf16 takes f32's exponent field verbatim, which is why the top 16 bits are the result and only the mantissa rounds. `docs/stdlib-core-prelude.md` copies that comment and was regenerated.

`origin/main` is also merged in (`8cfdf9a96` records the conflicts, `c2ce1b602` resolves them). The one source conflict was in `core:prelude`'s re-export list, where main replaced the `primitive.wado` glob with a named `pub use`; the `half.wado` glob sits beside it now. The golden WIR fixtures, the format fixtures and the stdlib reference docs were regenerated by their generators rather than hand-resolved.

---

gfx:
The benchmark alert is runner variance, not a regression from this branch. The `Benchmark` check itself passed — every ratio is under the 200% fail threshold.

The gh-pages history shows the whole suite moving together from run to run, and the baseline it compared against happened to be the fastest run of the day:

| commit | json/twitter/ser | mandelbrot | fts | zlib/decompress | sieve |
|-|-|-|-|-|-|
| `ed4d10a1b` (main) | 583.91 | 5.24 | 18.15 | 331.18 | 245.8 |
| `c1edabb9d` (main) | 585.04 | 5.24 | 18.15 | 331.41 | 244.47 |
| `c0bf19e14` (main, the baseline) | 1020 | 7.24 | 29.41 | 569.85 | 413.16 |
| `6c7f4fd8f` (this branch) | 580.16 | 5.24 | 18.15 | 328.73 | 247.29 |

This branch lands on main's typical numbers to three significant figures — `fts` is exactly 18.15 and `mandelbrot` exactly 5.24 on all three of those runs. `mandelbrot` and `syntax_highlight` moved by the same factor as everything else and simply stayed under the 1.5 threshold, which is what a host-speed change looks like rather than a codegen change.

Nothing in the branch is on these paths at runtime in any case: the primitive registry work is compile-time, and no benchmark names `f16` or `bf16`.

Worth noting separately, since it will keep happening: the alert compares one commit's absolute throughput against one predecessor's, with nothing controlling for runner speed, so any PR that follows an unusually fast baseline reads as a large regression. `benchmark/ab.ts` already solves this by running reference rows in both arms as a control. Wiring that idea into the CI alert is a change for its own pull request, not this one.

---

PR #2124, opened 2026-09-22
feat(loam,protobuf): compile an ONNX model to shape-checked Wado, over a new core:protobuf

gfx:
An ONNX model compiles to Wado through Kiln, and the shapes are checked while it
compiles. `use { forward, Weights } from "./model.onnx"` gives a `forward` whose
argument and result types carry one marker type per axis, so feeding a
`Tensor<f32, [Batch, In]>` where `[Batch, Out]` is wanted is a type error at the
call. Loam reads the model in its binary form, which needs the protobuf wire
format, so `core:protobuf` arrives with it: `#[wire(number = N)]` on a struct's
fields, `WireNumbered` as the bound a number-keyed format asks for, and an
encoder and decoder over `core:serde`.

Eight graphs from ONNX's own backend test data compile through Loam and
reproduce, within 1e-6, the outputs ONNX ships beside them.

## `core:protobuf`

The encoder runs a size pass and then a write pass, so a nested message's length
prefix is known before its bytes are written and no buffer is copied to make
room. The decoder reads varint, `fixed32`, `fixed64` and length-delimited
fields, takes a repeated field packed or unpacked, and skips a field the type
does not declare.

`#[wire(number = N)]` sits on struct fields and reaches both sides of serde.
A struct numbers every field or none, which is what `WireNumbered` answers, and
`FieldSchema::by_number` is how a decoder turns a tag into a field.

## Loam

Stage 0 is the shape algebra: axis identity is a type and an extent is a value,
and the schema's `layout` is what names an axis. It is wider than the kernels on
purpose, so `emit` reports where a kernel is narrower rather than silently
lowering something the kernel does not cover.

The kernels are `matmul`, `zip`, `zip_suffix`, `transpose2` and `map`, inlined
into the generated module so it stands alone. `Gemm` lowers to them under its
transpose flags, a `Transpose` that moves no axis is its operand, and a
`Constant` of one element folds into the closure that reads it rather than
allocating a tensor.

Every identifier the generated module declares is minted from one set. A tensor
name comes out of the model file, so one that collides takes a trailing `_`; an
axis name is the schema's own word, so a collision stops the build and names it.

## Compiler fixes

- `break` and `continue` inside a tuple or variadic for-of are refused, since
  the loop is expanded at compile time and neither has a loop to name. The check
  walks every block the body can hide one in, and `break LABEL` is allowed:
  it names a labeled block, which the expansion does not move.
- A local struct's field default resolves as a top-level struct's does, so a
  list literal reaches `ReflectStruct::defaults` through its `From<Array<T>>`
  coercion.

## Test data

The models and tensors Loam's tests read are committed under
`package-loam/tests/onnx/`, keeping ONNX's own layout, with the provenance and
the upstream commit in a `README.md` beside them. A submodule is absent wherever
it is not initialized, and a file an import names has to be there for the module
to compile, so reading one turns a test into a build failure.
`mise run sync-onnx-data` fetches them, reading the tests for which files to
take.

## Summary by CodeRabbit

* **New Features**
  * Added Protocol Buffers serialization and deserialization through `core:protobuf`.
  * Added `#[wire(number = N)]` field numbering with validation for protobuf-compatible data.
  * Loam can now import binary `.onnx` models as well as `.onnxtext` graphs.
  * Generated Loam modules now include tensor runtimes, weights, and executable forward passes.
  * Added support for ONNX `Constant`, `Gemm`, and `Transpose` operations.

* **Documentation**
  * Added the Grog Protocol Buffers proposal and expanded ONNX support documentation.

* **Tests**
  * Added comprehensive Protocol Buffers, tensor runtime, compiler, and ONNX conformance coverage.
  * Added validation for generated and vendored file review filters.


gfx:
## Review response

Eleven findings, all verified against the code before anything was written. They fell into three classes, and each class was closed rather than each site patched.

### Class 1 — a reader took a file's word where the file could contradict itself

`enter_packed` accepted any non-`LEN` tag for any scalar, so a tag naming the wrong wire type was read as if it named the field's own: `\x0d\x01\x10\x07` decoded to `{a: 1, b: 7}` from bytes that say no such thing. It now takes the scalar's `WireType` and refuses a tag that is neither it nor a packed run. `read_varint` refuses a tenth byte whose payload is above bit 0, instead of dropping the overflowing shift.

On the Loam side, a tensor's payload length is now checked against the product of its extents, so `raw_data` short of its shape is a read error rather than a tensor whose elements come from nowhere.

Both are pinned by tests that fail without the fix — the first two I wrote passed without it, each erroring three steps later for an unrelated reason, so they were replaced with inputs the old code decoded silently and wrongly.

### Class 2 — an invariant two callers had to remember separately

`Graph.inputs` never names an initializer. `onnx_proto::read` enforced that and `onnx_text::read` did not, which is the `Names::of` collision the review found at `emit.wado:182`. Both readers now pass their declared inputs through one shared `parameters` in `graph.wado`, and the struct states the invariant. `Gemm` refuses a bias of higher rank than the product rather than widening the node past the two axes it has.

### Class 3 — a script or a document stating something near the truth

`sync-onnx-data.sh` refuses a path climbing out of `tests/onnx`. Both WEP roadmaps state done-ness as a checklist, as `docs/CLAUDE.md` asks; the numbering is load-bearing (code and prose name "roadmap step 2"), so each item keeps its number and gains a box. The Loam WEP no longer claims this repository commits only what it can diff — `tests/onnx/` holds 31 `.onnx` files, and the text about them now says they are copies of ONNX's own data rather than anything edited here.

### Skipped, with reasons

- **Duplicate `#[wire(number = N)]` and `WireNumbered`** — `is_numbered_struct` is imprecise, but unreachably so: the duplicate is already a compile error at the declaration, which `serde_wire_number_duplicate.wado` pins. No program reaches the bound's answer.
- **`Constant` tensor shape** — recorded as a known gap. A one-element `Constant` is rank 0 whatever shape it declares, so one of shape `[1]` loses that axis. The consequence is a build error at an output rank check, never a wrong number. Carrying the extents would need a name for each axis, and nothing in the graph gives one.
- **Non-contiguous runs of a repeated field** — already a known gap in the Grog WEP. The text now says the earlier elements are dropped without an error, and that concatenating two encodings of one message is the format's defined merge and produces exactly this.
- **`dim_value == 0` reading as `?`** — recorded as a known gap resting on one of Grog's. `TensorShapeProto.Dimension` is a `oneof`, so presence is on the wire, and `core:protobuf` has no `oneof`. The textual reader keeps the distinction.

### Then distilled

`/distill` over the whole branch, three passes. `float_tensor`'s two payload paths share one tail, and `raw_floats` reads the little-endian bytes through `core:protobuf`'s own `read_fixed32`. `operand_rank` is the single place an `OperandRankConflict` is built. `element_width` moved onto `WireType` beside `bits`, and `last_packed` takes the wire type rather than a width whose `0` meant "a varint" — that magic value was this branch's recurring shape, a predicate standing near its rule instead of on it.

### Verification

| suite | result |
| --- | --- |
| `mise run test-wado` | 5786 passed, 0 failed (1415 files) |
| `cargo test -p wado-compiler --test e2e` | 5724 passed, 0 failed |
| `cargo clippy -p wado-compiler --all-targets` | clean |
| `mise run update-golden-fixtures` | no change |

PR #2126, opened 2026-09-22
fix(compiler)!: reject an impl that does not match its trait

gfx:
An `impl` binds every associated type its trait declares and defines every method it requires. The compiler compares the two member sets where every declaration and impl is in hand, so an incomplete impl is a diagnostic at its own header rather than an unsubstituted projection reaching code generation.

`TryFrom` names its error `Err` and bounds it by `Error`, so a caller reaching a conversion failure through the bound can always report the reason.

Five incomplete impls already in the tree are completed here. They are what the absent check had been admitting.

## What an impl owes its trait

| member | supplies every required one | may add its own |
| --- | --- | --- |
| methods | yes — the ones declared without a body | yes, as a helper its own bodies call on `self` |
| associated types | yes — all of them; there are no defaults | no |

A trait method written with a body is a default, which `has_body` on the method digest records. An associated type has no such fallback, so every one is required.

A method the trait does not declare is a helper reachable through that impl alone ([WEP: Trait Resolution](docs/wep-2026-09-01-trait-resolution.md)), and stays legal. An associated type the trait does not declare is a typo for one it does, which is how the real type comes to be unbound, so it is rejected.

A derivation request (`impl Trait for Type;`) writes no members, so it owes none. An `interface` has no trait declaration to compare against and never reaches the check.

Three diagnostics, all `TRAIT_DECL_INVALID`:

```
impl of trait 'Carry' does not bind associated type 'Load'
trait 'Carry' declares no associated type 'Cargo'
impl of trait 'Carry' does not define method 'other'
```

## `TryFrom::Err`

```wado
pub trait TryFrom<T> {
    type Err: Error;
    fn try_from(value: T) -> Result<Self, Self::Err>;
}
```

The bound is what a generic caller may rely on: through it there is no impl to read, so an unbounded error type is a value that says nothing. `FromStr::Err` carries the same bound for the same reason.

The name is `Err` because it names the `Result`'s `Err` side, and because `type Error` gives the name two meanings inside an impl that also has the `Error` trait in scope. Every stdlib impl uses `ConvertError`, which implements `Display` and `Error`.

## Impls completed

| file | missing |
| --- | --- |
| `lib/core/serde_test.wado` | `serialize_i128`, `serialize_u128` |
| `tests/fixtures/const_object_globalization_mut_borrow_projection.wado` | `serialize_null` |
| `tests/fixtures/compound_assign_trait_operand.wado` | `type Output` on `Shl` and `Add` |
| `tests/fixtures/operator_operand_span_error.wado` | `type Output` on `Shl` |

## Breaking

An impl that omits a required member no longer compiles. Every example in [WEP: Operator Overloading](docs/wep-2026-01-18-operator-overloading.md) already writes `type Output`, and the four files above are the whole of what the corpus had been getting away with.

`TryFrom` impls outside this tree write `type Err` in place of `type Error`.

https://claude.ai/code/session_01QHaydHwztGfE7Ne8opMtaZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHaydHwztGfE7Ne8opMtaZ)_

## Summary by CodeRabbit

* **New Features**
  * Added validation for trait implementations, including required associated types, declared associated types, and required methods.
  * Added clear compiler diagnostics for invalid trait implementations.
  * Updated `TryFrom` conversions to use the `Err` associated type, requiring errors to implement `Error`.

* **Documentation**
  * Expanded guidance on conversion traits, associated-type bounds, trait implementation rules, and supertrait behavior.

* **Tests**
  * Added coverage for trait conformance errors and conversion error reporting.


gfx:
Answered the review. Both findings were real; one of them is fixed in the opposite direction to what was proposed.

## Fixed

**The conformance check's two sides disagreed about which members an impl owes** (`63e653c`). The required side read the trait's own associated types; the accepted side also took a supertrait's, so a name it admitted could still go unbound — the very projection the check exists to catch.

The remedy proposed was to require supertrait-declared types of a subtrait's impl. That would reject the prelude on the first file: `impl<T> AsSlice for List<T>` binds no `Elem`, which its supertrait `Sequence` declares, because `impl Sequence for List<T>` does. [WEP: Super Traits §Obligation](https://github.com/wado-lang/wado/blob/main/docs/wep-2026-07-27-super-traits.md) settles it — `impl Sub for T` owes `T: Super`, satisfied by a separate impl or by structural derivation, and *that* impl is where `Super`'s members live.

So the accepted side is what was wrong, and it is now strict: an impl binds exactly the associated types its own trait declares. Two fixtures pin both halves — `trait_impl_subtrait_omits_supertrait_assoc.wado` (the `AsSlice` shape, compiles) and `error_impl_assoc_type_of_supertrait.wado` (binding the supertrait's name, rejected). The rule is recorded in both WEPs that own a half of it.

**Docs describing a `TryFrom` shape the code no longer has** (`63e653c`). Two instances, one of them reached by the review:
- `docs/wep-2026-03-16-conversion-traits.md:110` — the comment named a string error where the impl returns a `ConvertError`.
- `docs/wep-2026-09-22-half-precision-primitives.md:117-118` — still `type Error = ConvertError;`, the pre-rename spelling. Same parallel-landing miss as `half.wado` (fixed in `653f578`).

## Skipped

- `docs/research-from-into-framework.md:43,48` also reads `type Error`, but it is a ```rust block quoting Rust's own `std::convert`. Correct as written.

## Also in this round

`b657da80` is a `/distill` pass over the branch: comments back to the two-line cap, and a stray doc paragraph on `trait_sig_of` describing a different item, deleted.

## Verification

On the merged tree: `mise run test` exit 0 (e2e 5738 passed, 0 failed), `mise run test-wado` exit 0 (1415 files compiled and loaded, 5786 tests passed, 0 failed), `mise run format` clean.

---

